Issue #1, opened 2021-07-06
Discussion for ARC-0001 - Algorand Wallet Transaction Signing API


fabrice102:
Draft is https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0001.md
See also proposed PR that may be listed below.

DanBurton:
Nitpick: I suggest the function be referred to as `signTxns` rather than `signTxn`, since it accepts an array of transactions to sign.

DanBurton:
> If txns[i].signers is an empty array, the transaction txns[i] MUST not be signed and ret[i] MUST be set to null.

I propose that transactions which are "not to be signed" may include a field `stxn` which may take the place of `null` in the result.

Proposed replacement to the above wording:

> If `txns[i].signers` is an empty array, the transaction `txns[i]` MUST not be signed and `ret[i]` MUST be set to `txns[i].stxn`, or `null` if the `stxn` field is not present on `txns[i]`.

DanBurton:
For reference, my proposed changes regarding `txns[i].stxn` are influenced by [ARC-0010's additional requirements regarding LogicSigs](https://github.com/reach-sh/ARCs/blob/6fa84a13bec22baab6ced0d0fac10e2a3b1b8170/ARCs/arc-0010.md#additional-requirements-regarding-logicsigs) and writing [the source for DemoWallet.js](https://github.com/reach-sh/algo-demo-wallet/blob/2d1e5d79398ac2f21acab8362d074004c5390652/DemoWallet.js#L71-L73).

fabrice102:
Thanks for the feedback @DanBurton . See #29 and #28

PR #4, opened 2021-07-19
ARC-0001: Add support for a message per group

fabrice102:
algorandfoundation/ARCs#1


jasonpaulos:
This is a great addition 👍

emg110:
Considering having NOTE field which already exists with the same identical mission.

PR #6, opened 2021-07-19
ARC-0003: Support for metadata in note fields

fabrice102:
+ minor fixes and additional clarifications

algorandfoundation/ARCs#3

Do not hesitate to comment below.
Also, feel free to propose better URL than `algorand-ledger://txn/this/note`

### DIFF
--- a/ARCs/arc-0003.md
+++ b/ARCs/arc-0003.md
@@ -34,13 +34,15 @@ The ASA parameters should follow the following conventions:
         * If the resulting asset name can fit the *Asset Name* field, then `<name>` **SHOULD** be equal to the name in the JSON metadata file.
         * If the resulting asset name cannot fit the *Asset Name* field, then `<name>` **SHOULD** be a reasonable shorten version of the name in the JSON metadata file.
 * *URL* (`au`): a URI pointing to a JSON metadata file
-    > Shai: why not call it URI?
-    > Fabrice: This is the name in the specs
     * The URI **SHOULD** be persistent and allow to download the JSON Metadata file forever
-    * **RECOMMENDED** protocols (for compatibility and security): *https* and *ipfs*
-    >addie: what about I2P or arweave?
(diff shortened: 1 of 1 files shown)


barnjamin:
looks good to me

Why `algorand-leger` vs `algorand`?  `algorand` already taken by the txn generation protocol? 

It seems like this scheme would need some more definition. Possibly a new spec (sorry) or addition to the existing `algorand` spec defining a path to a distinct field in the ledger?

A separate direction for the url field may be to set it to `protocol://asset/{id}/create/note` where the `{id}` field is known to be swappable with the asset id.  The `create` keyword would need to be a special word in this case. 

This would give the benefit of allowing something like `https://example.com/{id}` where the id can be swapped in a block explorer or other client to the asset id and link directly to a page for that asset.

fabrice102:
These are very good comments.

The reason for `algorand-ledger` vs `algorand` was that `algorand` is already used for another purpose: https://developer.algorand.org/docs/reference/payment_prompts/
I wanted to avoid any potential conflict.

`protocol://asset/{id}/create/note` looks like a good option too.

I was thinking to do something reasonable for this protocol so that it can later be extended to a full spec.
This is to avoid delaying too much the publication of ARC-0003

fabrice102:
That makes sense. What we can do is just change ARC-0003 to allow for `{id}` in the URL.
So that later, we can just define the `algorand-ledger://` protocol format.
Do you know if any other blockchain has a URI for this purpose?
(I know blockchains have URI to make transactions, but my very quick search found no URI to read committed transactions.)

barnjamin:
https://w3c-ccg.github.io/blockchain-links/ and https://github.com/bitcoin/bips/blob/master/bip-0122.mediawiki may be decent prior art, though neither of them go down to the field of a txn.

It seems like there is a lot of overlap between a protocol like this and the Algod API so that may be a good target to achieve the right level of expression?

fabrice102:
subsumed by #6

PR #8, opened 2021-08-09
Reach wallet arcs (0005 thru 0011)

DanBurton:
I've also written a small runnable demo which demonstrates an implementation of all of the API functionality specified by this group of ARCs. [except for `window.algorand`, since it is a cli demo]

https://github.com/reach-sh/algo-demo-wallet

### DIFF
--- a/ARCs/arc-0005.md
+++ b/ARCs/arc-0005.md
@@ -1,0 +1,66 @@
+---
+arc: 5
+title: Algorand Wallet Transaction Signing API (Functionality Only)
+status: Draft
+---
+
+# Algorand Wallet Transaction Signing API (Functionality Only)
+
+> This API is a draft.
+> Some elements may change.
+> This ARC is intended to be completely compatible with [ARC-0001](ARC-0001.md).
+
+## Summary
+
+An API for a function used to sign a list of transactions on the Algorand blockchain.
+
+## Abstract
+
+[ARC-0001](ARC-0001.md) defines a standard for signing transactions with security in mind. This proposal is a strict subset of ARC-0001 that outlines only the minimum functionality required in order to be useable.
+
+Wallets that conform to ARC-0001 already conform to this API.
+
+## Specification
+
+### Interface `SignTxnFunction`
+
+```ts
+export type SignTxnFunction = (
+   txns: WalletTransaction[],
+   opts?: any,
+)
+   => Promise<(string | null)[]>;
+```
+
+A `SignTxnFunction`:
+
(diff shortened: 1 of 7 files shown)


jeapostrophe:
You can view rendered versions of these here:

https://github.com/reach-sh/ARCs/blob/reach-wallet/ARCs/arc-0005.md
https://github.com/reach-sh/ARCs/blob/reach-wallet/ARCs/arc-0006.md
https://github.com/reach-sh/ARCs/blob/reach-wallet/ARCs/arc-0007.md
https://github.com/reach-sh/ARCs/blob/reach-wallet/ARCs/arc-0008.md
https://github.com/reach-sh/ARCs/blob/reach-wallet/ARCs/arc-0009.md
https://github.com/reach-sh/ARCs/blob/reach-wallet/ARCs/arc-0010.md
https://github.com/reach-sh/ARCs/blob/reach-wallet/ARCs/arc-0011.md

jeapostrophe:
@fabrice102 And I chatted yesterday about ARC-0009 and how it is going to be very brittle to actually expose the real algosdk objects, when what we really care about is controlling access to the node behind them.

He proposed the idea of instead providing a simple HTTP client-style interface, with a `get`/`put`/etc interface that could be given to a DApp's version of the algosdk.

However, the pull request to js-algorand-sdk that would enable this was rejected --- https://github.com/algorand/js-algorand-sdk/pull/277 --- and in its conclusion they suggest override an internal variable with an undocumented interface.

If js-algorand-sdk refuses to support this pattern, then I think we're left with two choices:
1. (Current) Expose the SDK objects --- Brittle as the interface changes over time
2. Hack the planet and override the internal object and hope it works
3. Something else?

I think that we should move forward with these ARCs and try to produce a new version once we can get buy-in from Algorand to support this in their JS SDK.

jeapostrophe:
@DanBurton ^

DanBurton:
> I think that we should move forward with these ARCs and try to produce a new version once we can get buy-in from Algorand to support this in their JS SDK.

+1

If js-algorand-sdk doesn't want to play nice, we could consider creating a competing js sdk for algorand that does the things we want, and try to entice people to switch over to it. I don't love the idea of splitting the community, though. I do think that developers should be able to program against a single interface, and then it Just Works whether you're just using the js sdk, or going through a wallet.

ARC 9 is definitely brittle and is intended to be replaced at some point in the future with something better. But, in defense of ARC 9 as currently exists, I think it's reasonable to assume that js-algorand-sdk will at least maintain backwards compatibility with their `Algodv2` and `Indexer` classes. If this ends up being the case, then the idea isn't actually _that_ brittle.

PR #10, opened 2021-08-24
Update arc-0003.md

Bhaney44:
Added definitions in the summary section for ASA, NFT, and fungible token. Happy to hear any feedback. Thanks!

### DIFF
--- a/ARCs/arc-0003.md
+++ b/ARCs/arc-0003.md
@@ -8,7 +8,7 @@ status: Draft
 
 ## Summary
 
-This document introduce conventions for the parameters of Algorand Standard Assets (ASAs) for fungible tokens and non-fungible tokens (NFTs).
+This document introduce conventions for the parameters of Algorand Standard Assets (ASAs) for fungible tokens and non-fungible tokens (NFTs). An ASA is a digital proof, which may be tokenized for various purposes on the Algorand Network. Fungible tokens are digital assets which can be transfered between addresses via a blockchain network, and may operate as a type of currency or utility asset. By contrast, NFTs are decentralized data certificates, which authenticate both virtual and real world assets, such as a photographs, paintings, or music, on a blockchain.
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


ARCs/arc-0003.md
```diff
@@ -8,7 +8,7 @@ status: Draft
 
 ## Summary
 
-This document introduce conventions for the parameters of Algorand Standard Assets (ASAs) for fungible tokens and non-fungible tokens (NFTs).
+This document introduces conventions for the parameters of Algorand Standard Assets (ASAs) for fungible tokens and non-fungible tokens (NFTs). An ASA is a digital proof, which may be tokenized for various purposes on the Algorand Network. Fungible tokens are digital assets which can be transfered between addresses via a blockchain network, and may operate as a type of currency or utility asset. By contrast, NFTs are decentralized data certificates, which authenticate both virtual and real world assets, such as a photographs, paintings, or music, on a blockchain.
```
barnjamin, line 11: A `digital proof` and `decentralized data certificate` are new terms and unexplained. I don't think that even if they were expanded on they'd add anything to this spec.  The examples of what assets might be associated with an NFT/FT are valid though.  If you can compress this into 1 or 2 sentences just giving examples of those things I think this would be a good addition

barnjamin:
@fabrice102 I think this one can be closed

ARCs/arc-0003.md
Bhaney44: Sorry about the late reply, I never saw a notification for you feedback. How is this:

This document introduces conventions for the parameters of Algorand Standard Assets (ASAs) for fungible tokens and non-fungible tokens (NFTs).  Fungible tokens are digital assets which can be transferred between addresses via a blockchain network, and may operate as a type of currency or utility asset. NFTs are decentralized data certificates, which authenticate both virtual and real world assets, such as a photographs, paintings, or music, on a blockchain.

PR #13, opened 2021-08-30
ARC-0003: Support for Integrity Properties

fabrice102:
Currently, there is no easy way to check the integrity of files
pointed by the metadata JSON. In particular, when buying an
NFT, the seller may then change for example the image of the
NFT without the buyer having anyway to prove that the image
actually changed.

This PR proposes to fix this issue by adding additional integrity
properties.

### DIFF
--- a/ARCs/arc-0003.md
+++ b/ARCs/arc-0003.md
@@ -82,6 +82,10 @@ The JSON Metadata schema is as follows:
             "type": "string",
             "description": "A URI pointing to a resource with mime type image/* representing the asset to which this token represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
         },
+        "image_integrity": {
+            "type": "string",
+            "description": "The SHA-256 digest of the file pointed by the URI image. The property value is a single SHA-256 integrity metadata as defined in the W3C subresource integrity specification (https://w3c.github.io/webappsec-subresource-integrity)."
+        },
         "properties": {
             "type": "object",
             "description": "Arbitrary properties. Values may be strings, numbers, object or arrays."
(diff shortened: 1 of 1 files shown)


ARCs/arc-0003.md
```diff
@@ -92,13 +96,34 @@ The JSON Metadata schema is as follows:
...
+#### Integrity Properties
+
+Compared to ERC-1155, the metadata JSON schema allows to indicate digests of the files pointed by any URI field.
+This is to ensure the integrity of all the files referenced by the ASA.
+Concretly, every URI property `xxx` is allowed to have an optional associated property  `xxx_integrity` that specifies the digest of the file pointed by the URI.
+
+The digests are represented as a single SHA-256 integrity metadata as defined in the [W3C subresource integrity specification](https://w3c.github.io/webappsec-subresource-integrity).
+Details on how to generate those digests can be found on the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) (where `sha384` or `384` are to be replaced by `sha256` and `256` respectively as only SHA-256 is supported by this ARC).
+
+It is **RECOMMENDED** to specify all the `xxx_integrity` properties of all the `xxx` URI properties.
```
jannotti, line 108: DOn't we say **SHOULD** in these?

ARCs/arc-0003.md
```diff
@@ -92,13 +96,34 @@ The JSON Metadata schema is as follows:
...
+
+#### Sample
+
 An example of an ARC-3 Metadata JSON file follows. The properties array proposes some SUGGESTED formatting for token-specific display properties and metadata.
 
 ```json
 {
 	"name": "My Picture",
 	"description": "Lorem ipsum...",
 	"image": "https:\/\/s3.amazonaws.com\/your-bucket\/images\/{id}.png",
```
jannotti, line 125: Why are the slashes escaped with backslashes?

fabrice102: Apparently in JSON you can espace slashes by backslashes and you can also not do it.
The reason seems to be that when you embed JSON in HTML, you should never have `</` anywhere.
So you need to espace `/` by `\/` in that case.

The reason why this one has escapes is just copy-pasting from ERC-1155...

ARCs/arc-0003.md
fabrice102: RECOMMENDED and SHOULD have the same meaning in RFC: https://datatracker.ietf.org/doc/html/rfc2119

ARCs/arc-0003.md
```diff
@@ -106,6 +106,10 @@ The JSON Metadata schema is as follows:
             "type": "string",
             "description": "A URI pointing to a resource with mime type image/* representing the asset to which this token represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
         },
+        "image_integrity": {
+            "type": "string",
+            "description": "The SHA-256 digest of the file pointed by the URI image. The property value is a single SHA-256 integrity metadata as defined in the W3C subresource integrity specification (https://w3c.github.io/webappsec-subresource-integrity)."
```
robert-zaremba, line 111: Why `sha256`? Why not `sha3-256` or `blake2b`?

PR #14, opened 2021-08-30
ARC-0003 Support for Extra Metadata

fabrice102:
Previous version did not allow to store data that a smart contract could
access. This PR adds the option to hash extra metadata in the asset
metadata ASA parameter.

### DIFF
--- a/ARCs/arc-0003.md
+++ b/ARCs/arc-0003.md
@@ -40,7 +40,18 @@ The ASA parameters should follow the following conventions:
     * **RECOMMENDED** protocols (for compatibility and security): *https* and *ipfs*
     >addie: what about I2P or arweave?
     * **NOT RECOMMENDED** protocols: *http* (due to security concenrs)
-* *Asset Metadata* (`am`): the SHA256 digest of the JSON metadata file as a 32-byte string (as defined in [NIST FIPS 180-4](https://doi.org/10.6028/NIST.FIPS.180-4))
+* *Asset Metadata (Hash)* (`am`): 
+    * If the JSON metadata file specifies extra metadata `e` (property `extra_metadata`), then `am` is defined as:
+
+        ```plain
+        am = SHA-512/256("arc0003/am" || SHA-512/256("arc0003/amj" || content of JSON metadata file) || e)
+        ```
+
+        where `||` denotes concatenation and SHA-512/256 is defined in [NIST FIPS 180-4](https://doi.org/10.6028/NIST.FIPS.180-4).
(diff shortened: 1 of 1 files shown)


ARCs/arc-0003.md
```diff
@@ -132,6 +151,50 @@ An example of possible ASA parameters would be:
...
+#### Sample with Extra Metadata
+
+An example of an ARC-3 JSON Metadata file with extra metadata is provided below.
+
+```json
+{
+	"name": "My Picture",
+	"description": "Lorem ipsum...",
+	"image": "https://s3.amazonaws.com/your-bucket/images/MyPicture.png",
+    "extra_metadata": "iHcUslDaL/jEM/oTxqEX++4CS8o3+IZp7/V5Rgchqwc=",
```
jannotti, line 163: Funny indentation.

ARCs/arc-0003.md
```diff
@@ -132,6 +151,50 @@ An example of possible ASA parameters would be:
...
+}
+```
+
+The possible ASA parameters are the same as with the previous example, except for the metadata hash that would be the 32-byte string corresponding to the base64 string `aoKbNjrLXaeqI6eynopp//tfwmdS6s7XfRMOjK2L9jA=`.
+
+> For completeness, we provide below a Python program that computes this metadata hash:
```
jannotti, line 169: This is a great addition.

ARCs/arc-0003.md
```diff
@@ -78,6 +91,10 @@ The JSON Metadata schema is as follows:
             "type": "string",
             "description": "Describes the asset to which this token represents"
         },
+        "extra_metadata": {
+            "type": "string",
+            "description": "Extra metadata in base64. If the property is specified (even if it is an empty string) the asset metadata (am) of the ASA is computed differently than if it is not specified."
```
jannotti, line 96: I'm not sure we gain anything by specifying that extra-metadata is b64 encoded.  Why not leave it unspecified and concatenate that utf-8 encoding of the field, rather than decoding first.  I don't think that prevents people from doing what you specified, if they choose.  But if their metadata is not binary, I'd rather see it in the file unencoded.

ARCs/arc-0003.md
fabrice102: thanks! fixed

ARCs/arc-0003.md
fabrice102: Then we would need to have two fields: one base64, one not. I think it would become quite complex

PR #16, opened 2021-09-03
ARC-0003: Adding MIME types and OpenSea fields

fabrice102:
This PR is meant to try to provide clearer guidelines
so that clients such as marketplaces can more easily
display NFTs, following some discussions in
#3

Concretely:
1. It adds the fields `external_url`, `animation_url` and
`background_color` used on OpenSea.
2. It clarifies the use of those fields in the example of
a song.
3. It adds `_mimetype` fields that are optional
(for compatiblity with OpenSea) but highly recommended
4. It switched to the term `fields` for JSON fields
to avoid confusion with properties/attributes of the
asset.

Rationale:
1. Keep the fields in the "root" of the JSON metadata
file (as opposed to the properties) limited to
what is required for preview in a markeplace/gallery.
In particular `image` and `animation_url` can be,
for some assets, just preview, as opposed to the
full asset.
Remark for example that the field `image` restricts
a lot the maximum dimension of the image, so that
high-quality photography could not fit this field.
2. Full media files (e.g., full songs, high-quality photo),
licenses, and other such properties
are moved to properties and are left out of scope of this
proposal. Future proposals may try to standardize these
properties, although a global cross-chain standard
would be even more beneficial.

Points of dicussion:
1. Should MIME types be mandatory? This would break
backward compatibility with ERC-1155 and make most
existing NFT metadata invalid in the context of ARC-3.
The proposed solution is that clients may decide to
guess the MIME type from the extension of the file
(or from a HEAD query) when not available. The former
is what OpenSea seems to be doing.
This is much more brittle, hence the strong recommendation
for specifying MIME types.
2. Should URI properties (i.e., inside the `properties` field
and not at the root), should also follow the same rules.
In particular, should we allow `{id}` there? It may actually
be an issue as marketplace may not be able to recognize
which fields are URI.
3. Is this proposal too restrictive and should we finalize
ARC-0003 without it completely, sticking to being close
to ERC-1155 (with integrity fields and extra metadata though)
(In case this proposal is rejected, we may want to keep
the renaming of properties into fields.)

### DIFF
--- a/ARCs/arc-0003.md
+++ b/ARCs/arc-0003.md
@@ -110,7 +110,7 @@ The JSON Metadata schema is as follows:
         },
         "image_mimetype": {
             "type": "string",
-            "description": "The MIME type of the file pointed by the URI image. MUST be one of the MIME type image/* from the IANA media types list: https://www.iana.org/assignments/media-types/media-types.xhtml"
+            "description": "The MIME type of the file pointed by the URI image. MUST be of the form 'image/*'."
         },
         "background_color": {
             "type": "string",
@@ -126,7 +126,7 @@ The JSON Metadata schema is as follows:
         },
         "external_url_mimetype": {
             "type": "string",
-            "description": "The MIME type of the file pointed by the URI image. MUST be one of the MIME type from the IANA media types list: https://www.iana.org/assignments/media-types/media-types.xhtml. SHOULD be text/html"
+            "description": "The MIME type of the file pointed by the URI external_url. It is expected to be 'text/html' in almost all cases."
         },
         "animation_url": {
             "type": "string",
@@ -138,7 +138,7 @@ The JSON Metadata schema is as follows:
         },
         "animation_url_mimetype": {
             "type": "string",
-            "description": "The MIME type of the file pointed by the URI image. MUST be one of the MIME type from the IANA media types list: https://www.iana.org/assignments/media-types/media-types.xhtml. SHOULD be text/html"
+            "description": "The MIME type of the file pointed by the URI animation_url."
         },
         "properties": {
             "type": "object",
@@ -202,7 +202,6 @@ For example, if the field `hello_integrity` is specified, the field `hello` **MU
 Compared to ERC-1155, the metadata JSON schema allows to indicate the MIME type of the files pointed by any URI field.
 This is to allow clients to display appropriately the resource without having to first query it to find out the MIME type.
 Concretly, every URI field `xxx` is allowed to have an optional associated field  `xxx_integrity` that specifies the digest of the file pointed by the URI.
-MIME types **MUST** be one of the MIME type from the IANA media types list: https://www.iana.org/assignments/media-types/media-types.xhtml.
 
 > If MIME types are not properly specified, clients may need to rely on brittle methods to guess the MIME types such as using file extensions.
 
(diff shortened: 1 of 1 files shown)


runvnc:
Thank you sooo much! Couple of questions from glancing at the code changes:

* References a specific list of MIME types.  That list does not include a fewer newer MIME types or just less common, that may still be quite relevant.  For example, M3U (audio playlist, like for a music album) is `audio/xmpeg-url`.  Or for 3D models, becoming more common recently is `model/gltf+json`.  I have not found a single list of MIME types that is in fact comprehensive and up-to-date. 
* A few of them say that they SHOULD be text/html.  Maybe I don't understand the use of SHOULD.  They should be.. whatever they actually are?

wheresaddie:
confirming +1 on the following:
- specifying MIME types and maintaining comprehensive list
- I would merge ARC-0003 with integrity fields and extra metadata

for future PRs I would like to request we consider the following:

- develop a global cross-chain standardization for properties 
- MIME support for common (but possibly obscure) generative code based works usually released as .html, p5, mp4 and json
- default support for files up to 10gb (this is standard ceiling on ipfs)

Bhaney44:
**Asset Properties and Format**

I think [ARC-0003](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0003.md) provides a good overview of the asset properties, which as I understand it are as follows.

`NFT = [Unit_Name, Asset_Name, URL, Hash]`

First, we may consider adding an additional asset property for copyright and license. To claim a copyright under [U.S. Copyright Law](https://www.law.cornell.edu/uscode/text/17/401) requires notice in the following form.

`Copyright = ["Copyright", NAME, YEAR]`

We may also add a `License` property to allow for an attribution of an open source license.

Second, if I am correct, @fabrice102 suggested the rationale for using JSON files is because it will allow for simpler web integration and it may also facilitate [OpenSea](https://opensea.io/) integration. It may be worth considering using `.csv` file format because it is simpler and easier to work with. It will also be worth considering the value OpenSea provides as a marketplace builder and the degree to which integration should be facilitated. To me it appears to be a website building software for selling NFTs, but I am still very new and learning.

**Multipurpose Internet Mail Extensions**

Regarding [MIMEs](https://en.wikipedia.org/wiki/MIME), I am still unclear about the necessity for web properties such as URL and [URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier). Is this for purposes of hosting the NFT? Regarding support of files up to 10gb, as @wheresaddie suggested, how would this be facilitated? My intuition is it should be possible to store an NFT directly on-chain using a stateless smart contract or an Algogeneous smart contract, which includes both stateless and stateful functionality. Either way, I think adding this support is a great idea. 

**Cross Chain Properties**

Regarding cross chain properties, @fabrice102 mentioned [ERC-1155](https://eips.ethereum.org/EIPS/eip-1155), the multi-token standard. Additionally, I think reference should also be made to [ERC-721](https://eips.ethereum.org/EIPS/eip-721), the non-fungible token standard. I think it makes sense to focus on Ethereum interoperability to start, although Bitcoin interoperability may also be worth pursuing.

fabrice102:
@Bhaney44 

Regarding CSV vs JSON, I think that in practice JSON is now the defacto standard for data on the web.
CSV is also more limited in that you cannot nest properties.

Regarding MIME and URI, I am not completely sure what the issue may be.

Regarding cross-chain properties, my understanding is that ERC-721 is kind of subsumed by ERC-1155.
Since we are coming later, I guess that it is best to stick to ERC-721.

@wheresaddie @runvnc Let's remove the reference to a MIME list. By *SHOULD* I meant that the content is expected to be this. I'll rephrase it.

PR #18, opened 2021-09-06
ARC-0012 Name Service

barnjamin:
Very rough first draft for a Name Service

### DIFF
--- a/ARCs/arc-0012.md
+++ b/ARCs/arc-0012.md
@@ -25,12 +25,20 @@ The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL 
 
 An ARC-xxx Naming Service is embodied as two Smart Contracts to store Names and Properties and a method to Resolve the names:
 
+## Definitions:
+
+*Name* - A byte string used to provide a convenient reference to a set of Properties
+*Root* - The Domain under which the Name is registered
+*Property* - A Key/Value pair stored in Local State of a Name Record
+*Name Record* - A Contract Account with the Template Variable for the Name populated and Opted-In to a Name Registry 
+*Name Registry* - A Stateful Smart Contract that allows Name Records to opt-in and maintains Properties associated in their Local State
+*Owner* - The Account that has authority to update Properties for a Name Record
+
 #### Smart Contracts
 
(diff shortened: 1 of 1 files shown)


emg110:
Since Algorand has the naming infrastructure essentials (Private / Human Readable Public keys), IMHO,  it's a very important requirement to implement small remaining standardization, conventions, and implementations to have NS dynamic linking and the ability to use even more human-readable and shorter names. Making it DoS resistant would be paramount. I think this is a wonderful proposition.

barnjamin:
Also considering allowing something like `silvio.algo/twitter` where `silvio.algo` is the name and `/twitter` refers to a specific property rather than `twitter.silvio.algo`

ARCs/arc-0012.md
```diff
@@ -0,0 +1,194 @@
...
+
+- The key of a property MUST be a byte string
+- The length in bytes of a property key/value pair MUST be <= 128 bytes.
+
+
+# Contracts
+
+A Stateless Contract (Name Record) is used as a contract account. The logic is a standadized Template Contract with one Template Variable, the Name. This provides a predictable address to use for resolution.  Local State for the address is used to store arbitrary Properties that may be resolved for the name.  
+
+A Stateful Contract (Name Registry) validates the Name Record contract and allows storage of Properties assocaited with a Name to be stored in its local state.
```
jannotti, line 82: "associated"

ARCs/arc-0012.md
```diff
@@ -0,0 +1,194 @@
...
+- A Name Registry MUST allow the Name Record to deregister if the transaction is co-signed by the Owner
+
+
+### Name Register
+Initializing the Properties for a Name.
+
+To register a Name Record an Application OptIn transaction is submitted where:
+
+1. The ApplicationArgs array is populated with the Name Record bytecode and properties ([bytecode, key1, value1, key2, value2 ,...])
+2. The Account array is populated with the Name Record Account annd the registering Owner
```
jannotti, line 106: annd -> and

chicopaec:
@barnjamin 
> Also considering allowing something like silvio.algo/twitter where silvio.algo is the name and /twitter refers to a specific property rather than twitter.silvio.algo

What is the benefit over introducing a new path syntax that eschews the typical hierarchy of URLs? Using paths would be "$ID . $ROOT / $SUBID" whereas keeping with the traditional system would be "$SUBID . $ID . $ROOT". Not only more consistent, but also easier for people to learn.

Issue #9, opened 2021-08-15
Discussion for ARC-0000 :  ARC Purpose and Guidelines


fergalwalsh:
It would be great to have a document outlining the procedure for these ARCs. i.e. how they should be numbered, structured, decided upon, etc. Will rejected/abandoned ARC numbers be reused. Are the numbers required to be numbered N+1 or can a vanity number be chosen. Should a wide informal discussion take place somewhere before a formal ARC proposal? Or are these considered simply as documents and not necessarily standards of any kind?
These sorts of issues.

Perhaps the number 0 could be used for the ARC that retroactively defines the process?
If not an ARC, perhaps the repo README would be sufficient to define the process and expected structure?

I'd like to write an ARC for a specification so it would be useful to know the process. It would be self referential (like ARC3) so it would help to know the expectations for numbering.

Examples of initial proposals describing proposal processes:
https://www.python.org/dev/peps/pep-0001/
https://eips.ethereum.org/EIPS/eip-1

emg110:
May I suggest (for contents of ARC0) a simple diagram which determines and illustrates roles of `ARC`, `ARC PR` & `ARC ISSUE` and a **simple bulleted instruction** on how to contribute and on which `ground rules` basis.

fabrice102:
We should work on that.

In the meantime, let's keep things very simple. If you want to make an ARC:
1. Get support from the community on Discord/forum.algorand.org
2. Make a PR with a name: "New ARC: ARC-xxxx: title", where `xxxx` is replaced by the next available number. The PR should contain a single file `ARCs/arc-xxxx.md` which is formatted according to https://eips.ethereum.org/EIPS/eip-1 except for the header that should follow what is done in ARC-0001.

Discussion will happen on the PR which is simpler than on a separate issue (this is something we realized too late after making those issues).

We strongly encourage following the conventions:
> The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).

fergalwalsh:
Thanks for the info @fabrice102 👍

SudoWeezy:
[...]

PR #19, opened 2021-09-14
New ARC-0013

scholtz:
I request comment for ARC-0013 - Community Algorand Standards

### DIFF
--- a/ARCs/arc-0013.md
+++ b/ARCs/arc-0013.md
@@ -1,0 +1,96 @@
+---
+aip: 13
+title: Community Algorand Standards
+status: Draft
+---
+
+# Community Algorand Standards
+
+## Summary
+
+This document introduce conventions for central point for publishing community algorand standards.
+
+## Abstract
+
+The algorand blockchain is decentralized in the form that any person can create wallet, and use its assets for his or her own good. There are many methods how to store data to the blockchain, for example the ARC-0002 standard using the transation note field, or by creating the referenced document, or by providing data through the smart contracts. 
+
+Developers can create their own standards how they store or exchange data between applications.
+
+This ARC aims to bring together the developers to the central point of knowledge interchange and guides them how to create standards.
+
+## Specification
+
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
+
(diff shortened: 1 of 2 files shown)


ARCs/arc-0013.md
```diff
@@ -0,0 +1,96 @@
...
+
+Living - A special status for standard that are designed to be continually updated and not reach a state of finality.
+
+### Algorand foundation role
+
+Algorand foundation manages list of standards provided by the community and assigns them a standard identifier.
+
+### List of the Community Algorand Standards
+
+List of standards is available here: https://github.com/algorandfoundation/ARCs/blob/main/ARCs/CAS.json
```
barnjamin, line 57: this link 404s

barnjamin:
This is an interesting idea for having more machine readable ARC documents. 

As a first comment I think I'd prefer the URL point to the spec document _in_ this repo. 

Also maybe a PullRequest type field that points to the actual PR containing the proposal. 

Are you expecting that 2 prs are filed? 1 for the json object and 1 for the spec document?

ARCs/arc-0013.md
scholtz: it is 404 because the file is in this pull request

scholtz:
lets discuss the topic here please: https://github.com/algorandfoundation/ARCs/issues/20

After this pr will be committed, we will loose the discussion if we discuss it here.

Please note that proposed CAS.json file is included in this PR https://github.com/algorandfoundation/ARCs/pull/19/files

barnjamin:
>After this pr will be committed, we will loose the discussion if we discuss it here.

The conversation will persist in the same was as an issue. It may be closed but will still be visible. 

I'd prefer comments in the PR because any notes on the spec file itself will be inline as well.

jeremyv3:
I'm not clear on the necessity of this proposal at this time.

1) I wasn't able to find any example references that have defined such an ARC schema.  Do you have any example references to share to see what others have done for this? e.g. Ethereum, Python

2) It seems like Fergals original request to define ARC formatting on Algorand should be resolved before a standard JSON schema for ARC submittals is defined so that we know whether there are additional properties that are needed.
https://github.com/algorandfoundation/ARCs/issues/9

3) Where or in what process would this schema be used? When submitting to github? If so where?

4) Is there a need in the community at large for this? How great?

5) Does this proposal mean to suggest by the 'url' property name that the contents of the ARC requests could be kept outside of Algorand's github? I believe ARC's should remain on Algorand's github in all cases for historic purposes.

To me defining an official Algorand schema for this at this time seems premature and rushed before a wider community need for a schema of this type has been defined and established.

SudoWeezy:
Closed in favor of [ARC-0](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0000.md)

PR #21, opened 2021-09-15
ARC-0003: Support for relative URI

fabrice102:
As mentioned by @barnjamin in https://github.com/algorandfoundation/ARCs/issues/3#issuecomment-913177134, support for relative URI seems necessary to allow storing the JSON Metadata file together with the files it refers to in the same IPFS directory.


fabrice102:
No comment so merging.

PR #24, opened 2021-09-21
Require mimetype for animation_url

runvnc:
I don't know why it says the copyright line changed.. because it did not.


fabrice102:
ARC-3 was finalized but now there is some text recommending mimetype:
https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0003.md#mime-type-files
as per the discussion in #3

PR #28, opened 2021-09-21
ARC-0001: Support for stxn field

fabrice102:
Adding stxn field following comment
https://github.com/algorandfoundation/ARCs/issues/1#issuecomment-895374290

from @DanBurton


jeapostrophe:
Awesome!

PR #29, opened 2021-09-21
ARC-0001: Renaming signTxn to signTxns

fabrice102:
Following comment
https://github.com/algorandfoundation/ARCs/issues/1#issuecomment-895374290
from @DanBurton

This PR is to be dicussed.
`signTxns` is indeed more logical.

`signTxn` can still exist and be mapped to `signTxns` for backward
compatibility.

So I don't think this should create too many issues
but please raise your voice if it does.


jeapostrophe:
Awesome!

PR #33, opened 2021-10-17
Support of #arc3 in URL, MIME, CORS

fabrice102:
This PR applies the proposed changes in
https://github.com/algorandfoundation/ARCs/issues/3#issuecomment-930630064

as well as in:
https://github.com/algorandfoundation/ARCs/issues/3#issuecomment-926259384
https://github.com/algorandfoundation/ARCs/issues/3#issuecomment-926590736
(for MIME types)

More precisely:
* allow to use `#arc3` in URL
* explicitly indicate that MIME types should not be relied upon from a
security point of view
* make MIME types STRONGLY RECOMMENDED

### DIFF
--- a/ARCs/arc-0003.md
+++ b/ARCs/arc-0003.md
@@ -1,7 +1,7 @@
 ---
 arc: 3
 title: Algorand Standard Asset Parameters Conventions for Fungible and Non-Fungible Tokens
-status: Last Call
+status: Final
 ---
 
 # Algorand Standard Asset Parameters Conventions for Fungible and Non-Fungible Tokens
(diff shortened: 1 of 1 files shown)


ARCs/arc-0003.md
```diff
@@ -29,18 +29,30 @@ The ASA parameters should follow the following conventions:
...
-    * **SHOULD NOT** use the following URI scheme: *http* (due to security concerns).
-    * **MAY** be a relative URI when inside the JSON Metadata file. In that case, the relative URI is relative to the Asset URL. The Asset URL **SHALL NOT** be relative. Relative URI **MUST** not contain the character `:`. Clients **MUST** consider a URI as relative if and only if it does not contain the character `:`.
+    * (**RECOMMENDED**) or `<name>` where `<name>` is defined as above. In this case, the Asset URL **MUST** ends with `#arc3`.
+* *Asset URL* (`au`): a URI pointing to a JSON Metadata file. 
+    * This URI as well as any URI in the JSON Metadata file:
+        * **SHOULD** be persistent and allow to download the JSON Metadata file forever.
+        * **MAY** contain the string `{id}`. If `{id}` exists in the URI, clients **MUST** replace this with the asset ID in decimal form. The rules below applies after such a replacement.
+        * **MUST** follow [RFC-3986](https://www.ietf.org/rfc/rfc3986.txt) and **MUST NOT** contain any whitespace character
+        * **SHOULD** use one of the following URI schemes (for compatibility and security): *https* and *ipfs*:
```
pbennett, line 42: I would like to see acknowledgement of schemes like ipns:// - this was would allow mutable metadata but still all be backed by IPFS.

pbennett (changes requested):

ARCs/arc-0003.md
pbennett: I just wouldn't want to define a metadata hash as it wouldn't be fixed.  The URL would point to an ipns link and the ARC3 metadata would be dereferenced via ipns [dnslink] -> ipfs CIDs.  This data would intentionally be mutable but being stored via ipfs would have decentralization benefits.
Personally, I think if ARC3 is going to draw a line in the sand that the metadata must be completely immutable I think that's an issue for some use cases.

PR #39, opened 2021-10-28
ARC-0000 : Definition and Guidelines of the ARC submission processes

barnjamin:
Heavily borrowed from [eip-1](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1.md) with some edits and omissions. 
 Namely, I chose to remove the components that dealt with core protocol changes.

There are several comments with TODOs for things I'm not sure how we should deal with.

I welcome any and all comments on this pr.


SudoWeezy:
I am thinking about moving ARC-0 to the status *Living*.

The main idea is to standardize the flow of Creating and following the life of an ARC. 

One goal is to avoid having too many "Issues" in the repo.
> (We have more issues in our repo than they have in the Ethereum one)
Requesters should debate ideas in the Discord #ARC Channel or the forum. Then if the idea has been vetted, they will make ARC submissions through a pull request.
I only want to use the issues section to discuss ARC with status *Draft/Last Call/Living*.

Then I want ARCs that pass requirements to obtain the status *Draft* and be merged into the repo. Even if they are far from the *final status*, it might incentivize more people to participate.

Only ARC editors will choose the ARC's number when merging in the repo to avoid random numbers.

Please give me your thought on this.

PR #41, opened 2021-11-09
ARC-0014 : Authorization and Authentication using transactions

scholtz:
Please comment on ARC-0014 - Algorand Standard for authentication

### DIFF
--- a/ARCs/arc-0014.md
+++ b/ARCs/arc-0014.md
@@ -1,0 +1,67 @@
+---
+ams: 14
+title: Algorand standard for Authentication
+status: Draft
+---
+
+# Algorand Standard for authentication
+
+## Summary
+
+This document introduces the standard for authentication using the algorand accounts.
+
+## Abstract
+
+The goal of this standard is to define authorization process and authentication procedures for communication between web application and backend services or between two backend services. 
+
+## Specification
+
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+Base64 is the standard described in [RFC-4648](https://www.ietf.org/rfc/rfc4648.txt).
+
+## Request for comments
+
+Please comment or suggest pull request for this document here: [https://github.com/scholtz/AMS/issues](https://github.com/scholtz/AMS/issues)
+
+### Authentication
+
+Authentication is process in which person generates the authorization token for communication with backend service.
+
+Token MUST be generated from Algorand public and private key - from algorand address.
+
+In the first step of the authentication, the message to be signed is generated. The message MUST contain note field with configurable Realm. Realm is identifier of the service. It MAY be the web address of the service for intended purpose.
+
+The message MUST be self signed - the receiver is the same as the sender.
+
+The message MUST have network parameters filled in. The network genesis hash SHOULD be configured for validation pursposes by backend service. The latest block is the time of expiration of the token. The latest block MUST be provided.
+
+In the second step, the message MUST be signed by the account private key.
+
+The message MUST not be submitted to the network (mainnet, testnet) nor should be published in any other way. It may be used only for communcation to the backend service.
+
+### Authorization
+
+Authorization determines what resources a user can access.
+
+Authorization is the process of validating authorization token by the backend application and allowing the account permissions to the backend services.
+
+Authorization token SHOULD be send to backend service using the authorization header.
+
+Header data MUST start with prefix "SigTx "
+
+Authorization message must follow prefix in Base64 encoding.
+
+Example of the HTTP header:
+
+```
+Authorization: SigTx gqNzaWfEQJ4FWNWiXuRz5DKu1RYL5qHlR+iP/3qW4BF+pPD/ok20tJSqBICQn2jWysFD88W3a0ojEBM+IWvh5tyfvZyZ+AKjdHhuiaNmZWXNA+iiZnbOAQx8LaNnZW6sdGVzdG5ldC12MS4womdoxCBIY7UYpLPITsgQ8i1PEIHLD3HwWaesIN7GL39w5Qk6IqJsds4BDIAVpG5vdGXEEURSRU0tQXV0aGVudGljYXRlo3JjdsQgG1z5khU3SjAofF/H7uWij05Nzy1ZVn2sYVEzIHauIAWjc25kxCAbXPmSFTdKMCh8X8fu5aKPTk3PLVlWfaxhUTMgdq4gBaR0eXBlo3BheQ==
+```
+
+Service MUST return 401 response if the service is configured to check for validity of the expiration of the token and the current block at the specified network is higher. Service SHOULD be configured to check for validity of the expiration of the token.
+
+Service MUST return 401 response if the token data is not Base64 valid data.
+
+Service MUST return 401 response if the token data cannot be parsed to Algorand signed transaction.
+
+Service MUST return 401 response if the signed transaction has invalid signature.


ARCs/arc-0014.md
```diff
@@ -0,0 +1,67 @@
...
+
+Base64 is the standard described in [RFC-4648](https://www.ietf.org/rfc/rfc4648.txt).
+
+## Request for comments
+
+Please comment or suggest pull request for this document here: [https://github.com/scholtz/AMS/issues](https://github.com/scholtz/AMS/issues)
+
+### Authentication
+
+Authentication is process in which person generates the authorization token for communication with backend service.
```
barnjamin, line 29: will it always be a backend service?

ARCs/arc-0014.md
```diff
@@ -0,0 +1,67 @@
...
+
+Token MUST be generated from Algorand public and private key - from algorand address.
```
barnjamin, line 31: from algorand address seems redundant here

ARCs/arc-0014.md
```diff
@@ -0,0 +1,67 @@
...
+
+In the first step of the authentication, the message to be signed is generated. The message MUST contain note field with configurable Realm. Realm is identifier of the service. It MAY be the web address of the service for intended purpose.
```
barnjamin, line 33: I like this idea.

ARCs/arc-0014.md
```diff
@@ -0,0 +1,67 @@
...
+
+The message MUST be self signed - the receiver is the same as the sender.
```
barnjamin, line 35: lets define this as a pay transaction, of 0 amount, from the authenticator to the authenticator (or a better word?) with the note field populated with the above and the rest of the parameters below. like a template transaction 
```
txn {
...
}
```

ARCs/arc-0014.md
```diff
@@ -0,0 +1,67 @@
...
+
+The message MUST have network parameters filled in. The network genesis hash SHOULD be configured for validation pursposes by backend service. The latest block is the time of expiration of the token. The latest block MUST be provided.
+
+In the second step, the message MUST be signed by the account private key.
+
+The message MUST not be submitted to the network (mainnet, testnet) nor should be published in any other way. It may be used only for communcation to the backend service.
```
barnjamin, line 41: Maybe we can make it such that the transaction will fail, so this is not an issue.

Set fee to 0?

ARCs/arc-0014.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Authorization
+
+Authorization determines what resources a user can access.
+
+Authorization is the process of validating authorization token by the backend application and allowing the account permissions to the backend services.
+
+Authorization token SHOULD be send to backend service using the authorization header.
```
barnjamin, line 49: Think we should not define it as a "backend service" but any verifier. The verifier could be a backend service or some front end protected service

barnjamin (commented):
I like the idea, seen a number of people comment looking for something like this and I'd like to provide a standard way of doing it. 

Posted some thoughts

ARCs/arc-0014.md
barnjamin: maybe define a standard scheme for this so that wallets can let the users know its a "sign in" type request

ARCs/arc-0014.md
wheresaddie: yes, I agree the param for this should be verifier as it allows the definition to vary from backend or front end sec

ARCs/arc-0014.md
scholtz: The frontend creates this token, so it should not need authentication.. But if two people communicate at FE and somehow interchange the tokens, this might proof the authenticity of the other person..

ARCs/arc-0014.md
scholtz: i like the idea of having fee below the acceptance on the network.. perhaps 1 should be also good

barnjamin:
@scholtz can you update with whatever edits you'd like from the above suggestions then mark them as resolved. 

@fabrice102  when you get a sec, please review after updates

wheresaddie:
lgtm -- will defer to @fabrice102 to PR

ARCs/arc-0014.md
urtho: If the note field is to be used for Realm then where one should put nonce (to avoid replay attacks)

ARCs/arc-0014.md
urtho: IMHO there should be some "stronger" violation than a low tx fee - one that survives future changes to the network.

urtho: also, I would love to see two signatures :) 

The first one without any nonce - so I could use the sig as a "password" for end2end or local storage encryption. 
The second one, with a nonce, for authentication .

Maybe, in my case, I would send a group of two "authentication" transactions in one go.  
One with all fields static (no nonce, fv=lv=0) and one with fv=lv=nonce

ARCs/arc-0014.md
barnjamin: the note field may still contain a nonce or any other unique data to ensure uniqueness

ARCs/arc-0014.md
barnjamin: > IMHO there should be some "stronger" violation than a low tx fee - one that survives future changes to the network.

setting the block heights to something in the past or last round < first round in the transaction should ensure that the transaction _cannot_ be accepted by the chain

ARCs/arc-0014.md
urtho: How about FE creates the auth txn and FE wallet signs it (one-click login)

ARCs/arc-0014.md
urtho: > 

How about this scheme so that we can have a nice , extensible, wallet "sign-in"

auth/v1:j{"realm":"MetaHealth","notice":"You are signing-in to MetaHealth platfrom","nonce":"5aa15e8fa86ecd71455c56871c328070"}

with auth/v1 defined as

interface authInfoV1 {
  realm: string,
  notice?: string,
  nonce?: string,
}


Sample typescript code:
https://gist.github.com/urtho/610fb4e891f665eed2e7955e0fb3e74e

ARCs/arc-0014.md
fabrice102: I think forcing the genesis hash and genesis ID to be some specific values should prevent posting the transaction whatever happen.
In addition fee=0, fv=lv=0 looks like an additional good idea.
And everything else in the note field

fabrice102:
I am sorry for the very late look at this. @scholtz @barnjamin @wheresaddie 

Thank you very much for this proposal.

After discussion with @shaih, to better understand how to ensure security, we have a few questions:

## Use cases

0. Is the authentication mechanism meant to be used in a browser with a wallet such as AlgoSigner/WalletConnect/MyAlgoConnect? Or is it also meant to be used in standalone apps?

1. How do you plan to use such an authentication/authorization mechanism?
1a. Is it expected to be a way to "sign in" to a website as OAuth/SAML?
1b. Is there any goal of creating a secure channel between the wallet or the browser and the website?
1c. Is there a need for mutual authentication?

## Design goals

### Compatibility

2. I guess one design goal is to allow all current devices to work, including Ledger. If it is not the case, it would be easier to just sign arbitrary message with a fresh prefix. Is my understanding correct?

### Interactivity

3. Is the absence of interactivity, that is the fact that the website requires the authentication and the wallet answers with an authentication token necessary? Adding interactivity can provide higher security mechanism

### Support for non-ed25519 keys

4. What about multisig?
5. What about logicsigs? Allowing them may be dangerous as some escrow accounts may allow signing the proposed transactions.

## About the authorization part

6. Is the wallet expected to handle this? I am thinking of the example of a dApp using e.g. AlgoSigner. Most likely AlgoSigner would just return the signed transaction. In any case it would not be able to add headers to queries. Do you think this part is important to standardize or it can be left to the ARCs user to decide on the implementation?

## Comments on security

The main kind of attacks we can see is man-in-the-middle.
Suppose you create a dApp A that needs such an authentication.
An attacker can create a dApp B that looks legit (and actually does legit things, say for example compounding rewards).
Then the attacker asks the user to sign an authentication transaction for A.
Most users U will not even read the note and allow B to access A as U.

One solution is to ask the realm to be a domain name and to require the wallet to check that the domain matches the domain that asked for the transaction to be signed (and that the website is in https, otherwise there is no security).

Another risk is re-use of an authentication token.
The nonce should prevent that though.

## Comments on impossiblity to run transactions in the network

I think one of the simplest solution is just forcing the genesis hash and genesis name to a specific value.
For example:
* genesis hash = SHA-512/256 of the string ARC-0014-authentication
* genesis name = ARC-0014-authentication

Such transactions will never be played on mainnet.
Other fields may be forced to specific values as proposed.

One thing to check is whether Ledger Nano would sign such transactions.

barnjamin:
> Is the authentication mechanism meant to be used in a browser with a wallet such as AlgoSigner/WalletConnect/MyAlgoConnect? Or is it also meant to be used in standalone apps?

I take it to mean something where any of the wallets should allow this transaction to be signed and not fail for one of the issues that makes it a technically invalid transaction. A Verifier may inspect the transaction and signature passed back to validate the secret holder.

>How do you plan to use such an authentication/authorization mechanism?
1a. Is it expected to be a way to "sign in" to a website as OAuth/SAML?

Or be used physically to validate someone physically enter an event for example?

> 1b. Is there any goal of creating a secure channel between the wallet or the browser and the website?

The secure channel should be left to https for example

> 1c. Is there a need for mutual authentication?

Not in the cases I'd considered

>I guess one design goal is to allow all current devices to work, including Ledger. If it is not the case, it would be easier to just sign arbitrary message with a fresh prefix. Is my understanding correct?

Many wallets do not support arbitrary bytes, yes this is a sort of "hack" to get around this.

>Is the absence of interactivity, that is the fact that the website requires the authentication and the wallet answers with an authentication token necessary? Adding interactivity can provide higher security mechanism

Can you give an example?

>Support for non-ed25519 keys
What about multisig?
What about logicsigs? Allowing them may be dangerous as some escrow accounts may allow signing the proposed transactions.

I think this would only be useful in the case of a single account and only for ed25519 signatures unless we start supporting another one.


> Is the wallet expected to handle this? I am thinking of the example of a dApp using e.g. AlgoSigner. Most likely AlgoSigner would just return the signed transaction. In any case it would not be able to add headers to queries. Do you think this part is important to standardize or it can be left to the ARCs user to decide on the implementation?

I think we should be fine without headers, the txn preparer should have enough to verify the signature and take further action.


>One solution is to ask the realm to be a domain name and to require the wallet to check that the domain matches the domain that asked for the transaction to be signed (and that the website is in https, otherwise there is no security).

Good point. Lets add this

>I think one of the simplest solution is just forcing the genesis hash and genesis name to a specific value.

I don't think the mobile wallet supports this either at the moment. The use of fv/lv/fee=0 seems like thinks we should recommend either way.

We should ask that wallets recognize this type of transaction though and prompt that it constitutes a sign in request from  the domain.

fabrice102:
Additional question I realized I forgot:

What about rekeying? What should happen in that case?
Should the original address be used even if rekeying occurred?

barnjamin:
> What about rekeying? What should happen in that case?
Should the original address be used even if rekeying occurred?

I think we should allow the authorizing address to sign for the transaction but leave it up to the application to decide if that is acceptable for their use case.

ARCs/arc-0014.md
scholtz: the real use case is when you want to store some data to know who created them.. to visualize the authenticated user addres with one click login purely on fe does not make a lot of sense for me, but sure the service is not be limitted

ARCs/arc-0014.md
scholtz: i like this idea, even though if user would be able to sign the transaction with replay attack, something is wrong elsewhere.. the uniqueness of the tx is guaranteed by the algorand tx id, but if you guys thing we need to put more information to the message lets do it

ARCs/arc-0014.md
scholtz: i would not do it as a must.. the authenticator does not have to have wish to reveal his address. also he might want to use it in different domains.. eg google authentication service is on one domain, and it serves multiple other hosts backends.. the only thing that the backend has to verify when writing some user data that the realm is within its scope.. if the realm will be in the note or as a receiver, it is up to the user. if you guys believe we should change MUST to SHOULD go on..

ARCs/arc-0014.md
scholtz: to know the genesis of the network is crucial to check the expiration of the token
keep in mind that you send this token with each request to the backend, as the jwt works.. so we should keep the size to minimum
i think setting the fee to zero is the nicest solution never to add the tx to the network

joe-p:
I noticed that activity on this ARC has become stagnant, but I think this is a very important standard to reach and I'll explain why down below. I do, however, have some nitpicks with this specific proposal. 

With any ARC, we need to identify what the goal we're trying to accomplish is and then identify what standards need to be in place to accomplish said goal. 

With authentication, the main problem is a UX one. It's easy enough to sign and very arbitrary bytes with Algorand's SDKs and ed25519 libraries. The hard part is clearly telling the user that they are signing data for verification purposes for a specific service. Given the tools we have available now, the most reasonable way for an application to authenticate with a wallet is to have the user sign an arbitrary transaction that is never submitted to the network. Needless to say, this is not intuitive for anyone that isn't familiar with the Algorand development ecosystem. 

That being said, this ARC seems to address some issues not related to the core UX problem and it's still fairly unclear what exactly we'd want from wallets in terms of supporting an authentication standard. Below are my criticisms of this ARC as it stands and some of my proposals. 

# Criticisms 

## SigTx header
 
Why try to implement a new standard (the `SigTx` HTTP header) when we already have well-adopted standards for authorization and session management? As I mentioned early, the real problem we should be trying to address here is the UX of authentication. Once a user is verified to own a specific address, we should not try to dictate what the service does with that information, especially not if it'd require a new non-standard HTTP header. Once we have a clear standard for verifying wallet ownership (authentication), the service should be able to use whatever available methods are appropriate for authorization and session management. For example, once a client authenticates via verifying the ownership of a wallet, the server can distribute a JWT or cookie to manage the session and handle authorization. 

## Signed Content

This ARC and pursuant discussion seems to be focused on signing a full transaction for the sake of authentication. I understand that this is what makes the most sense *right now* due to constraints in most wallets (not being able to sign arbitrary bytes), but that should not limit what we are calling for in a standard. It is possible for wallets to sign arbitrary bytes, so why create a standard around signing transactions when it would certainly make more sense to sign a authentication-specific message. Yes, this would require development time on the wallet's end, but having a clear standard to support will be better for the UX. You could make the argument that the wallet UI could just look at the transaction field of a signed transaction to determine authentication information, but at that point if development effort is being spent parsing transactions why not just implement a authentication-specific standard. 

# Proposals

## What to Sign

Right now, the standard for signing arbitrary messages with the SDKs is to prepend "MX" to the bytes being signed for domain separation. My proposal is that we implement a auth-specific prefix, such as "AUTH" followed by a payload containing the data being signed. The contents of the payload should at the very least contain the following information:

* Service the authentication is for
* A description of the service and why it needs authentication
* Wallet Address
* Nonce

Once this data is signed and verified, the secure service can then handle sessions and authorization however it sees fit (JWT, cookies, etc.). This means we don't need a ton of session metadata baked into the authentication payload itself. 

## Verification Flow

The client should initiate the verification flow on their side

1. Client initiates verification by sending service their wallet address
2. Secure service generates authentication payload and sends it to the client
3. Client signs AUTH payload and sends back signature
4. Secure service verifies signature against the AUTH payload and public key

## User Experience

Step three of the above, the signing of the AUTH payload, will likely be done via some sort of wallet-connection service (WalletConnect, MyAlgo Connect). When prompting the user for the AUTH payload to be signed, the following information should be displayed

* Service the authentication is for
* A description of the service and why it needs authentication
* Wallet Address

# Required Development

At the minimum, for this standard to be considered adopted we should expect the following services to provide the above user experience. This means support for signing the AUTH payload and display the payload information. 

* MyAlgoConnect
* WalletConnect
* Ledger

Without the support of these major wallets/connections services, it's hard to say the standard has been adopted and the user experience will be fragmented. 

# Temporary Standard

If we wanted to come up with a standard that can be adopted *right now* we'd need to follow some transaction-based standard that follows a similar for to the aforementioned flow, expect instead of an AUTH payload it's a transaction with the payload information in the note. This transaction should be invalid and to/from the address being authenticated. We should keep in mind, however, that a standard like this doesn't really solve the main problem at hand, which is the user-experience. It's only a temporary standard that developers can follow while a more suitable standard is being adopted.

0x9090:
There is an Algorand Authentication PoC here that seems close to the previous proposal - https://github.com/NullableLabs/AlgoAuth

ARCs/arc-0014.md
```diff
@@ -0,0 +1,67 @@
...
+
+The message MUST be self signed - the receiver is the same as the sender.
+
+The message MUST have network parameters filled in. The network genesis hash SHOULD be configured for validation pursposes by backend service. The latest block is the time of expiration of the token. The latest block MUST be provided.
+
+In the second step, the message MUST be signed by the account private key.
+
+The message MUST not be submitted to the network (mainnet, testnet) nor should be published in any other way. It may be used only for communcation to the backend service.
+
+### Authorization
```
FrankSzendzielarz, line 43: Given the Arc is for an authentication method I think this Authorization section is redundant. If I was implementing a web or mobile app I would probably stick with some flavour of OpenID connect and have Identity Provider accept a signed nonce.

ARCs/arc-0014.md
```diff
@@ -0,0 +1,67 @@
...
+
+Authorization message must follow prefix in Base64 encoding.
+
+Example of the HTTP header:
+
+```
+Authorization: SigTx gqNzaWfEQJ4FWNWiXuRz5DKu1RYL5qHlR+iP/3qW4BF+pPD/ok20tJSqBICQn2jWysFD88W3a0ojEBM+IWvh5tyfvZyZ+AKjdHhuiaNmZWXNA+iiZnbOAQx8LaNnZW6sdGVzdG5ldC12MS4womdoxCBIY7UYpLPITsgQ8i1PEIHLD3HwWaesIN7GL39w5Qk6IqJsds4BDIAVpG5vdGXEEURSRU0tQXV0aGVudGljYXRlo3JjdsQgG1z5khU3SjAofF/H7uWij05Nzy1ZVn2sYVEzIHauIAWjc25kxCAbXPmSFTdKMCh8X8fu5aKPTk3PLVlWfaxhUTMgdq4gBaR0eXBlo3BheQ==
+```
+
+Service MUST return 401 response if the service is configured to check for validity of the expiration of the token and the current block at the specified network is higher. Service SHOULD be configured to check for validity of the expiration of the token.
```
FrankSzendzielarz, line 61: This assumes the service being connected to is aware of the Algo chain. It would be nice to use Algo tools to sign in to off chain web apps.

ARCs/arc-0014.md
FrankSzendzielarz, line 29: I find the terminology confusing. In authorisation code flow authentication typically yields an Identity Token and a code, which can ultimately be exchanged for Access Tokens. In my mind authentication and authorisation are unrelated.

ARCs/arc-0014.md
FrankSzendzielarz, line 41: Seeing as the token's intended purpose is being conflated with server access token, where is the token expiration (we cant assume the server must be algo chain aware) , and how does the front end reacquire fresh access tokens ? For this protocol to work with mobile devices it would probably need to consider device unique id as part of the token, as there is no equivalent of a refresh token, to handle multi device global sign out, lost device logout, etc

FrankSzendzielarz:
> I noticed that activity on this ARC has become stagnant, but I think this is a very important standard to reach and I'll explain why down below. I do, however, have some nitpicks with this specific proposal.
> 
> With any ARC, we need to identify what the goal we're trying to accomplish is and then identify what standards need to be in place to accomplish said goal.
> 
> With authentication, the main problem is a UX one. It's easy enough to sign and very arbitrary bytes with Algorand's SDKs and ed25519 libraries. The hard part is clearly telling the user that they are signing data for verification purposes for a specific service. Given the tools we have available now, the most reasonable way for an application to authenticate with a wallet is to have the user sign an arbitrary transaction that is never submitted to the network. Needless to say, this is not intuitive for anyone that isn't familiar with the Algorand development ecosystem.
> 
> That being said, this ARC seems to address some issues not related to the core UX problem and it's still fairly unclear what exactly we'd want from wallets in terms of supporting an authentication standard. Below are my criticisms of this ARC as it stands and some of my proposals.
> 
> # Criticisms
> ## SigTx header
> Why try to implement a new standard (the `SigTx` HTTP header) when we already have well-adopted standards for authorization and session management? As I mentioned early, the real problem we should be trying to address here is the UX of authentication. Once a user is verified to own a specific address, we should not try to dictate what the service does with that information, especially not if it'd require a new non-standard HTTP header. Once we have a clear standard for verifying wallet ownership (authentication), the service should be able to use whatever available methods are appropriate for authorization and session management. For example, once a client authenticates via verifying the ownership of a wallet, the server can distribute a JWT or cookie to manage the session and handle authorization.
> 
> ## Signed Content
> This ARC and pursuant discussion seems to be focused on signing a full transaction for the sake of authentication. I understand that this is what makes the most sense _right now_ due to constraints in most wallets (not being able to sign arbitrary bytes), but that should not limit what we are calling for in a standard. It is possible for wallets to sign arbitrary bytes, so why create a standard around signing transactions when it would certainly make more sense to sign a authentication-specific message. Yes, this would require development time on the wallet's end, but having a clear standard to support will be better for the UX. You could make the argument that the wallet UI could just look at the transaction field of a signed transaction to determine authentication information, but at that point if development effort is being spent parsing transactions why not just implement a authentication-specific standard.
> 
> # Proposals
> ## What to Sign
> Right now, the standard for signing arbitrary messages with the SDKs is to prepend "MX" to the bytes being signed for domain separation. My proposal is that we implement a auth-specific prefix, such as "AUTH" followed by a payload containing the data being signed. The contents of the payload should at the very least contain the following information:
> 
> * Service the authentication is for
> * A description of the service and why it needs authentication
> * Wallet Address
> * Nonce
> 
> Once this data is signed and verified, the secure service can then handle sessions and authorization however it sees fit (JWT, cookies, etc.). This means we don't need a ton of session metadata baked into the authentication payload itself.
> 
> ## Verification Flow
> The client should initiate the verification flow on their side
> 
> 1. Client initiates verification by sending service their wallet address
> 2. Secure service generates authentication payload and sends it to the client
> 3. Client signs AUTH payload and sends back signature
> 4. Secure service verifies signature against the AUTH payload and public key
> 
> ## User Experience
> Step three of the above, the signing of the AUTH payload, will likely be done via some sort of wallet-connection service (WalletConnect, MyAlgo Connect). When prompting the user for the AUTH payload to be signed, the following information should be displayed
> 
> * Service the authentication is for
> * A description of the service and why it needs authentication
> * Wallet Address
> 
> # Required Development
> At the minimum, for this standard to be considered adopted we should expect the following services to provide the above user experience. This means support for signing the AUTH payload and display the payload information.
> 
> * MyAlgoConnect
> * WalletConnect
> * Ledger
> 
> Without the support of these major wallets/connections services, it's hard to say the standard has been adopted and the user experience will be fragmented.
> 
> # Temporary Standard
> If we wanted to come up with a standard that can be adopted _right now_ we'd need to follow some transaction-based standard that follows a similar for to the aforementioned flow, expect instead of an AUTH payload it's a transaction with the payload information in the note. This transaction should be invalid and to/from the address being authenticated. We should keep in mind, however, that a standard like this doesn't really solve the main problem at hand, which is the user-experience. It's only a temporary standard that developers can follow while a more suitable standard is being adopted.



> 

I am leaning this direction too. Feels to me like we just need an Identity Provider extensiin to oauth2 authorisatiin code grant flow or some oidc derivative

fionnachan:
Agree with @joe-p.

Authentication of wallet ownership is possible by sending a valid or invalid transaction at the moment but it's terrible UX. Users should not be expected to send any sort of payment transactions (even if expected to fail) just to validate their identity.

The original proposal describes a technical flow created under current limitations but since the current possible ways to authenticate do not communicate well the intention to authenticate (because they are not a specific type of transaction dedicated to authentication), I think the current possible methods (incl the original content of this proposal) should not become a standard. At best, it should just be a recommended workflow until the arrival of a solution that can communicate authentication seamlessly to normal users.

FrankSzendzielarz:
Agree with @fionnachan . I would go a step further and actively prevent transactions being used or encouraged for this purpose at all, out of security concerns. I can envisage a scenario where users are given payloads to sign by malicious apps that replace auth payloads with payments for example. The WalletConnect specs could be modified to support auth payloads that explicitly prohibit decoding into Transaction objects.

FrankSzendzielarz:
For the particular scenarios described in the ARC and in @joe-p 's proposals, I wonder why we are trying to reinvent parts of the OIDC and Oauth2 protocol specs. For example, if we implemented WalletConnect at a "Sign In with Algorand" identity provider , then the PKCE (_code challenge_) and _state_ request parameters already take care of the concept of nonce above. The realm and _scopes_ are already displayed on the login page most likely, so the identity provider already has a large payload it uses to identify the request and why. Even then, it's up to that server if to use that as the payload or a one off GUID that represents that authN request. That server (optionally - depends on SSO and if a cookie is present etc) initiates a WalletConnect session by displaying a QR code - at this point there is a vulnerability where a victim could be persuaded to part with their unlocked phone for a short while, so it would be to good make sure any authorisation process involved PIN/Biometric confirmation every time - which would mean that the Algorand customisations to the WalletConnect protocol _allow arbitrary non transaction payloads for signing_. I would also propose that the auth spec and wallet implementations explicitly prevent transaction objects from being allowed in the payload. In fact, I wonder if a safer and simpler UX would be a separate Algorand Authenticator app to help ensure that no one is persuaded to sign a payment when trying to authenticate.

AlgoDoggo:
Please don't make yet another authenticator app. You can simply decide on one transaction type that would be flagged by the wallet and redirect to a different UI.

Such as:
`if (!fee && !amount && from === to && firstRound === constant && lastRound === constant) => authUI`

Where constant would be a past round we all agree on. Domain and date/nonce are fine in the note field. Login expiry can be set in a jwt as usual.

When a transaction hit wallets with this round constant it would trigger a function that verifies the other params and redirects to the auth UI. The whole thing would work off chain and is very simple to implement.

FrankSzendzielarz:
Could have both. (separate authenticator app vs separate sets of authenticator screens in different wallet apps,)  But yeah.

I just don't like the idea of perverting the concept of "Transaction" , which implies some interaction with the blockchain, by twisting it towards the purpose of Authentication. Seems hacky to me.

AlgoDoggo:
Transaction or auth message it's really just about getting a hash and a sig.
Wallets would check the tx params under the hood, unbeknownst to users. They'd get a clean looking UI.
It would also be immediately compatible with ledgers and just about any wallet. Setting a round constant would remove the need to mess with genesis ID or hash which would be a red flag in some wallets.

Algorand compatible wallets could make an auth UI in no time by just watching for this given set of params. If Algorand switches to lattice based crypto or others it would also update seamlessly with the wallet.

scholtz:
@joe-p .. i think you misunderstood the intention of this proposal..

primary use case is to have algo aware Backend API, with minimum impact on wallets integration

i have coded already library that implements the standard as it was proposed. in .net person just install a nuget package and in controller setup the [Authenticated] attribtes for specific methods

it is not possible to make security on FE.. on FE you can just show who's account is user using, but real authentication must come when you load or store data to the database for example through apis.. 

if you are trying to undermine this proposal by stating that with ed25519 you can sign arbitrary data, i understand, but current implementation of all wallets makes it much effecient to use simple thing - signing and verifying of transactions

@0x9090 please checkout https://github.com/scholtz/AlgorandAuthenticationDotNet .. it is few months older

i see that in your repo you are trying to solve the issue Wallet <> FE .. are you storing some data to the database? are you using the cookie for it? Do you have wallet connect setup between BE and Wallet or FE and Wallet? What advantage does it have to use cookie for session handling than proposed Authentication SigTx [TxData]

joe-p:
> @joe-p .. i think you misunderstood the intention of this proposal..
> 
> primary use case is to have algo aware Backend API, with minimum impact on wallets integration

My comments on the scope of this ARC we're directly towards the initial proposal, but also on the discussion happening around it. Many people were discussing various information that could be added to the note field in addition to the "Realm". 

Also like I mentioned, I think the main goal of coming up with a standard is to create a good UX for authentication. Any improvement of UX would require changes with wallets and their respective bridges. 

> if you are trying to undermine this proposal by stating that with ed25519 you can sign arbitrary data, i understand, but current implementation of all wallets makes it much effecient to use simple thing - signing and verifying of transactions

Right, which is why I proposed that we could define a more temporary standard with the constraints of current wallet functionality in mind, but ideally we should have a more streamlined and domain-separated authentication standard. 

> i see that in your repo you are trying to solve the issue Wallet <> FE .. are you storing some data to the database? are you using the cookie for it? Do you have wallet connect setup between BE and Wallet or FE and Wallet? What advantage does it have to use cookie for session handling than proposed Authentication SigTx [TxData]

I know you're not asking me directly, but I am also using cookies to store session data.  The benefit of cookies over a custom header is that the browser will automatically send cookies with every request and update them after every response.

This leads into what I was saying about creating new standards. Why create a new one when we already have multiple standards in place? Some applications might use cookies, some might use JWT, some could use custom headers.  Also, not every application will use HTTP. I think we should have a non-protocol-specific standard for verifying ownership of a wallet. 

You are the one proposing a new standard here... so I think it's also important you expound upon the benefits of the proposed header over using already-existing standards.

barnjamin:
Looking at https://github.com/algorandfoundation/did-algo#322-proofs, it seems like exactly what we'd need to verify some account without resorting to having the user sign a transaction that happens to be invalid.

Should we try to put this in an ARC for wallets to support?

barnjamin:
Another defensive transaction configuration might be to set the `sender` on the transaction to the zero address and pass the account you want to auth for in the `signers` field

NBDanT:
As the topic looks to be considered above, ... is there a current/planned way that NFTs can be set to 'expire' at a future-time/block/etc interval? Application is NFT-gated access.  Thx!

scholtz:
Note that i have implemented the proposed ARC-0014 standard for graphql hasura server.

Docker image: https://hub.docker.com/r/scholtz2/hasura-algorand-auth-web-hook (best used in k8s)

Source code: https://github.com/scholtz/HasuraAlgorandAuthWebHook

(uses https://www.nuget.org/packages/AlgorandAuthentication/1.0.4 for auth)

ARCs/arc-0014.md
scholtz: this standard is meant to make authentication for backend, and backend will always be connected to internet and can verify the round att the blockchain

SudoWeezy:
Closing this in favor of https://github.com/algorandfoundation/ARCs/pull/84

PR #49, opened 2021-12-06
arc-0015 AlgoSMS - Algorand Standard for Encrypted Short Messages initial draft

urtho:
Initial draft, to be revised heavily. 
Example SDK will follow shortly.

### DIFF
--- a/ARCs/arc-0015.md
+++ b/ARCs/arc-0015.md
@@ -37,13 +37,13 @@ The **ALGOSMSV0_NOTE** object contains, among other fields, an encrypted UINT8Ar
 ```ts
 // UINT8Array ecrypted note contents as msgpack encoded object
 interface ALGOSMSV0_NOTE {
-	/** Object type eq to ALGOSMSV0_TYPE_NOTE **/
-	_t: string ;
+  /** Object type eq to ALGOSMSV0_TYPE_NOTE * */
+  t: string;
 
-	/** one-time public key of the NACL's sealed box **/
+  /** one-time public key of the NACL's sealed box * */
   otPK: Uint8Array;
 
-	/** cipherText of the ALGOSMSV0_MESSAGE structure **/
+  /** cipherText of the ALGOSMSV0_MESSAGE structure * */
   cT: Uint8Array;
 }
 ```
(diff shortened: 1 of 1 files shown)


fabrice102:
Thank you very much for your proposal!
Using the same key for decryption and signature is usually quite dangerous.
If you know an academic paper proving this is ok in that specific case, please let us know.
If not, we will need to write a proof that your specific instantiation is secure.
At first glance, it looks ok, but cryptography can quickly go wrong. (For example, using plain RSA for decryption and signature is completely insecure.)

Other points to analyze/discuss:
* How do you handle rekeying?
* How do you handle non-ed25519 public keys? You can easily see that some addresses are not ed25519 public keys as about 1/8 of possible addresses are valid ed25519 public keys. But with probability 1/8 a multisig or logicsig address may look like a valid ed25519 public key. Using the indexer can tell you what the address actually is, but only if the address transacted.
* Are there any addresses that are incorrectly generated and that would allow anyone to decrypt the ciphertext? At first glance if you filter for valid ed25519 public keys and the point at infinity, I think that should be good, but this need to be checked to.

urtho:
Thank You for finding the time to parse through this. :)

Am not a crypto guy but is the recent https://eprint.iacr.org/2021/509.pdf referenced @ https://libsodium.gitbook.io/doc/advanced/ed25519-curve25519 enough for this use-case ?

Good point on the non-ed25519 public keys. The standard should definitely forbid encryption if the destination is not a valid public key. Also, the main use-case is to replace notification emails in sites that do wallet-based (ARC-0014) logins instead of email-based logins. Users would use dedicated login-only accounts so we can focus on accounts that are valid ed25519.

Question: can a wallet (like AlgoSigner) generate an account that is not valid ed25519 public key (on purpose)?

As for the rekeyed accounts, we could just ignore them, and treat them the same as invalid ed25519, as they are mostly outside of the use-case. But let me sit on the rekeying issue as this is a common one with ARC-0014.

urtho:
Once an account is rekeyed, its private "spending" key should be considered forgotten.
This limits the use-case a lot :(

An encrypted message cannot be sent to an address if the address is:
- a multi-sig address
- a logicsig address
- rekeyed-to a multi-sig account :)
- rekeyed-to a logicsig 
- rekeyed-to a normal address, as the address can be the authorized one for many other addresses and we should not assume that the person authorized for spending is also authorized for reading private notes. 
- rekeyed-to a hardware wallet - probably next to impossible to get the hardware to support this. 

Q: Is there a way to check for 1:1 relation in rekey-to ?

cusma:
> Q: Is there a way to check for 1:1 relation in rekey-to ?

@urtho this [Issue](https://github.com/algorand/go-algorand/issues/3044) tracks how to expose `AuthAddr` to the AVM. In this way the SMS should be sent by an AlgoSMS dApp that could check if the receiver has been rekeyed or not. I think the AlgoSMS protocol should include a bit of TEAL logic in order to validate proper SMS receivers, approving or rejecting SMS sending.

scholtz:
1) if there would be information in /v2/account/{id} in algod or indexer about the current address that it is rekeyed to, this would solve the isusse also for arc0014 .. perhaps there should be also some information with which account it has been encoded if it is not the same as the original public address.. this however waits for algorand action.. @fabrice102 am i right?

2) @urtho would you mind to change it also for unencrypted messages? if i send a public message to other person using algorand network, it costs me 0.001 algo, and if it is my decision to make it public it may serve several use cases.. mainly something like governance chat, trading chat and so on..

3) have you give a thought to make it also external? if the encoded message is longer than note field or when algorand will decide to require higher fee if there is longer message, it might be good to have external capabilities.. for example you could store just ipfs link to the encoded message and if receiver wants to read it he downloads it

urtho:
ad. 1. The "spending" account info is there in the postgres DB on the indexer so no it "just" needs to be exposed.

ad. 2  Nice one. I'll upgrade the specs to include non encrypted texts. The minimum free was here to avoid spam. Any ideas how to allow for you use-case while making mass spamming cost prohibitive ?

ad. 3 There are some solutions already that do encrypted chain mail just by sending hashes and storing content on IPFS.  I wanted to start with simple SMS and possibly MMS (with image in URL and/or meta fields) but did not want to make a full-blown email solution out of it. Anyways it does not hurt the spec to have an additional field for IPFS-like content so I will do a revision this week with your suggestions. @scholtz

SudoWeezy:
@urtho If this PR is still relevant, please, reach out to me

scholtz:
i believe it is.. we are using similar encryption system in the vote coin standard

PR #53, opened 2021-12-07
Minor improvements to the ABI

jasonpaulos:
This PR contains a small set of improvements to the ABI. I've separated them into two categories:

### Technical changes
* Remove the `appId` field in favor of a new `networks` field on `Contract`. This change has two benefits:
  * A single contract JSON object can now represent a contract deployed to multiple networks, such as to TestNet and MainNet.
  * A contract JSON object can now describe a contract that has not yet been deployed by omitting the `networks` field. Previously the `appId` field was always required.
* If a method has exactly 15 arguments, the last argument is no longer required to be encoded in a tuple.
  * This makes the encoding more efficient for this number of arguments, rather than making the 15th argument appear in a tuple of size 1.
* Clarify that in the case of multiple logged values with the special 4-byte return prefix, the last one is the return value.

### Formatting changes
* Add comments to all JSON definitions and enable code highlighting
* Expand example JSON objects to contain valid fields
* Rename "foreign" types to reference types

### DIFF
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -128,26 +128,37 @@ information about methods in IDEs for HLLs.
 
 The JSON structure for such an object is:
 
-```
+```typescript
 interface Method {
+  /** The name of the method */
   name: string,
+  /** Optional, user-friendly description for the method */
   desc?: string,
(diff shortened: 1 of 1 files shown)


jannotti:
These look like good changes to me.

Issue #44, opened 2021-11-17
Discussion for ARC-0004: ABI


fabrice102:
We're moving the discussion for ARC-0004 from #7 to here.
Draft is https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md

fabrice102:
Moving the questions out of ARC-0004 there:

* Would people prefer “Method prototype” here in order to avoid using the word “signature” again?
* Should we allow the contract description to specify methods that require an opt in
    * As a subquestion: should we allow methods to require an opt-in in some cases, but not in other cases?
* Further questions: Should it be possible to describe requirements of certain fields of the transactions named this way? For example, can the amount of recipient of a pay transaction be specified? A specific, important instance of this problem: How can logicsigs on these transactions be provided in the JSON interface?

barnjamin:
>For example, can the amount of recipient of a pay transaction be specified?

Does it make sense to define a format to specify a 'transaction filter'? Strict comparison only to validate certain fields. 

It's come up elsewhere: 
https://github.com/algorandfoundation/ARCs/pull/41/files#r754713743

>How can logicsigs on these transactions be provided in the JSON interface?

How do you see logic sigs being used in the method?

wheresaddie:
> Would people prefer “Method prototype” here in order to avoid using the word “signature” again?

yes. is important to delineate from my pov.

AustP:
> Some AVM opcodes require specific values to be placed in the "foreign arrays" of the application call transaction. These three types allow methods to describe these requirements. To encode method calls that have these types as arguments, the value in question is placed in the Accounts (apat), Foreign Assets (apas), or Foreign Apps (apfa) arrays, respectively, and a uint8 containing the index of the value in the appropriate array is encoded in the normal location for this argument.

I think having a uint8 that contains the index of where the reference type is in the respective array is a little overkill. It would be much more straightforward if the order that the types appeared in the method signature was the order that the values appeared in the foreign arrays. For example, assume I have a method signature: `transfer(asset,account,account)void` which transfers an asset from the first account to the second account.

Currently, getting the asset ID requires three opcodes:

```
txna ApplicationArgs 1
btoi
txnas Assets
```

If we required the callers to put the values in the foreign arrays according to the order of the method signature, I could get the asset ID in a single opcode:

```
txna Assets 0
```

[...]

PR #55, opened 2021-12-09
ARC-0004: Misc changes

fabrice102:
Changes:
* minor style changes, in particular using **MUST**…
* allow descriptions `desc` in interfaces and contracts
* add comment that interfaces designer should be mindful to prevent collisions of method selectors between interfaces that are likely to be implemented together by the same application.
* clarify that method names do not need to be uniq
* restrict contract and interface names to Interface to  `[_A-Za-z][A-Za-z0-9_]*` , which would be useful if later we want to say that contract XXX implements interface YYY. Allowing full UTF-8 or spaces can quickly make things cumbersome.
* make interface names `ARC-XXXX...` reserved for ARCs
* allow method selector for `_optIn` and `_closeOut` to not be unique
* add a note that reference types can be part of a tuple is in the 15-th argument or after
* clarifying the order of transactions using two possibles transactions
* enforce some additional rules when
* removing the section “New Features in support of the ABI”
* removed the questions and moving them there: https://github.com/algorandfoundation/ARCs/issues/44

### DIFF
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -593,6 +593,7 @@ a client would create a transaction group containing, in this order:
 1. an asset transfer
 2. a payment
 3. the actual application call
+
 When encoding the other (non-transaction) arguments, the client 
 **MUST** act as if the transaction arguments were completely absent 
 from the method signature. The application call would contain the method 
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -603,9 +603,9 @@ ApplicationArgs[1], and the fourth (uint32) argument in ApplicationArgs[2].
 
 ARC-4 Apps **SHOULD** be constructed to allow their invocations to be
 combined with other contract invocations in a single atomic group if
-they can do so safely. For example, they **SHOULD** use `txns` to examine
+they can do so safely. For example, they **SHOULD** use `gtxns` to examine
 the previous index in the group for a required `pay` transaction,
-rather than hardcode an index with `txn`.
+rather than hardcode an index with `gtxn`.
 
 In general, an ARC-4 app method with n transactions as arguments **SHOULD** 
 only inspect the n previous transactions. In particular, it **SHOULD NOT**
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -43,7 +43,7 @@ as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
 
 > Comments like this are non-normative.
 
-All strings of type `string` are UTF-8 encoded.
+Interfaces are defined in TypeScript. All the objects that are defined are valid JSON objects, and all JSON `string` types are UTF-8 encoded.
 
 ### Overview
 
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -509,7 +509,7 @@ encoding these types are covered above.
 ### Reference Types
 
 Three special types are supported _only_ as the type of an
-argument. They **MUST** not be embedded in arrays or tuples, as 
+argument. They **MUST NOT* be embedded in arrays or tuples, as 
 their encoding is unusual. (They **MAY** implicitely be part of
 a tuple if they are the type of the 15-th argument or an argument
 further away.)
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -11,7 +11,8 @@ status: Draft
 ## Summary
 
 A standard for encoding contract call transactions to invoke methods
-on an Algorand _Smart Contract_ (aka "stateful contract" or "app").
+on an Algorand _Smart Contract_ (aka "stateful contract", "contract",
+"application", or "app").
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


ARCs/arc-0004.md
```diff
@@ -72,23 +74,21 @@ encoded byte arrays conveniently in the AVM.
...
-> Would people prefer "Method prototype" here in order to avoid using
-> the word "signature" again?
-
 A method signature is a unique identifier for a method. The signature
 is a string that consists of the method's name, an open parenthesis, a
 comma-separated list of the types of its arguments, a closing
 parenthesis, and the method's return type, or `void` if it does not
-return a value. The names of arguments are _not_ included in a
-method's signature, and whitespace does not appear anywhere.
+return a value. The names of arguments **MUST NOT** included in a
```
barnjamin, line 81: Must not be

ARCs/arc-0004.md
```diff
@@ -218,18 +239,19 @@ For example:
 
 ### Contracts
 
-A Contract is the complete set of methods that an app implements. It
+A contract is the complete set of methods that an app implements. It
```
barnjamin, line 242: Why lowercase? Capitalized elsewhere and I like it as a proper noun

ARCs/arc-0004.md
```diff
@@ -474,8 +509,10 @@ encoding these types are covered above.
 ### Reference Types
 
 Three special types are supported _only_ as the type of an
-argument. They must not be embedded in arrays or tuples, as their
-encoding is unusual.
+argument. They **MUST** not be embedded in arrays or tuples, as
```
barnjamin, line 512: MUST NOT

ARCs/arc-0004.md
fabrice102: Yes, there is a bit of inconsistency, because "interface" was always lower case. So not sure what convention to adopt.

ARCs/arc-0004.md
fabrice102: Don't hesitate to use the suggestion button (leftmost button of the Github editor comment), which allows easier merging.
See:
```suggestion
argument. They **MUST NOT** not be embedded in arrays or tuples, as 
```

ARCs/arc-0004.md
```diff
@@ -183,16 +196,24 @@ The JSON structure for such an object is:
...
+  desc?: string,
   /** All of the methods that the interface contains */
   methods: Method[]
 }
 ```
 
+Interface names **MUST** satisfy the regular expression `[_A-Za-z][A-Za-z0-9_]*`.
+Interface names starting by `ARC` are reserved to interfaces defined in ARC.
+Interfaces defined in `ARC-XXXX` (where `XXXX` is a 0-padded number) **SHOULD**
+start with `ARC-XXXX`.
```
jasonpaulos, line 209:
```suggestion
start with `ARC_XXXX`.
```
The hyphen character is not accepted by the above regex, so probably an underscore should be used instead.

ARCs/arc-0004.md
jasonpaulos: I'd support capitalizing Interface and Contract everywhere for consistency

ARCs/arc-0004.md
```diff
@@ -555,42 +588,40 @@ group, which is undesirable. Instead, the locations of the
...
-> certain fields of the transactions named this way? For example, can
-> the amount of recipient of a pay transaction be specified? A
-> specific, important instance of this problem: How can logicsigs on
-> these transactions be provided in the JSON interface?
+For example, to invoke the method `deposit(string,axfer,pay,uint32)void`,
+a client would create a transaction group containing, in this order:
+1. an asset transfer
+2. a payment
+3. the actual application call
+When encoding the other (non-transaction) arguments, the client
```
jasonpaulos, line 596:
```suggestion
3. the actual application call

When encoding the other (non-transaction) arguments, the client 
```

This needs another line break, otherwise the "When encoding..." sentence is part of the list

ARCs/arc-0004.md
```diff
@@ -555,42 +588,40 @@ group, which is undesirable. Instead, the locations of the
...
+selector in ApplicationArgs[0], the first (string) argument in 
+ApplicationArgs[1], and the fourth (uint32) argument in ApplicationArgs[2].
 
 
 ARC-4 Apps **SHOULD** be constructed to allow their invocations to be
 combined with other contract invocations in a single atomic group if
-they can do so safely. For example, they should use `txns` to examine
+they can do so safely. For example, they **SHOULD** use `txns` to examine
 the previous index in the group for a required `pay` transaction,
 rather than hardcode an index with `txn`.
```
jasonpaulos, line 607:
```suggestion
they can do so safely. For example, they **SHOULD** use `gtxns` to examine
the previous index in the group for a required `pay` transaction,
rather than hardcode an index with `gtxn`.
```

Tangential to this change, but this should be `gtxns` and `gtxn`

ARCs/arc-0004.md
```diff
@@ -555,42 +588,40 @@ group, which is undesirable. Instead, the locations of the
...
-
-Some changes are needed to the Algorand protocol and its tools to
-better support this ABI:
+In general, an ARC-4 app method with n transactions as arguments **SHOULD** 
+only inspect the n previous transactions. In particular, it **SHOULD NOT**
+inspect transactions after and it **SHOULD NOT** check the size of a transaction 
+group (if this can be done safely). 
+In addition, a given method **SHOULD** always expect the same
+number of transactions before itself. For example, the method 
+`deposit(string,axfer,pay,uint32)` is always preceded by two transactions.
```
jasonpaulos, line 615:
```suggestion
`deposit(string,axfer,pay,uint32)void` is always preceded by two transactions.
```
Return type missing

ARCs/arc-0004.md
```diff
@@ -42,6 +42,8 @@ as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
 
 > Comments like this are non-normative.
 
+All strings of type `string` are UTF-8 encoded.
```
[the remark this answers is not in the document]
fabrice102: I was thinking the `string` in the TS interfaces. But what about the strings in the ABI?

ARCs/arc-0004.md
fabrice102: I've modified this. Doing so, I've realized there may be some slightly confusing naming.
Let me know what you think. I see two types of objects:
* a "smart contract" or "application" or "app" or "stateful application" or "contract" is an actual application on Algorand
* a "Contract" is "the complete set of methods that an app implements"

Maybe to clarify should we restrict the vocabulary to:
* "application" for the one above
* "Contract" for the one below

so no "contract call transaction" but "application call transaction".

What do you think?

ARCs/arc-0004.md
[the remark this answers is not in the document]
jasonpaulos: Got it, maybe this would be more clear?

```suggestion
Interfaces are defined in TypeScript. All the objects that are defined are valid JSON objects, and all JSON `string` types are UTF-8 encoded.
```

And ABI strings are "assumed to contain UTF-8 encoded content."

jasonpaulos (approved):
Changes look good to me

fabrice102:
I'm taking care of the comment above 
in another commit:
https://github.com/algorandfoundation/ARCs/pull/55#discussion_r768239652

and merge this PR since it looks that people are happy with it.

#44

PR #57, opened 2021-12-11
ARC-0004: Change how contract creation and OnCompletion actions work

jasonpaulos:
In order to simplify and make more consistent how contract creation and OnCompletion actions work, this PR removes the special methods `_optIn` and `_closeOut`. In their place, I've explicitly defined "bare application calls" and provided guidance for how and when they should be supported.

The general idea is that if your app creation call or OnCompletion action requires no arguments and no return value, then your contract **SHOULD** implement these operations as bare application calls. Otherwise, if there are some situations in which you require arguments or a return value, your contract **SHOULD NOT** allow bare application calls for this case, and instead provide one or more ABI methods which can be called with the appropriate action.

Additionally, in order to make it easier to read the return value of a contract-to-contract ABI method call, this PR requires the return value logged item to be the _final_ logged item by an application execution.

### DIFF
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -60,14 +60,14 @@ an application call transaction. A method must have a name, it may
 take a list of arguments as input when it is invoked, and it may
 return a single value (which may be a tuple) when it finishes
 running. The possible types for arguments and return values are
-described later.
+described later in the [Encoding](#encoding) section.
 
 Invoking a method involves creating an application call transaction to
 specifically call that method. Methods are different from internal
-"helper functions" that may exist in a contract, but are not
-externally callable. In the future, applications will be able to invoke
-methods from other applications, so it should be possible to build ABI
-encoded byte arrays conveniently in the AVM.
+subroutines that may exist in a contract, but are not externally
+callable. Methods may be invoked by a top-level application call
+transaction from an off-chain caller, or by an application call inner
+transaction created by another smart contract.
 
 
 #### Method Signature
@@ -117,8 +117,8 @@ beyond its signature. This description is encoded in JSON and consists
 of a method's name, description, arguments (their types, names, and
 descriptions), and return type (and possible description for the
 return type). From this structure, the method's signature and selector
-can be calculated. The Algorand SDKs will provide convenience
-functions to calculate signatures and selectors from such JSON files.
+can be calculated. The Algorand SDKs provide convenience functions
+to calculate signatures and selectors from such JSON files.
 
 These details will enable high-level languages and dapps/wallets to
 properly encode arguments, call methods, and decode return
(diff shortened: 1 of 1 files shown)


ARCs/arc-0004.md
```diff
@@ -143,12 +143,30 @@ interface Method {
...
-For example:
+Where `OnCompletionAction` is defined as:
+
+```typescript
+type OnCompletionAction = "noop" | "optin" | "closeout" | "clearstate" | "update" | "delete";
+```
+
+Each string value represents an application call transaction OnCompletion
+(`apan`) value. The mapping from these string values to the numeric values
+for transaction encoding is:
```
barnjamin, line 161: Should we link directly to the go source file that defines these?

wheresaddie: yes. I would

jasonpaulos: Good suggestion, I'll add a link tomorrow

ARCs/arc-0004.md
```diff
@@ -221,23 +242,40 @@ For example:
...
+A Contract **MAY** implement a special method with the name
+`_create`. This method describes how a caller can instantiate a
+specific contract. This method **MAY** have arguments and a
+return value, and there **MAY** be multiple methods with this
+name as long as each method produces a unique selector. The
+`_create` method **MUST** only be invoked by an application
+creation transaction (i.e. an application call transaction with
+the special application ID 0). The Contract implementation
+**MUST** reject the transaction if a non-creation application
+call invokes the `_create` method.
```
jasonpaulos, line 267: We may also want to add the option to specify schema sizes and extra pages to a Contract's JSON to make the Contract easier to create.

ARCs/arc-0004.md
```diff
@@ -143,12 +143,31 @@ interface Method {
...
   /** Information about the method's return value */
   returns: { type: string, desc?: string }
 }
 ```
 
-For example:
+Where `OnCompletionAction` is defined as:
+
+```typescript
+type OnCompletionAction = "noop" | "optin" | "closeout" | "clearstate" | "update" | "delete";
```
fabrice102, line 156: Should we switch to the TEAL constants (https://developer.algorand.org/docs/get-details/dapps/avm/teal/specification/#oncomplete):
`NoOp`, `OptIn`, `CloseOut`, `ClearState`, `UpdateApplication`, `DeleteApplication`
for consistency?

ARCs/arc-0004.md
fabrice102: Is the `_create` method like the old `_optIn` and `_closeOut`, that is you provide the empty string as first application argument.
Or do you use the normal selector as first application argument?
I think the latter, but just wanted to confirm.

ARCs/arc-0004.md
```diff
@@ -221,23 +243,40 @@ For example:
...
+OnCompletion actions at runtime if they do not apply to the Contract's
+implementation). Additionally, the Contract's method definition **MAY**
+include additional allowed OnCompletion actions if its implementation
+supports them.
+
+Method names in Contracts **MAY** begin with underscore, but these
+names are reserved for use by this ARC and its future revisions.
+
+A Contract **MAY** implement a special method with the name
+`_create`. This method describes how a caller can instantiate a
```
fabrice102, line 260: Is the `OnCompletion` restricted to `NoOp`?

ARCs/arc-0004.md
```diff
@@ -221,23 +243,40 @@ For example:
...
+**MUST** reject the transaction if a non-creation application
+call invokes the `_create` method.
+
+Additionally, a Contract **MAY** allow application calls with
+no arguments. The primary purpose of this is to allow the execution
+of an OnCompletion action which requires no input data and has no
+return value. A Contract **MAY** allow this for all OnCompletion
+values, or only for a subset of them. Of course, great care should
+be taken when allowing these operations.
+
```
fabrice102, line 276: As in the previous version, should we **RECOMMEND** to have an empty optin/closeout?

ARCs/arc-0004.md
```diff
@@ -221,23 +243,40 @@ For example:
...
+return value, and there **MAY** be multiple methods with this
+name as long as each method produces a unique selector. The
+`_create` method **MUST** only be invoked by an application
+creation transaction (i.e. an application call transaction with
+the special application ID 0). The Contract implementation
+**MUST** reject the transaction if a non-creation application
+call invokes the `_create` method.
+
+Additionally, a Contract **MAY** allow application calls with
+no arguments. The primary purpose of this is to allow the execution
```
fabrice102, line 271: Should this be specified explicitly in the description?
Like with an "emptyAllowedActions" field?

ARCs/arc-0004.md
```diff
@@ -339,18 +387,28 @@ the latest one is the return value.
...
+approve or reject the transaction, and the following steps **MUST** be
+ignored.
+
+2. **MUST** Examine `txna ApplicationArgs 0` and `txn OnCompletion` to
+identify the selector of the method being invoked and the OnCompletion
+value of the call. If the contract does not implement a method with
+that selector, or the method with that selector does not supported
+the OnCompletion value, the call **MUST** fail.
+
+3. **SHOULD** branch to the body of the method indicated by the selector.
```
fabrice102, line 406:
```suggestion
3. **SHOULD** execute the actions that the method indicate by the selector prescribes. In general, this works by branching to the body of the method indicated by the selector.
```

fabrice102 (commented):
Another question removed is:
should the description of a contract indicates which method requires opt-in.
The issue is that it's possible that one method requires opt-in in some cases and not others.
Should we recommend that this does not happen?

In any case, if we specify this, I believe we need to provide the following warning:

WARNING: Callers **MUST NOT** rely on whether a given method is expected to require opt-in or not to know whether the caller account is opted in or not. In particular, for applications that restrict opt-in to a small subset of accounts, callers **MUST NOT* use the fact that a given application call succeed to conclude that the sender is in this small subset of accounts.

> The above warning is to prevent tempting optimizations (implicitly checking opt-in of an account via application call instead of explicitly checking it) that can lead to important security issues.

ARCs/arc-0004.md
```diff
@@ -143,12 +143,31 @@ interface Method {
     /** Optional, user-friendly description for the argument */
     desc?: string
   }>,
+  /** A list of allowed OnCompletion values for the method. Must not be empty */
+  allowedActions: OnCompletionAction[],
```
jannotti, line 147: If these are specified in an interface, must a contract support all of them in order to be considered an implementation of the interface?  Can it support more?  I think perhaps this belongs only in Contract methods, not Interface methods.  I dislike that nearly so many methods will be annotated that they allow optin.  I prefer that the ABI says "all methods should support optin".  I prefer to make rules that try to regularize ABI contracts, rather than expand they ABI to support describing arbitrary complexity.

ARCs/arc-0004.md
jannotti: It seems weird to have clearstate here.  Contracts can't say no.  They could return 0, which I suppose would cause the effects not to occur, but the clearstate would still happen.

ARCs/arc-0004.md
```diff
@@ -143,12 +143,31 @@ interface Method {
...
+Each string value represents an application call transaction OnCompletion
+(`apan`) value from https://github.com/algorand/go-algorand/blob/18cd6cda606c7d035bed4d2cfefaf4f25d009fa3/data/transactions/application.go#L51-L84.
+The mapping from these string values to the numeric values for transaction
+encoding is:
+* `noop`: 0
+* `optin`: 1
+* `closeout`: 2
+* `clearstate`: 3
+* `update`: 4
+* `delete`: 5
```
jannotti, line 168: I don't think this needs to be here.

ARCs/arc-0004.md
```diff
@@ -158,6 +177,7 @@ For example:
     { "type": "uint64", "name": "a", "desc": "The first term to add" },
     { "type": "uint64", "name": "b", "desc": "The second term to add" }
   ],
+  "allowedActions": [ "noop" ],
```
jannotti, line 180: I dislike that this will appear everywhere.

ARCs/arc-0004.md
```diff
@@ -221,23 +243,40 @@ For example:
...
-OnCompletion values to be set to other values in order to support
-methods that delete or update the contract. Of course, great care
-should be taken when allowing these operations.
+(specifically, they must have unique method selectors).
+
+Methods in a Contract that originate from an implemented Interface **MUST**
+contain every allowed OnCompletion action from the Interface's definition
+of the method (though the Contract's program **MAY** reject some of these
+OnCompletion actions at runtime if they do not apply to the Contract's
+implementation). Additionally, the Contract's method definition **MAY**
```
jannotti, line 252: I don't understand what this means at all.  They must specify that they allow the operation, but they might not actually allow the operation?

ARCs/arc-0004.md
jannotti: How can a contract support creation?  You would have to supply the bytecode itself.  So the contract would have to specify the bytecode. How else could you call it?

jannotti (commented):
I think we should be focused on making optin and optout easy and clear, and not worry about create/delete/clearstate or consistency with them.  They are very weird operations for normal contract consumers to care about.

ARCs/arc-0004.md
jasonpaulos: Not just noop. Any `OnCompletion` value except close out or clear state can be specified for this, as those are the only ones that the network will reject.

Otherwise it's perfectly valid to opt-in during creation, or even set `OnCompletion` to delete to immediately self destruct the app. It's also valid to set `OnCompletion` to update, though this is functionally the same as noop.

ARCs/arc-0004.md
jasonpaulos: The `_create` method would behave like other methods, meaning the selector should always be the first arguments.

jasonpaulos:
Based on more discussion, we've decided to make the ABI less opinionated about how specific OnCompletion actions and create methods are implemented. I've updated the PR description and changes accordingly.

fabrice102:
Merged after some conflict fixes. Hopefully there was no mix-up.

PR #58, opened 2021-12-13
Contract templates

barnjamin:
*ROUGH* draft

Submitting to get feedback as I work it over

### DIFF
--- a/ARCs/arc-000N.md
+++ b/ARCs/arc-000N.md
@@ -6,7 +6,7 @@ author: Ben Guidarelli<@barnjamin>
 discussions-to: TODO 
 status: Draft
 type: Standards Track
-category:  ARC>
+category:  ARC
 created: December 13, 2021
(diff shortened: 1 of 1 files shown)


barnjamin:
This ARC would require these changes be made available:

https://github.com/algorand/go-algorand/pull/2921

https://github.com/algorand/go-algorand/pull/2885

cusma:
@barnjamin I like the idea and the approach!

SudoWeezy:
We might want to pursue this, currently, Wallets should not run *LSIG* blindly (Due to the risk).

Projects who want to use LSIG somewhere, (For example, RandGallery with Offers) need to submit a PR with their template code to the repository associated with this ARC https://github.com/algorandfoundation/ARCs/tree/main/assets/ARC-X.

It might be a way for wallets to have an "Audited Repository."

WDYT?

PR #59, opened 2021-12-18
ARC-0004 Minor clarifications and last call

fabrice102:
Change ABI to last call.

Handling
https://github.com/algorandfoundation/ARCs/pull/55#discussion_r768239652:
Reduced terminology to "Application", "Contract", "Interface". No more
"smart contracts" to indicate Algorand Applications.

Change "future revisions" to "future ABI-related ARC" as if we follow
Ethereum EIP, only ARC-0 should be living and have revisions.

One question after this pass:
Why prefixing the return log value?
Is the caller supposed to check the prefix?

### DIFF
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -1,7 +1,7 @@
 ---
 arc: 4
 title: Algorand Application Binary Interface (ABI)
-status: Draft
+status: Last Call
 ---
 
 # Algorand Transaction Calling Conventions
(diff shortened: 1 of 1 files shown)


jasonpaulos:
These changes look good to me, except it does not actually change the status to last call?

> Why prefixing the return log value?

We reserve the right to add support for other logged "events" with different 4 byte prefixes in a future revision of this ARC. So if return values did not have a prefix, it would be possible to incorrectly interpret a return log as a different event if its first 4 encoded bytes happen to match that event's prefix.

> Is the caller supposed to check the prefix?

Yes, the caller should verify the final logged item has the 4 byte return prefix. This acts as an extra bit of insurance to make sure that the value the client is decoding is meant to be the return value.

fabrice102:
Merging as consensus seems achieved.

PR #60, opened 2021-12-23
Fix incorrect capitalization for arc filename refs

GibsDev:
Some references to other arc files returned a 404 on GitHub due to incorrect capitalization of arc filenames.


barnjamin (approved):
Thanks!

PR #61, opened 2022-01-04
ARC for standardizing traits (attributes) in token metadata

keeganthomp:
Establishes a standard for how to declare an NFT's traits inside a token's metadata.

### DIFF
--- a/ARCs/arc-0016.md
+++ b/ARCs/arc-0016.md
@@ -16,7 +16,7 @@ This document introduces the standard for declaring traits (attributes) inside a
 
 ## Abstract
 
-The goal is to establish a standard for traits are declared inside a non-fungible ASA's (NFT) metadata.
+The goal is to establish a standard for how traits are declared inside a non-fungible ASA's (NFT) metadata.
 
 ## Specification
 
--- a/ARCs/arc-0016.md
+++ b/ARCs/arc-0016.md
@@ -84,7 +84,7 @@ The JSON schema for `traits` is as follows:
 ## Rationale
 
 This propoal is to standard how traits are structured for NFT's on Algorand.
-A standard is needed for traits so programs know what to expect in order to calculate things rarity.
+A standard is needed for traits so programs know what to expect in order to calculate things like rarity.
 
 ## Copyright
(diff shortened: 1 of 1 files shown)


whereisrysmind:
+1

koen-vr:
ATM this would be valid, and further nesting as well? If that is the goal, it should be more explicit?

```json
{
  ...
  "properties": {
    "creator": "Tim Smith",
    "created_at": "January 2, 2022",
    "traits": {
      "background": "red",
      "shirt_color": "blue",
      "glasses": {
        "style": 5,
        "color": "green"
      },
      "tattoos": [
        4, 2, 8
      ],
    }
  }
}
```

keeganthomp:
No, that would not be valid according to this spec.

"Values must be strings or numbers".

My goal is to propose a simple solution that is easily digestible by programs. Once you allow nesting, it's a slippery slope that will inevitably lead a pile of spaghetti.

The people seemed to really like the unofficial ARC69. This is narrowed heavily from that and inspired from ARC3.

koen-vr:
> No, that would not be valid according to this spec.
>
> "Values must be strings or numbers".

Ok, I was not clear about your intent, Then on line 49 I would change 'may' to "must" to be in line with [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt). Then repeat the rule outside of the description or add validation rules to the schema so the following the schema enforces the rule.

```json
{
  "title": "Traits for Non-Fungible Token",
  "type": "object",
  "properties": {
    "traits": {
      "type": "object",
      "description": "Traits (attributes) that can be used to calculate things like rarity. Values must be strings or numbers",
      "minProperties": 1,
      "additionalProperties": {
        "type": ["number", "string"],
      }
    }
  }
}
```

fabrice102:
There seems to be some consensus there.
So merging it in the main repo and creating an issue for discussion:
Discussion is now on https://github.com/algorandfoundation/ARCs/issues/62

jannotti:
I'm concerned about using "none" as a special value.  Why not use `null`, an existing JSON value?

PR #63, opened 2022-01-24
ARC-0019: NFT ASA URL templating proposal for mutable IPFS CID replacement in URLs.

pbennett:
Initial draft

### DIFF
--- a/ARCs/arc-draft_asa_url_templating.md
+++ b/ARCs/arc-draft_asa_url_templating.md
@@ -1,0 +1,181 @@
+---
+arc: <to be assigned>
+title: Templating of NFT ASA URLs for mutability
+description: A proposal to allow a templating mechanism of the URL so that changeable data in an asset can be substituted by a client, proving a mutable URL.
+author: Patrick Bennett / TxnLab Inc. (@pbennett)
+discussions-to: #arcs in Algorand Discord
+status: Draft
+type: Standards Track
+category ARC
+created: 2021-01-23
+requires 
+---
+
+## Abstract
+This ARC describes a template substitution for URLS in ASAs, initially for ipfs:// scheme URLS allowing mutable CID replacement in rendered URLs.
+
+The proposed template-XXX scheme has substitions like: 
+```
+template-ipfs://{ipfscid:(version):(multicodec):(field name containing 32-byte digest, ie reserve):(hash type)}[/...]
+```
+
+This will allow modifying the 32-byte 'Reserve address' in an ASA to represent a new IPFS content-id hash.  Changing of the reserve address via an asset-config transaction will be all that is needed to point an ASA URL to new IPFS content.  The client reading this URL, will compose a fully formed IPFS Content-ID based on the version, multicodec, and hash arguments provided in the ipfscid substitition.
+
+## Motivation
+
+While it is appropriate for many NFTs to be completely immutable, like ARC3 NFTs, there are cases where some type of mutability is desired for NFT metadata and/or images.  The data referenced should itself be immutable, but the ability to change what data is pointed to should be mutable. The data should also be able to be of significant size if necessary.
+
+Algorand ASAs support mutation on-chain of a few parameters - the Manager, Clawback, Freeze, and Reserve addresses [unless previously cleared].  These are changed via an asset-config transaction.  An asset-config transaction may include a note, but it is limited to 1KB and accessing this value requires clients to use an indexer to iterate/retrieve the values.
+
+Of the changeable properties, the Reserve address is somewhat distinct in that it is not used for anything directly on-chain.  It is used solely for determining what is in/out of circulation (by subtracting supply from that held by the reserve address).  With an NFT, the Reserve address is irrelevant as it is a 1 of 1 unit.  Thus, the Reserve address is usable as a 32-byte 'bitbucket'.
+
+These 32-bytes can thus be repurposed as efficient storage for a SHA2-256 hash uniquely referencing the desired content for the ASA (ARC3-like metadata for example)
+
+Using the reserve address in this way means that what an ASA 'points to' for metadata can be changed with a single asset config transaction, changing only the 32-bytes of the reserve address, and being able to reference any size of content in IPFS.  The new data is also accessible via even non-archival nodes with a single call to /v2/assets/xxx API to fetch the ASA data that, for something wanting to fetch/display NFTS much be fetched anyway.  No expensive or slow indexer calls are required.  Asset fetches are always near instant as all nodes maintain current state of the asset definition. 
+
+## Specification
+
+An indication that this ARC is in use is defined by an ASA URL's "scheme" having the prefix "template-".   This proposal is putting forth a simple solution for a specific problem, namely mutability for IPFS hosted content-ids.  The intention is that FUTURE ARCs could define additional template substitutions, but this is not meant to be a kitchen sink of templates, only to establish a possible baseline of syntax.
+
+An Asset meeting this specification **MUST** have:
+
+1. **URL Scheme of "template-ipfs"**
+
+```template-ipfs://(...)```
+
+The ifps:// scheme is already somewhat of a meta scheme in that clients interpret the ipfs scheme as referencing an IPFS CID (version 0/base58 or 1/base32 currently) followed by optional path within certain types of IPFS DAG content (IPLD CAR content for example).  The clients take the CID and use to fetch directly from the IPFS network directly via IFPS nodes, or via various IPFS gateways (https://ipfs.io/ipfs/CID[/...], pinata, etc.)).  
+
+2. **An "ipfscid" _template_ argument in place of the normal CID.**
+
+Where the format of templates are {*template type* [':' delimited parameters...])
+
+The ipfscid template definitions is based on properties within the IPFS CID spec: https://github.com/multiformats/cid  
+```	ipfscid:{version}:{multicodec content-type name}:[field name containing 32-byte digest, ie reserve]:{hash type}```
+
+The intent is to recompose a complete CID based on the content-hash contained within the 32-byte reserve address, but using the correct multicodec content type, ipfs content-id version, and hash type to match how the asset creator will seed the IPFS content.  If a single file is added using the 'ipfs' CLI via 'ipfs add --cid-version=1 metadata.json' then the resulting content will be encoded using the 'raw' multicodec type.  If a directory is added containing one or more files, then it will be encoded using the dag-pb multicodec.  CAR content will also be dag-pb.  Thus based on the method used to post content to IPFS, the ipfscid template should match. 
+
+The parameters to the template ipfscid are:
+2. IPFS _version_, **MUST** be '0' or '1' representing the IPFS CID version.
+3. _Multicodec content-type name_, **MUST** be an IPFS multicodec name of 'raw' or 'dag-pb'.  Other codecs SHOULD be supported but are beyond the scope of this proposal.
+4. _Field name_ **MUST** be 'reserve' to represent the reserve address is used for the 32-byte hash but is specified here so future iterations of the specification may allow other fields or syntaxes to reference other mutable field types.
+5. _Hash type_ **MUST** be 'sha2-256' but is explicitly specified so as to allow future versions the option of using other hashes. 
+
+IPFS may add future versions of the cid spec, and add additional multicodec types or other supported hashes.  Implementations SHOULD use IFPS libraries where possible that accept multicodec and hash types as named values and allow a CID to be composed generically.
+
+### Examples
+
+```
+ASA URL: template-ipfs://{ipfscid:0:dag-pb:reserve:sha2-256}/arc3.json
+ASA URL: template-ipfs://{ipfscid:1:raw:reserve:sha2-256}
+ASA URL: template-ipfs://{ipfscid:1:dag-pb:reserve:sha2-256}/metadata.json
+```
+
+## Rationale
+The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages.
+
+## Backwards Compatibility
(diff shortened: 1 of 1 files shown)


ARCs/arc-draft_asa_url_templating.md
```diff
@@ -0,0 +1,196 @@
...
+
+Of the changeable properties, the Reserve address is somewhat distinct in that it is not used for anything directly on-chain.  It is used solely for determining what is in/out of circulation (by subtracting supply from that held by the reserve address).  With an NFT, the Reserve address is irrelevant as it is a 1 of 1 unit.  Thus, the Reserve address is usable as a 32-byte 'bitbucket'.
+
+These 32-bytes can thus be repurposed as efficient storage for a SHA2-256 hash uniquely referencing the desired content for the ASA (ARC3-like metadata for example)
+
+Using the reserve address in this way means that what an ASA 'points to' for metadata can be changed with a single asset config transaction, changing only the 32-bytes of the reserve address, and being able to reference any size of content in IPFS.  The new data is also accessible via even non-archival nodes with a single call to /v2/assets/xxx API to fetch the ASA data that, for something wanting to fetch/display NFTS much be fetched anyway.  No expensive or slow indexer calls are required.  Asset fetches are always near instant as all nodes maintain current state of the asset definition.
```
barnjamin, line 33:
```suggestion
Using the reserve address in this way means that what an ASA 'points to' for metadata can be changed with a single asset config transaction, changing only the 32-bytes of the reserve address.  The new value is accessible via even non-archival nodes with a single call to the `/v2/assets/xxx` REST endpoint.  
```

ARCs/arc-draft_asa_url_templating.md
```diff
@@ -0,0 +1,196 @@
...
+
+Using the reserve address in this way means that what an ASA 'points to' for metadata can be changed with a single asset config transaction, changing only the 32-bytes of the reserve address, and being able to reference any size of content in IPFS.  The new data is also accessible via even non-archival nodes with a single call to /v2/assets/xxx API to fetch the ASA data that, for something wanting to fetch/display NFTS much be fetched anyway.  No expensive or slow indexer calls are required.  Asset fetches are always near instant as all nodes maintain current state of the asset definition. 
+
+## Specification
+
+An indication that this ARC is in use is defined by an ASA URL's "scheme" having the prefix "template-".   This proposal is putting forth a simple solution for a specific problem, namely mutability for IPFS hosted content-ids.  The intention is that FUTURE ARCs could define additional template substitutions, but this is not meant to be a kitchen sink of templates, only to establish a possible baseline of syntax.
```
barnjamin, line 37:
```suggestion
This proposal specifies a method to provide mutability for IPFS hosted content-ids.  The intention is that FUTURE ARCs could define additional template substitutions, but this is not meant to be a kitchen sink of templates, only to establish a possible baseline of syntax. 

An indication that this ARC is in use is defined by an ASA URL's "scheme" having the prefix "template-".  
```

ARCs/arc-draft_asa_url_templating.md
```diff
@@ -0,0 +1,196 @@
...
+
+An Asset meeting this specification **MUST** have:
+
+1. **URL Scheme of "template-ipfs"**
+
+```template-ipfs://(...)```
+
+The ifps:// scheme is already somewhat of a meta scheme in that clients interpret the ipfs scheme as referencing an IPFS CID (version 0/base58 or 1/base32 currently) followed by optional path within certain types of IPFS DAG content (IPLD CAR content for example).  The clients take the CID and use to fetch directly from the IPFS network directly via IFPS nodes, or via various IPFS gateways (https://ipfs.io/ipfs/CID[/...], pinata, etc.)).
```
barnjamin, line 45: make this non-normative with `>` and link to where its specified

ARCs/arc-draft_asa_url_templating.md
```diff
@@ -0,0 +1,196 @@
...
+
+2. **An "ipfscid" _template_ argument in place of the normal CID.**
+
+Where the format of templates are {*template type* [':' delimited parameters...])
+
+The ipfscid template definitions is based on properties within the IPFS CID spec: https://github.com/multiformats/cid  
+```	ipfscid:{version}:{multicodec content-type name}:[field name containing 32-byte digest, ie reserve]:{hash type}```
+
+The intent is to recompose a complete CID based on the content-hash contained within the 32-byte reserve address, but using the correct multicodec content type, ipfs content-id version, and hash type to match how the asset creator will seed the IPFS content.  If a single file is added using the 'ipfs' CLI via 'ipfs add --cid-version=1 metadata.json' then the resulting content will be encoded using the 'raw' multicodec type.  If a directory is added containing one or more files, then it will be encoded using the dag-pb multicodec.  CAR content will also be dag-pb.  Thus based on the method used to post content to IPFS, the ipfscid template should match.
```
barnjamin, line 54: more non-normative

ARCs/arc-draft_asa_url_templating.md
```diff
@@ -0,0 +1,196 @@
...
+		cidResult = cid.NewCidV1(uint64(codec), hash)
+	}
+	return fmt.Sprintf("ipfs://%s", strings.ReplaceAll(asaUrl, matches[0], cidResult.String())), nil
+}
+```
+
+#### Typescript Implementation
+
+A modified version of a simple ARC3 viewer can be found at https://github.com/TxnLab/arc3.xyz specifically the code segment at https://github.com/TxnLab/arc3.xyz/blob/main/src/lib/nft.ts#L41
+This is a fork of https://github.com/barnjamin/arc3.xyz
```
barnjamin, line 188: nnooiccee

barnjamin (commented):
Looks good for a first pass. 

Most of the suggestions were just around wording and trying to make it a little more brief

TommyAlex79 (commented):
Excellent proposal👍. Implementing this, will tweak if amended before merging. Thanks.

ARCs/arc-draft_asa_url_templating.md
```diff
@@ -0,0 +1,197 @@
...
+
+As can be seen here: https://testnet.algoexplorer.io/asset/66753108
+
+Using the forked https://github.com/TxnLab/arc3.xyz repo, with testnet selected, and the /nft/66753108 url - the browser will display the original content as-is, using only the Reserve address as the source of the content hash.
+
+## Rationale
+The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages.
+
+## Backwards Compatibility
+The 'template-' prefix of the scheme is intended to break clients reading these ASA URLS outright.  Clients interpreting these URLS as-is would likely yield unusual errors.  Code checking for an explicit 'ipfs' scheme for example will not see this as compatible with any of the default processing and SHOULD treat the URL as if it were simply unknown/empty.
```
fionnachan, line 93: URLS -> URLs

ARCs/arc-draft_asa_url_templating.md
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multihash"
+)
+// ...
+func ReserveAddressFromCID(cidToEncode cid.Cid) (string, error) {
+    decodedMultiHash, err := multihash.Decode(cidToEncode.Hash())
+    if err != nil {
+        return "", fmt.Errorf("failed to decode ipfs cid: %w", err))
+    }
+    return types.EncodeAddress(decodedMultiHash.Digest)
```
fionnachan, line 112: I'm slightly concerned that the 4-byte checksum added would cause confusion for decoding but I guess it's okay if it's used in an address field like `reserve`. I wonder how the encoding would be handled for the note field though.

fionnachan (commented):
Looks good in general

You probably have a specific use case for this proposed ARC but I think ipns also has its place for mutable NFTs. Curious why ipns is entirely left out of the conversation for NFT mutability. Haven't seen it in any ARCs so far or maybe I've missed it....

ARCs/arc-draft_asa_url_templating.md
pbennett: > You probably have a specific use case for this proposed ARC but I think ipns also has its place for mutable NFTs. Curious why ipns is entirely left out of the conversation for NFT mutability. Haven't seen it in any ARCs so far or maybe I've missed it....

Correct.  I looked at IPNS originally and intended to use it, but talking w/ IPFS devs, it's not really meant for large-scale use and has many problems, particularly with scale.  The key is effectively fixed to a particular clusters key and if referencing, for example, millions of 'files' off that single mutable ipns root then any change to any file requires a republishing of the entire root dag node and republishing the 'single' ipns resolution link.  
So, providing distinct, independent mutable content isn't appropriate for single ipns resolution, and on top of that, resolving ipns is extremely slow (as they admit).

pbennett: > I'm slightly concerned that the 4-byte checksum added would cause confusion for decoding but I guess it's okay if it's used in an address field like `reserve`. I wonder how the encoding would be handled for the note field though.

Oh, are you referring to the Algorand's 'display' of an Algorand Address based off the raw 32-byte public key where it adds a checksum and uses that for the 58 byte base32 'public account address?'  
Correct, it's completely ignored (and is within algorand as well - other than as input 'validation').  The SDK will handle going back and forth and as long as developers use the base 32-bytes of the PK it should work fine.

pbennett: ..continuing w/ IPNS point (going off the confused emoji response).
If you had an ASA URL of ipns://xxxxxxxx/somename.json yet you had millions of 'somename' (I will in my case at least) then you basically have a single IPFS DAG at root (imagine a single directory with a million files) with nodes for every file off it.  To add/change any file within that virtual directory - the hash of the node changing changes, as well as the creation of new root node (referencing ALL the children) and its new hash.  That single new 'root' hash has to then be published as "the" new IPNS hash to represent that ipns:// >- xxxx <-- link.  
It's 'ok' if you're doing for single root level changes but since the base ipns key that you're providing a replacement for is explicitly part of the configuration of the ipfs node itself, you really can't manage millions of those base entries like you'd want per ASA.

fionnachan: I am annoyed to hear that ipns doesn't scale and is slow. 😤 Your proposal looks like a decent solution to this problem and I might use it at some point as ipns wouldn't suit my use case with what you have described. For now, without testing the implementation, I don't have any further comments.

fergalwalsh:
This looks like a smart solution to the problem of wanting mutable metadata. 
But it seems like a problem that shouldn't exist. Surely this is further evidence of the need for well defined mutable (small) data field on Assets OR simply allowing URLs to be mutable going forward. I'd be much happier to see this solved at the protocol level rather than with a hack at the standards level.

pbennett:
> This looks like a smart solution to the problem of wanting mutable metadata. But it seems like a problem that shouldn't exist. Surely this is further evidence of the need for well defined mutable (small) data field on Assets OR simply allowing URLs to be mutable going forward. I'd be much happier to see this solved at the protocol level rather than with a hack at the standards level.

I'd love it at the protocol level as well, and will switch my solution's use of it if/when that becomes present but Algorand seems to want to take a hard line stance on mutability & protocol changes.  Allowing the URL to be mutable would be wonderful honestly (assuming the manager address was set and allowed it of course) but over time I suspect the restrictions of ASAs may just push more people to contract solutions.  

Assuming discovery is there, the ABI should allow interface-based assets, so we should be able to have ERC721/1155/etc equivalent 'smart assets' on Algorand.  NFTs wanting more flexible mutability, special restrictions, dynamic minting, royalties, etc. could go that route.  We then have another catch-up situation though where no wallets, marketplaces or explorers will support these changes initially.

I understand and respect the concerns over chain data costs and wanting to ensure that is 'paid' for (ie: the minimum balance reqs are almost like a storage lease), but given that txns can be issued at will w/ up-to 1K notes, allowing an ASAs (much smaller) URL to be changed at will seems an odd place to draw a line in the sand.  I think there's still a huge place for ASA NFTs in Algorand, and definitely support making the URL mutable.

fergalwalsh:
Fully agree 👍 

> but over time I suspect the restrictions of ASAs may just push more people to contract solutions.
Yes, and having native assets is one of the better features of Algorand. Moving towards everything being a contract would not be good for anyone. There is a huge advantage right now for apps and wallets to work with native assets, it would be a huge shame to loose that. The awful hacks introduced by arc69 have made a nightmare situation for wallets and explorers. Allowing for mutability would have avoided all that.

> I understand and respect the concerns over chain data costs and wanting to ensure that is 'paid' for (ie: the minimum balance reqs are almost like a storage lease), but given that txns can be issued at will w/ up-to 1K notes, allowing an ASAs (much smaller) URL to be changed at will seems an odd place to draw a line in the sand. I think there's still a huge place for ASA NFTs in Algorand, and definitely support making the URL mutable.

Definitely seems odd to put such limits on ASAs and to try to stick to them at ALL costs. I understand the difference between a note and an ASA field because notes don't get stored in the ledger but only the history whereas asset data is stored in the ledger that all nodes must hold. 
BUT, making urls mutable does not cost any storage at all. Adding an extra mutable field of the same size as the url would also have a very minimal cost. But mutable URLs seems like a win for everyone.

So the question is how do we petition Algorand Inc to consider these kind of changes?

pbennett:
> So the question is how do we petition Algorand Inc to consider these kind of changes?

We nag them, but not in this spec, that I need, today...  😜

ARCs/arc-draft_asa_url_templating.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Reference Implementation
+
+### Encoding
+#### Go[lang] implementation
+
+```go
+import (
+    "github.com/algorand/go-algorand-sdk/types"
+	"github.com/ipfs/go-cid"
```
fabrice102, line 103: minor: indentation issue

fabrice102 (changes requested):

fabrice102:
I like the concept. I made a few minor comments before we can merge as a draft, switch to last call and then make it final.
If some mutable fields are added, it could easily be extended, which is also very nice.

The main comment is that it will no more be a valid ARC-3 asset, because the metadata hash will not be valid after mutation.
Since the CID contains the hash, there is no security issue. So the main question I would say is how to document this properly and what convention to choose for the metadata hash in that case:
* Should it be the 0 string to make it clear it's mutable and we are not using ARC-3 metadata hash?
* Should it be the metadata hash of the initial JSON metadata?

Issue #64, opened 2022-02-02
Discussion for ARC-69: Community Algorand Standard Asset Parameters Conventions for Digital Media Tokens


fabrice102:
Following merging of ARC-69 as a living standard, this issue replaces #38 to discuss ARC-69.

PR #38, opened 2021-10-28
New ARC: ARC-0069: Community Algorand Standard Asset Parameters Conventions for Digital Media Tokens

ChrisAntaki:
This PR adds the popular [ARC69](https://github.com/algokittens/arc69) metadata spec to the `algorandfoundation/ARCs` GitHub repository

This ARC helps NFTs load as quickly and reliably as possible, with attributes, [backwards compatibility for existing NFTs](https://github.com/algokittens/arc69/blob/master/pre-existing-NFTs.md), and updateability (upgrading stats in an RPG, fixing typos, etc.)

The Algo NFT community has enthusiastically adopted this ARC, with more series adding metadata every day! Here are just a few of the awesome series using this ARC:
- Yieldlings
- M.N.G.O
- Eclectic Flamingos
- Banana Mints
- Stupid Horses

image

This ARC works great in production, on popular platforms like [NFT Explorer](https://www.nftexplorer.app/asset/359181888), [ALGOxNFT](https://algoxnft.com/listings/live), and [Rand Gallery](https://www.randgallery.com/algo-collection/?address=359181888)

image
NFT Explorer

Screenshot_20211030-025910
ALGOxNFT

image
Rand Gallery

### Artists
- [Add metadata to your existing NFTs](https://github.com/algokittens/arc69/blob/master/pre-existing-NFTs.md)
### Developers
- [Fetch and parse metadata for NFTs with JavaScript](https://github.com/RandGallery/arc69js)

Thanks for reading this!

PR #65, opened 2022-02-02
Extension of ARC-0069: Adding mime type for media and integrity fields

fabrice102:
Following comments in https://github.com/algorandfoundation/ARCs/pull/38

Field names are not consistent with ARC-3 that uses `mimetype`
in particular instead of `mime_type`.
But since ARC-69 already used the latter, I thought we should
try to be consistent within ARC-69.

Please comment in particular on naming.

See #64 too


robbiemu:
I'm very much a spectator as of yet: I am writing my first nft-related app currently. From what I gather, I think these points are relevant.

ARCs are meant to act as specifications. Historically this may not have been the case but in general this should mean that accepted ARCs are immutable. There is a PR here for ARC0000 that is meant to address intended usage (because arc0069 was out of sequence, and this led to some controversy) that highlights _informational ARCs_ as a type of ARC meant to provide guidance on standard practice. I am wondering if your problem couldn't be reduced to a compatibility layer under properties and converted to an informational ARC.

SudoWeezy:
@ChrisAntaki, I would like to have your feedback on this

ChrisAntaki:
I don't see how this PR solves real world problems

ARC3 has 5.5% marketshare while ARC69 has 75.6% marketshare. This PR's description starts with "Field names are not consistent with ARC-3" though. Clearly, the market thinks that's fine

ChrisAntaki:
I'm seeing almost all NFTs using IPFS URLs, which are based on deterministic hashes of file contents, so the `media_integrity` would be redundant there. Also, I'm not seeing NFTs or platforms really using the `media_url` param, so giving it a mimetype doesn't seem impactful

evanmaltz:
@ChrisAntaki I tend to agree. Both fields are redundant and so probably shouldn't be required, though I think adding the mime type is somewhat convenient for caching.

SudoWeezy:
I agree with both of you, @ChrisAntaki & @evanmaltz. 

- **Most** NFTs use IPFS, so it might be redundant to add integrity in most cases. 
- media_mime_type is always suitable for handling media files so we should keep it.

But if for some reason, the link is just a link to some random hosting without a hash, the IPFS is down, or it has been referenced with a folder:
> e.g. https://ipfs.io/ipfs/bafybeihk7p7ppu3gg5gtk3pqoxob7ccwoqysur67nuks4bv4qsweayt6nq/Gemini1.png

With *integrity fields*, the owner (If he downloaded the corresponding file) can easily prove that he owns it. 

As those fields are **OPTIONAL** and not **REQUIRED**, we should let them go into the ARC.
This way, we can put in **Final** status.

ChrisAntaki:
Just to clarify, I see no reason to add these fields. I'm not aware of any real world demand for them. The additional fields would increase the complexity of the spec for integrators, for pretty much no benefit. Also, I think the "Living" status is great to keep, much like the successful HTML spec (https://html.spec.whatwg.org/multipage/), unrelated to this PR

ChrisAntaki:
I'd be interested to hear feedback from @algokittens and @TommyAlex79 , who co-designed ARC69 with me. Do you two see real world demand for these fields?

PR #66, opened 2022-02-06
ARC-0004: fix typos

fionnachan:
Saw some typos while reading it.

For the commas -> semicolons, my prettier plugin was complaining about it so I also committed the change.


fabrice102:
Thanks!
For commas -> semicolons, it looks like TypeScript is ok with both. But since ARC-1 is written with semi-colon and https://github.com/algorand/js-algorand-sdk is also with semi-colons, let's stick with semi-colons.

PR #67, opened 2022-02-20
arc-0017 - ASA2ASA Stateless orderbook smart contract 

scholtz:
ASA2ASA Stateless orderbook smart contract for asa2asa p2p DEX usecases

### DIFF
--- a/ARCs/arc-0017.md
+++ b/ARCs/arc-0017.md
@@ -1,0 +1,283 @@
+---
+aip: 17
+title: A0B001 - ASA TO ASA STATELESS NO ROYALTY P2P DEX ORDERBOOK SMART CONTRACT
+status: Draft
+---
+
+# A0B001 - ASA2ASA Stateless orderbook smart contract
+
+## Summary
+
+This document introduce standard for P2P onchain orderbook smart contract.
+
+## Abstract
+
+The algorand blockchain is efficient decentralized smart contract capable blockchain. This standard introduces P2P trading opportunities with onchain orderbook capabilities. 
+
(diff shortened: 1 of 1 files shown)


SudoWeezy:
@scholtz If this PR is still relevant, please reach out to me.

scholtz:
> @scholtz If this PR is still relevant, please reach out to me.

Yes, see the progress in the discussion thread.. There are already applications using this.

https://github.com/algorandfoundation/ARCs/issues/68

SudoWeezy:
Thanks, in that case, could you please format it based on the ARC-0 recommendation?

scholtz:
what do you have in mind? can you do the formatting?

SudoWeezy:
I can do the formatting. 

But before that, the ARC should have a redesign, I would like to have the interface instead of plain `code`. (ARC are not meant to share code).
We can let the code as an example (just need to add a disclaimer that it has not been audited).

Is there a way to use the note field to find the logicsig?  It might be easier for everyone to read Price & Multiplier than using the prefix.

scholtz:
algorand indexer does not search by note field. It does not search by hash of the logicsig, but first bytes in the logicsig might be the same always.. cusotm indexer has to be writtend down to find all these type of transactions which i already did and it is open source

i believe the standardized arc should have code inside.. there is no other way to verify integrity of the contract than just to take whole code and check it with initial variables. I was asking Jason why the native code does not use the variables from the input in the later code and the response was they do not plan to implement it ..

if the byte code would be the same from some byte, with hash of logicsig from that byte it would be possible to create simple integrity check which is not possible right now. Therefore everyone who wants to interact with this smart contract must have the same version of it, therefore the code must be standardized.

Issue #27, opened 2021-09-21
Discussion: NFT Royalties 


barnjamin:
This issue is to discuss requirements and possible specification for an enforceable Royalty system for Algorand NFTs.

Please comment with any relevant information or ideas. 

If you're already doing this please describe the method you're using.

whereisrysmind:
There are several considerations here so far that I see. 1) We need to know the royalty percent, I'm thinking to at least 2 decimals of precision. 2) There should be a way to identify not just the asset creator but a group of creators or collaborators who are also credited. 3) In the case of multiple creators we might want to also to enable weighted distribution of the royalty.

Obviously this would all be optional, and need some information gathering at the time of asset creation, as well as some smart contracts capable of consuming this data. 

Im envisioning implementing metadata about people/groups for accreditation thusly:
Person/Organization Schema
```json
{
    "title": "Person, Group or Organization",
    "type": "object",
    "properties": {
        "name": {
            "type": "string",
            "description": "The name of the entity"
        },
        "description": {
            "type": "string",
            "description": "A simple description of the entity"
        },
        "address": {
            "type": "string",
            "description": "Optional blockchain wallet address for the entity"
        }
    }
}
``` 

And then somewhere in the Metadata:
```
...
        "creator": {
            "title": "Person, Group or Organization",
            "description": "The entity"
        },
        "royalty": {
            "type": "number",
            "description": "This value indicates the percentage of sales of this asset which are required by the creator. Value should be set with up to two decimals in precision."
        },
        "contributors": {
            "type": "array",
            "items": {
                "title": "Person, Group, or Organization"
            },
            "description": "List of all credited creators who are to be awarded credit for this work."
        },
...
```

barnjamin:
Good input, reformatting and commenting here
```
1) We need to know the royalty percent, I'm thinking to at least 2 decimals of precision. 
```
Seems reasonable, and that is assumed to be for final sale price right? Preventing asset xfer unless the royalty txn that meets the requirement is attached atomically?


```
2) There should be a way to identify not just the asset creator but a group of creators or collaborators who are also credited. With the option for weighted distribution of the royalty.
```
This makes sense. Right now transactions are limited to 16 grouped atomically, so in the case that there are more collaborators than that it will need to go to some multisig or other account. 

Also where would you expect the royalty info would be stored? The values will need to be encoded in whatever sc is validating the xfer to ensure the royalty is actually paid.

JuWeber99:
[...]

PR #70, opened 2022-02-28
ARC-0018 ASA Transfer Royalty Enforcement Specification

barnjamin:
Draft of Royalty Enforcement Specification using freeze/clawback roles to perform the transfer.

Example implementation: https://github.com/algorand-devrel/royalty

### DIFF
--- a/ARCs/arc-0018.md
+++ b/ARCs/arc-0018.md
@@ -65,7 +65,7 @@ The Royalty Collector is a generic account that could be set to a Single Signatu
 
 The Royalty Asset is an ASA subject to royalty collection and MUST be created with the the appropriate roles and parameters. 
 
-> Because the protocol does not allow updating a role after it's been deleted, if the creator thinks she may want to modify them later, they must be set to some non-zero address.
+> Because the protocol does not allow updating a role after it's been deleted, if the creator thinks they may want to modify them later, they must be set to some non-zero address.
 
 
 #### Asset Roles
--- a/ARCs/arc-0018.md
+++ b/ARCs/arc-0018.md
@@ -361,6 +361,38 @@ For an offered Asset, the authorizing address and amount offered should be store
 }
 ```
 
+
+#### Flow for a Marketplace
+
+```
+Let Alice be anyone with authority to update the Royalty Policies, shes specified herself as Royalty Recipient
+Let Bob be the Royalty Asset holder
+Let Carol be a buyer of a Royalty Asset
+```
+
+```mermaid
+sequenceDiagram
+    Alice->>Royalty Enforcer: set_policy with Royalty Recipient
+
+    par List
+        Bob->>Royalty Enforcer: offer
+        Bob->>Marketplace: list
+    end
+
+    Par Buy
+        Carol->>Marketplace: buy
+        Marketplace->>Royalty Enforcer: transfer
+        Bob->>Carol: clawback issued by Royalty Enforcer
+        Royalty Enforcer->>Alice: royalty payment
+    end
+
+    par Delist
+        Bob->>Royalty Enforcer: rescind
+        Bob->>Marketplace: delist
+    end
+```
+
+
 ### Metadata
 
 The metadata associated to an asset MAY conform to [ARC-3](./arc-0003.md) or [ARC-69](./arc-0069.md) with an additional field in the `attributes` section specifying the Application ID for the Royalty Enforcer Contract. 
--- a/ARCs/arc-0018.md
+++ b/ARCs/arc-0018.md
@@ -18,11 +18,11 @@ requires: ARC-4
 
 ## Abstract
 
-A specification to describe a set of methods that offer an API to enforce [Royalty Payments](https://en.wikipedia.org/wiki/Royalty_payment) to a Royalty Collector given a policy describing the royalty shares, both on primary and secondary sales.
+A specification to describe a set of methods that offer an API to enforce [Royalty Payments](https://en.wikipedia.org/wiki/Royalty_payment) to a Royalty Receiver given a policy describing the royalty shares, both on primary and secondary sales.
 
 ## Motivation
 
-This ARC is defined to provide a consistent set of asset configurations and ABI methods that, together, enable a royalty payment to a Royalty Collector.
+This ARC is defined to provide a consistent set of asset configurations and ABI methods that, together, enable a royalty payment to a Royalty Receiver.
 
 An example may include some music rights where the label, the artist, and any investors have some assigned royalty percentage that should be enforced on transfer. During the sale transaction, the appropriate royalty payments should be included or the transaction must be rejected.
 
@@ -128,6 +128,8 @@ A Marketplace may call the transfer request as long as the address initiating th
 
 The following is a set of methods that conform to the [ABI](arc-0004.md) specification meant to enable the configuration of a Royalty Policy and perform transfers.
 
+Any Inner Transactions that may be performed as part of the execution of the Royalty Enforcer application **SHOULD** set the fee to 0 and enforce fee payment through fee pooling by the caller.
+
 #### Set Policy  (REQUIRED):
     
 ```
@@ -152,7 +154,7 @@ The `payment_asset` argument represents the ASA id that is acceptable for paymen
 
 The `allowed` argument is a boolean representing whether or not this asset is allowed.
 
-The Royalty Collector **MUST** be opted into the full set of assets contained in this list of payment_assets. 
+The Royalty Receiver **MUST** be opted into the full set of assets contained in this list of payment_assets. 
 
 > In the case that an account is not opted into an asset, any transfers where payment is specified for that asset will fail until the account opts into the asset. or the policy is updated.
 
@@ -519,14 +521,14 @@ For an offered Asset, the authorizing address and amount offered should be store
 
 ```
 Let Alice be the creator of the Royalty Enforcer and Royalty Asset
-Let Alice also be the Royalty Recipient
+Let Alice also be the Royalty Receiver
 Let Bob be the Royalty Asset holder
 Let Carol be a buyer of a Royalty Asset
 ```
 
 ```mermaid
 sequenceDiagram
-    Alice->>Royalty Enforcer: set_policy with Royalty Basis and Royalty Recipient
+    Alice->>Royalty Enforcer: set_policy with Royalty Basis and Royalty Receiver 
     Alice->>Royalty Enforcer: set_allowed_asset with any asset that should be accepted as payment 
 
     par List
@@ -571,7 +573,7 @@ Example:
 
 The motivation behind defining a Royalty Enforcement specification is the need to guarantee a portion of a payment is received by select royalty collector on sale of an asset. Current royalty implementations are either platform specific or are only adhered to when an honest seller complies with it, allowing for the exchange of an asset without necessarily paying the royalties.
 
-The use of a smart contract as a clawback address is a guaranteed way to know an asset transfer is only ever made when certain conditions are met, or made in conjunction with additional transactions. The Royalty Enforcer is responsible for the calculations required in dividing up and dispensing the payments to the respective parties. The present specification does not impose any restriction on the Royalty Collector distribution logic (if any), which could be achieved through a Multi Signature account, a Smart Signature or even through another Smart Contract.
+The use of a smart contract as a clawback address is a guaranteed way to know an asset transfer is only ever made when certain conditions are met, or made in conjunction with additional transactions. The Royalty Enforcer is responsible for the calculations required in dividing up and dispensing the payments to the respective parties. The present specification does not impose any restriction on the Royalty Receiver distribution logic (if any), which could be achieved through a Multi Signature account, a Smart Signature or even through another Smart Contract.
 
 On Ethereum the EIP-2981 standard allows for ERC-721 and ERC-1155 interfaces to signal a royalty amount to be paid, however this is not enforced and requires marketplaces to implement and adhere to it.
 
--- a/ARCs/arc-0018.md
+++ b/ARCs/arc-0018.md
@@ -299,7 +299,7 @@ For an offered Asset, the authorizing address and amount offered should be store
     "name":"demo-arc-18",
     "methods": [
         {
-            "name": "set_policy",
+            "name": "set_royalty_policy",
             "desc": "Sets the royalty basis for a given app id",
             "args": [
                 {
@@ -308,10 +308,12 @@ For an offered Asset, the authorizing address and amount offered should be store
                 },
                 {
                     "name": "royalty_receiver",
-                    "type": "address"
+                    "type": "account"
                 }
             ],
-            "returns":"void"
+            "returns":{
+                "type":"void"
+            }
         },
         {
             "name":"set_payment_asset",
@@ -326,7 +328,9 @@ For an offered Asset, the authorizing address and amount offered should be store
                     "type":"boolean"
                 }
             ],
-            "returns":"void"
+            "returns":{
+                "type":"void"
+            }
         },
         {
             "name": "offer",
--- a/ARCs/arc-0018.md
+++ b/ARCs/arc-0018.md
@@ -350,6 +350,10 @@ The relevant keys are:
 
 `royalty_receiver` - The account that should be paid the royalty
 
+Another key is used to store the current administrator account:
+
+`administrator` - The account that is allowed to make administrative calls to this Royalty Enforcer application
+
 #### Local Storage
 
 For an offered Asset, the authorizing address and amount offered should be stored in a Local State field keyed by the 8 byte string representing the uint64 asset id of the asset being offered.
--- a/ARCs/arc-0018.md
+++ b/ARCs/arc-0018.md
@@ -305,7 +305,7 @@ The `offered_amount` is the number of units of this asset currently offered. Thi
 
 Three methods are specified here as `read-only` as defined in ARC22.
 
-#### Get Policy
+#### Get Policy (REQUIRED):
 ```
 get_policy()(address,uint64)
 ```
@@ -315,7 +315,7 @@ Gets the current [Royalty Policy](#royalty-policy) setting for this Royalty Enfo
 The return value is a tuple of type `(address,uint64)`, where the `address` is the [Royalty Receiver](#royalty-receiver) and the `uint64` is the [Royalty Basis](#royalty-basis).
 
 
-#### Get Offer
+#### Get Offer (REQUIRED):
 
 ```
 get_offer(
@@ -333,7 +333,7 @@ The `from` parameter is the account that placed the offer
 The return value is a tuple of type `(address,uint64)`, where `address` is the authorizing address that may make a transfer request and the `uint64` it the amount offered. 
 
 
-#### Get Administrator
+#### Get Administrator (OPTIONAL unless set_administrator is implemented then REQUIRED):
 
 ```
 get_administrator()address
--- a/ARCs/arc-0018.md
+++ b/ARCs/arc-0018.md
@@ -7,7 +7,7 @@ author:
     Ben Guidarelli (@barnjamin)
     Steve Ferrigno (@nullun)
     Adriano Di Luzio (@aldur)
-    Jason Weathersby (@todo)
+    Jason Weathersby (@JasonWeathersby)
 discussions-to: <URL>
 status: Draft
 type: Standards Track
@@ -30,13 +30,13 @@ An example may include some music rights where the label, the artist, and any in
 The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
 
 
-[Royalty Policy](#royalty-policy) - The name for the collection of settings that define how royalty payments are collected. 
+[Royalty Policy](#royalty-policy) - The name for the settings that define how royalty payments are collected. 
 
 [Royalty Enforcer](#royalty-enforcer) - The application that enforces the royalty payments given the Royalty Policy and performs transfers of the assets.
 
-[Royalty receiver](#royalty-receiver) - The account that receives the royalty payment. It can be any valid Algorand account. 
+[Royalty Receiver](#royalty-receiver) - The account that receives the royalty payment. It can be any valid Algorand account. 
 
-[Asset Subject to Royalties](#royalty-asset) - The asset that is subject to royalty payments on transfer
+[Royalty Asset](#royalty-asset) - The ASA that should have royalties enforced during a transfer. 
 
 [Asset Offer](#asset-offer) - A data structure stored in local state for the current owner representing the number of units of the asset being offered and the authorizing account for any transfer requests.
 
@@ -46,19 +46,19 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL 
 
 ```ts
 interface RoyaltyPolicy {
-    royalty_basis:     number,  // The percentage of the payment due, specified in basis points (0-10,000)
-    royalty_recipient: string,  // The address that should collect the payment
+    royalty_basis:     number   // The percentage of the payment due, specified in basis points (0-10,000)
+    royalty_recipient: string   // The address that should collect the payment
 }
 ```
 
-A Royalty Share consists of a `royalty_receiver` that should receive a Royalty payment and a `royalty_basis` representing some share of the total payment amount. The Royalty Share **MUST** be specified in terms of [basis points](https://en.wikipedia.org/wiki/Basis_point) of the payment amount.
+A Royalty Share consists of a `royalty_receiver` that should receive a Royalty payment and a `royalty_basis` representing some share of the total payment amount. The Royalty Basis **MUST** be specified in terms of [basis points](https://en.wikipedia.org/wiki/Basis_point) of the payment amount.
 
 
 ### Royalty Enforcer
 
-The Royalty Enforcer is an instance of a contract, an application, that controls the transfer of ASAs subject to the Royalty Policy. 
+The Royalty Enforcer is an instance of the contract, an Application, that controls the transfer of ASAs subject to the Royalty Policy. 
 
-This is accomplished by exposing an interface defined as a set of [ABI Methods](#abi-methods) allowing a grouped transaction call.
+This is accomplished by exposing an interface defined as a set of [ABI Methods](#abi-methods) allowing a grouped transaction call containing a payment and a [Transfer](#transfer) request.
 
 ### Royalty Receiver
 
@@ -68,16 +68,17 @@ The Royalty Receiver is then responsible for any further royalty distribution lo
 
 ### Royalty Asset 
 
-The Royalty Asset is an ASA subject to royalty payment collection and **MUST** be created with the the appropriate mutable and immutable parameters. 
+The Royalty Asset is an ASA subject to royalty payment collection and **MUST** be created with the the [appropriate parameters](#royalty-asset-parameters). 
 
-> Because the protocol does not allow updating an address parameter after it's been deleted, if the creator thinks they may want to modify them later, they must be set to some non-zero address.
+> Because the protocol does not allow updating an address parameter after it's been deleted, 
+> if the creator thinks they may want to modify them later, they must be set to some non-zero address.
 
 #### Asset Offer
 
 The Asset Offer is the a data structure stored in the owners local state. It is keyed in local storage by the byte string representing the ASA Id. 
 
 ```ts
-interface OfferedAsset {
+interface AssetOffer {
     auth_address:   string // The address of a marketplace or account that may issue a transfer request
     offered_amount: number // The number of units being offered
 }
@@ -89,7 +90,7 @@ In order for a transfer to occur, this field **MUST** be set and the parameters 
 
 > A transfer matching the offer would require the transfer amount <= offered amount and that the transfer is sent by auth_address
 
-#### Asset Address Parameters 
+#### Royalty Asset Parameters 
 
 The Clawback parameter **MUST** be set to the Application Address of the Royalty Enforcer.
 
@@ -105,8 +106,6 @@ The Manager parameter **MUST** be set to the Application Address of the Royalty 
 
 The Reserve parameter **MAY** be set to anything.
 
-#### Immutable Asset Parameters
-
 The `DefaultFrozen` **MUST** be set to true.
 
 ### Third Party Marketplace
@@ -117,10 +116,9 @@ A Marketplace may call the transfer request as long as the address initiating th
 
 ### ABI Methods
 
-#### Required Methods
+The following is a set of methods that conform to the [ABI](arc-0004.md) specification meant to enable the configuration of a Royalty Policy and perform transfers.
 
-
-##### Set Policy 
+#### Set Policy  (REQUIRED):
     
 ```
 set_policy(royalty_basis: uint64, royalty_recipients: account)
@@ -130,9 +128,7 @@ Sets the policy for any assets using this application as a Royalty Enforcer.
 
 The `royalty_basis` is the percentage for royalty payment collection, specified in basis points (e.g., 1% is 100).  
 
-A Royalty Basis **SHOULD** be immutable, if an application call is made that would overwrite an existing value it **SHOULD** fail. 
-
-> TODO: should this be a **MUST**? only should because a contract may want to lower it over time or absolve it completely
+A Royalty Basis **MUST** be immutable, if an application call is made that would overwrite an existing value it **MUST** fail. 
 
 > The immutability of a royalty basis is important since mutability introduces the possibility for a situation where, after an initial sale, the royalty policy is updated from 1% to 100% for example. This would make any further sales have the full payment amount sent to the royalty recipient and the seller would receive nothing.
 
@@ -143,7 +139,7 @@ The Royalty Collector **MUST** be opted into the full set of assets contained in
 > In the case that an account is not opted into an asset, any transfers where payment is specified for that asset will fail until the account opts into the asset. or the policy is updated.
 
 
-#### Set Payment Asset
+#### Set Payment Asset (REQUIRED):
 
 ```
 set_payment_asset(payment_asset: asset, allowed: boolean)
@@ -154,7 +150,7 @@ The `payment_asset` argument represents the ASA id that is acceptable for paymen
 The `allowed` argument is a boolean representing whether or not this asset is allowed.
 
 
-##### Transfer: 
+#### Transfer (REQUIRED): 
 
 ```
 transfer(
@@ -195,7 +191,7 @@ Prior to each transfer the the Royalty Enforcer **SHOULD** assert that `AuthAddr
 
 > This reasoning for this check is described in [Security Considerations](#security-considerations). It is purposely left to the implementor to decide if it should be checked.
 
-#### Offer:
+#### Offer (REQUIRED):
 
 ```
 offer(
@@ -237,12 +233,7 @@ If a [transfer](#transfer) or [royalty_free_move](#royalty-free-move) is called 
 
 > Exactly how to update the offer is left to the implementer. In the case of a partially filled offer, the amount may be updated to reflect some new amount that represents `offer_amount - amount transferred` or the offer may be deleted completely.
 
-
-#### Optional
-
-The following methods **MAY** be implemented but are not required by this specification.
-
-#### Get Offer:
+#### Get Offer (OPTIONAL):
 
 ```
 get_offer(from: account, royalty_asset: uint64)(address,uint64)
@@ -254,7 +245,7 @@ The `from` parameter is the account to inspect for an offer in local state with 
 
 The `royalty_asset` parameter is the ASA Id that has been offered.
 
-##### Royalty Free Move 
+#### Royalty Free Move (OPTIONAL) :
 
 ```
 royalty_free_move(
@@ -281,7 +272,6 @@ The `current_offered_amount` is the number of units of this asset currently offe
 The `current_auth_address` is the authorized address for the current offer, the zero address is passed if no offer is outstanding.
 
 > The `current_*` values for the offer are passed to prevent the race or attack described in [Security Considerations](#security-considerations).
-
 
 
 ### Storage
@@ -485,7 +475,8 @@ For an offered Asset, the authorizing address and amount offered should be store
 #### Example Flow for a Marketplace
 
 ```
-Let Alice be anyone with authority to update the Royalty Policies, shes specified herself as Royalty Recipient
+Let Alice be the creator of the Royalty Enforcer and Royalty Asset
+Let Alice also be the Royalty Recipient
 Let Bob be the Royalty Asset holder
 Let Carol be a buyer of a Royalty Asset
 ```
@@ -546,9 +537,10 @@ https://github.com/algorand-devrel/royalty
 
 ## Security Considerations
 
-# TODO: make comments about manager being set && changing the clawback or metadata on arc69
 
-There are a number of ways to circumvent the enforcement of the Royalty Policy.
+There are a number of security considerations that implementers and users should be aware of. 
+
+*Spoofed payment*
 
 While its possible to enforce the group size limit, it is possible to circumvent the royalty enforcement logic by simply making an Inner Transaction application call with the appropriate parameters and a small payment, then in the same outer group the "real" payment. The counter-party risk remains the same since the inner transaction is atomic with the outers. 
 
@@ -563,15 +555,27 @@ Some ways to prevent a small royalty payment and larger payment in a later trans
 
 The exact logic that should determine _if_ a transfer should be allowed is left to the implementer.
 
+*Rekey to swap*
+
 Rekeying an account can also be seen as circumventing this logic since there is no counter-party risk given that a rekey can be grouped with a payment. We address this by suggesting the `auth_addr` on the buyer and seller accounts are both set to the zero address.
 
+*Offer for unintended clawback*
+
 Because we use the clawback mechanism to move the asset, we need to be sure that the current owner is actually interested in making the sale. We address this by requiring the [offer](#offer) method is called to set an authorized address OR that the AssetSender is the one making the application call. 
+
+*Offer double spend*
 
 If the [offer](#offer) method did not require the current value be passed, a possible attack or race condition may be taken advantage of.
 
 - There's an open offer for N.
 - The owner decides to lower it to N < M < 0
 - I see that; decide to "frontrun" the second tx and first get N, [here the ledger should apply the change of offer, which overwrites the previous value — now 0 — with M], then I can get another M of the asset.
+
+*Mutable asset parameters*
+
+If the ASA has it's manager parameter set, it is possible to change the other address parameters. Namely the clawback and freeze roles could be changed to allow an address that is _not_ the Royalty Enforcer's application address.  For that reason the manager **MUST** be set to the zero address or to the Royalty Enforcer's address.
+
+In the case of [ARC69](arc-0069.md) ASA's the manager is the account that may issue `acfg` transactions to update metadata, so some consideration will need to be made for those assets.
 
 ## Copyright
 
--- a/ARCs/arc-0018.md
+++ b/ARCs/arc-0018.md
@@ -123,23 +123,18 @@ The following is a set of methods that conform to the [ABI](arc-0004.md) specifi
 #### Set Policy  (REQUIRED):
     
 ```
-set_policy(royalty_basis: uint64, royalty_recipient: account)
+set_royalty_policy(royalty_basis: uint64, royalty_recipient: account)
 ```
 
 Sets the policy for any assets using this application as a Royalty Enforcer.
 
 The `royalty_basis` is the percentage for royalty payment collection, specified in basis points (e.g., 1% is 100).  
 
-A Royalty Basis **MUST** be immutable, if an application call is made that would overwrite an existing value it **MUST** fail. 
+A Royalty Basis **SHOULD** be immutable, if an application call is made that would overwrite an existing value it **SHOULD** fail. 
 
 > The immutability of a royalty basis is important since mutability introduces the possibility for a situation where, after an initial sale, the royalty policy is updated from 1% to 100% for example. This would make any further sales have the full payment amount sent to the royalty recipient and the seller would receive nothing.
 
 The `royalty_receiver` is the address of the account that should receive a partial share of the payment for any [transfer](#transfer) of an asset subject to royalty collection.
-
-The Royalty Collector **MUST** be opted into the full set of assets contained in this list of payment_assets. 
-
-> In the case that an account is not opted into an asset, any transfers where payment is specified for that asset will fail until the account opts into the asset. or the policy is updated.
-
 
 #### Set Payment Asset (REQUIRED):
 
@@ -151,6 +146,9 @@ The `payment_asset` argument represents the ASA id that is acceptable for paymen
 
 The `allowed` argument is a boolean representing whether or not this asset is allowed.
 
+The Royalty Collector **MUST** be opted into the full set of assets contained in this list of payment_assets. 
+
+> In the case that an account is not opted into an asset, any transfers where payment is specified for that asset will fail until the account opts into the asset. or the policy is updated.
 
 #### Transfer (REQUIRED): 
 
--- a/ARCs/arc-0018.md
+++ b/ARCs/arc-0018.md
@@ -117,9 +117,25 @@ The `royalty_share` is the share of the payment that should be sent to the `roya
 
 The `allowed_assets_n` arguments represents the ASA ids that are acceptable for payment. The contract logic **MUST** opt into the set of assets specified in order to accept the payments as part of a transfer.
 
+The Royalty Collector SHOULD be opted into the full set of assets contained in this list. In the case that an account is not opted into an asset, any transfers where payment is specified for that asset will fail until the account opts into the asset.
+
+#### Get Policy
+
+```
+    get_policy(royalty_asset: uint64)(address,uint64,uint64,uint64,uint64,uint64)
+```
+
+Gets the royalty for a given asset id.
+
+> This method is *READ ONLY* as defined in [arc22](./arc-0022.md)
+
+The `royalty_asset` parameter is the ASA Id of the royalty asset to get the policy for.
+
+The default policy is returned unless a policy is defined for the passed ASA Id. If no default policy is returned all return values are set to the zero value for their type.
+
+
 > TODO: since assets cant be in dynamic arrays, this bit is annoying. We could define N selectors (up to 8 for assets) and dispatch them all to the same set of logic? for now just hardcoded to 4 and caller expected to pass 0s if they don't need all 4.  As of https://github.com/algorandfoundation/ARCs/pull/77, this limitation has been changed. Update spec accordingly once SDK support is available.
 
->The Royalty Collector SHOULD be opted into the full set of assets contained in this list. In the case that an account is not opted into an asset, any transfers where payment is specified for that asset will fail until the account opts into the asset.
 
 
 ##### Transfer: 
@@ -181,6 +197,8 @@ Removes the asset from being offered by the sender.
 
 The `royalty_asset` is the ASA ID that is being rescinded.
 
+
+
 #### Optional
 
 The following methods **MAY** be implemented but are not required by this specification.
@@ -227,6 +245,29 @@ For an offered Asset, The authorizing address and amount offered should be store
     "name":"demo-arc-18",
     "methods":[
         {
+            "name":"set_policy",
+            "desc":"Sets the royalty policy for a given asset id",
+            "args":[
+                {"name":"royalty_asset", "type":"asset"}, 
+                {"name":"royalty_receiver", "type":"address"},
+                {"name":"royalty_share", "type":"uint64"},
+                {"name":"allowed_asset_1", "type":"asset"},
+                {"name":"allowed_asset_2", "type":"asset"},
+                {"name":"allowed_asset_3", "type":"asset"},
+                {"name":"allowed_asset_4", "type":"asset"}
+            ],
+            "returns":{"type":"void"}
+        },
+        {
+            "name":"get_policy",
+            "desc":"Gets the royalty policy for a given asset id",
+            "readonly":true,
+            "args":[
+                {"name":"royalty_asset", "type":"uint64"}, 
+            ],
+            "returns":{"type":"(address,uint64,uint64,uint64,uint64,uint64)"}
+        },
+        {
             "name":"offer",
             "desc":"Flags that an asset is offered for sale and sets address authorized to submit the transfer",
             "args":[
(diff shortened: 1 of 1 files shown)


JuWeber99:
I really like it and I think it’s already quite good. My idea whould be not only to enforce that the „owner“ has to initiate the transfer call but it’s possible to let the owner set a „bookkeeper“ account. This is especially useful if inner Transactions get more widespread. This way a user can authorize an offer address for NFTs like an auction to be the bookkeeper/authorizer for the next transfer. What do you think?

barnjamin:
Wallets should have an easy way to determine if an asset is subject to royalty polices and some way to determine _which_ app is responsible for the authorization of transfers. We might consider something like arc3 where the `#arc18` is listed at the end of the URL and the metadata contains information about the Royalty Enforcement application.

Currently the "seller" of the asset needs to sign the application call to initiate the Transfer. This is suboptimal since sales can't happen asynchronously.  Ideally the asset can be marked as "for sale" with some indication of the preconditions that should be met for the Transfer to be initiated.  
We could consider using local state for app/seller to hold the details of the preconditions to be met like that the sender of the Transfer is a specific marketplace address and if those conditions are met the Transfer is allowed to take place asynchronously, without a signature from the seller.

Curtis-D:
It is probably worth checking the group size on the transfer call (and possibly others) so a user can't "sell" the NFT for 1 ALGO meanwhile allowing other transactions in the group to pay for actual price of the NFT. Similar to an atomic swap but with a sale of 1 ALGO

ARCs/arc-0018.md
```diff
@@ -0,0 +1,211 @@
...
+### Royalty Collector
+
+The Royalty Collector is a generic account that could be set to a Single Signature, a Multi Signature, a Smart Signature or even to another Smart Contract. The Royalty Collector is then responsible for the royalty distribution logic, making the Royalty Enforcement Specification more general and composable.
+
+
+### Royalty Asset 
+
+The Royalty Asset is an ASA subject to royalty collection and MUST be created with the the appropriate roles and parameters. 
+
+> Because the protocol does not allow updating a role after it's been deleted, if the creator thinks she may want to modify them later, they must be set to some non-zero address.
```
nullun, line 69: Are we talking explicitly about the clawback address here?

barnjamin: clawback/freeze both I think

cusma:
[Here](https://github.com/algorandfoundation/ARCs/pull/70/commits/c98c6bd8c6bc5f4d7b59706ff2496cb12a9a020c#diff-1426c1290b01352591253416bac2473fa879076ef13fd7a1b2a0ac110a4584a0R40) we say:

> A Royalty Share consists of an `address` that should receive a Royalty payment and a `share`

But then the structure is:

```ts
type RoyaltyPolicy {
    share:              number,  // The share of the payment due, specified in basis points
    royalty_collector:  string,  // The address that should collect this share of the payment
    allowed_assets:     number[] // The list of asset ids that are acceptable for payment
}
```

I think that, given the introduction of a `Royalty Collector` the `Royalty Share` should be just a `number` right?

cusma:
Quick sanity check: suppose I own a `Royalty ASA`, the royalty is payed with another `Share ASA`. If the `Royalty Collector` opts-out the `Share ASA` any transfer of `Royalty ASA` would fail. 

Should we have any countermeasure if this happens? 

Suggestion: If this happens I would expect that the `Royalty ASA` is now “royalty free” (since the Royalty Collector doesn't want to receive `Share ASA` at all). So if `Royalty Enforcer` looks at `Royalty Collector` and see that it opted-out the `Share ASA` then `move` method should be available to anyone.

WDYT?

ARCs/arc-1337.md
```diff
@@ -0,0 +1,170 @@
...
+The `asset` is the ASA id of the asset the policy is set for.
+
+The `royalty_shares` are encoded as an array of `tuple(address, uint64)` that represents the Royalty Recipient and Royalty Share in basis points. The address of a Royalty Recipient SHOULD NOT be repeated, only the last occurence of the address will be used for their Royalty Share. 
+
+> The maximum number of Royalty Recipients is limited to the number of foreign accounts that may be passed from a grouped set of application call transactions.  As of go-algorand v3.4.0 this number is:
+> 4 [max accts per call] * (16 [max calls per group] - 1 [transaction reserved for payment]) = 60
+
+The `allowed_assets` list contains the ASA ids that are acceptable for payment. The contract logic MUST opt into the set of assets specified in order to accept the payments as part of a transfer.
+
+> All the Royalty Recipients SHOULD be opted into the full set of assets contained in this list but MAY NOT be checked ahead of time. In the case that an account is not opted into an asset, any transfers where payment is specified for that asset will fail until the account opts into the asset.
```
cusma: Quick sanity check: suppose I own a `Royalty ASA`, the royalty is payed with another `Share ASA`. If the `Royalty Collector` opts-out the `Share ASA` any transfer of `Royalty ASA` would fail. 

Should we have any countermeasure if this happens? 

Suggestion: If this happens I would expect that the `Royalty ASA` is now “royalty free” (since the `Royalty Collector` doesn't want to receive `Share ASA` at all). So if `Royalty Enforcer` looks at `Royalty Collector` and see that it opted-out the `Share ASA` then `move` method should be available to anyone.

WDYT?

ARCs/arc-0018.md
```diff
@@ -125,13 +131,35 @@ The `from` parameter is the address for the account the ASA is transferred from.
...
+Upon each transfer the the Royalty Enforcer MUST assert that `AuthAddr == AssetSender` for the Seller (`from` parameter) and `AuthAddr == AssetReceiver` for the Buyer (`to` parameter). In other words Seller and Buyer accounts not have been Rekeyed. Exceptions can be handled with `Move` method.
+
+> The rational for this check is ensuring that `Rekeying` is not exploited to transfer the `royalty_asset` ownership circumventing the Royalty Policy. Two parties involved in the trade could create a Group Transactions in which: 1. Seller rekeys the `from` to the Buyer `to`, 2. Buyer sends payment to `other_seller_account`. Note that nothing can prevent the users to execute such trade but the Royalty Enforcer tries to dissuade this behavior: if such trade happens, the Receiver will no longer be able to re-sell the Royalty ASA in the regular way.
+
+
+#### Offer:
+
+    offer(royalty_asset: asset, authorized_address: account)
```
cusma, line 152: See this [comment](https://github.com/algorandfoundation/ARCs/pull/70#discussion_r838336925). I think we should add `RoyaltyAssetAmount` here too.

ARCs/arc-1337.md
barnjamin: Hm, I think we should fail at the protocol level if this happens. It may be the case that the `Royalty Collector` no longer wants to accept that asset and they've just not updated the `Royalty Policy` yet. 

If we default to move then accidentally opting out of the asset will make them all royalty free until the `Royalty Collector` opts back in or the `Royalty Policy` is updated.  Unless someone is watching closely they might not notice this condition if we allow the transaction to succeed by moving but they'd definitely notice if folks were complaining about failed transactions.

There is a broader point here where if some pre-condition fails folks aren't able to trade the asset anymore, which I think _should_ have more thought put into it.

cusma: > There is a broader point here where if some pre-condition fails folks aren't able to trade the asset anymore, which I think _should_ have more thought put into it.

Exactly, the corner case I want to avoid (and be able to handle somehow) is the "lock-in effect" by which people are no longer able to transfer and trade their ASA not because they refuse to comply with the Royalty Policy but because that Policy is broken by the creator ex-post.

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+Upon each transfer the the Royalty Enforcer SHOULD assert that `AuthAddr == AssetSender` for the Seller (`from` parameter) and `AuthAddr == AssetReceiver` for the Buyer (`to` parameter). In other words Seller and Buyer accounts not have been Rekeyed. Exceptions can be handled with `Move` method.
+
+> The rational for this check is ensuring that `Rekeying` is not exploited to transfer the `royalty_asset` ownership circumventing the Royalty Policy. Two parties involved in the trade could create a Group Transactions in which: 1. Seller rekeys the `from` to the Buyer `to`, 2. Buyer sends payment to `other_seller_account`. Note that nothing can prevent the users to execute such trade but the Royalty Enforcer tries to dissuade this behavior: if such trade happens, the Receiver will no longer be able to re-sell the Royalty ASA in the regular way.
```
fabrice102, line 149: This does not prevent the use of smart contract controlling the account holding the asset.
That being said, such an attack requires all the users to agree on using such a smart contract from the beginning, so I guess this is unlikely to be a real threat.

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+    offer(royalty_asset: asset, royalty_asset_amount: uint64, authorized_address: account)
+
+Flags the asset as transferrable and sets the address that may initiate the transfer request.
+
+The `royalty_asset` is the ASA ID that is offered. It MUST have a policy defined prior to attempting to offer
+
+The `royalty_asset_amount` is the number of units of the ASA ID that are offered. The account making this call MUST have at least this amount. 
+
+The `authorized_address` is the address that may initiate the transfer. This address may be any valid address in the Algorand network including and Application Account's address.
```
fabrice102, line 163:
```suggestion
The `authorized_address` is the address that may initiate the transfer. This address may be any valid address in the Algorand network including an Application Account's address.
```

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+Upon each transfer the the Royalty Enforcer SHOULD assert that `AuthAddr == AssetSender` for the Seller (`from` parameter) and `AuthAddr == AssetReceiver` for the Buyer (`to` parameter). In other words Seller and Buyer accounts not have been Rekeyed. Exceptions can be handled with `Move` method.
+
+> The rational for this check is ensuring that `Rekeying` is not exploited to transfer the `royalty_asset` ownership circumventing the Royalty Policy. Two parties involved in the trade could create a Group Transactions in which: 1. Seller rekeys the `from` to the Buyer `to`, 2. Buyer sends payment to `other_seller_account`. Note that nothing can prevent the users to execute such trade but the Royalty Enforcer tries to dissuade this behavior: if such trade happens, the Receiver will no longer be able to re-sell the Royalty ASA in the regular way.
+
+
+
+#### Offer:
+
+    offer(royalty_asset: asset, royalty_asset_amount: uint64, authorized_address: account)
```
fabrice102, line 155: What happens if called multiple times?

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+The `allowed_assets_n` arguments represents the ASA ids that are acceptable for payment. The contract logic MUST opt into the set of assets specified in order to accept the payments as part of a transfer.
+
+> TODO: since assets cant be in dynamic arrays, this bit is annoying. We could define N selectors (up to 8 for assets) and dispatch them all to the same set of logic? for now just hardcoded to 4 and caller expected to pass 0s if they don't need all 4.  As of https://github.com/algorandfoundation/ARCs/pull/77, this limitation has been changed. Update spec accordingly once SDK support is available.
+
+>The Royalty Collector SHOULD be opted into the full set of assets contained in this list but MAY NOT be checked ahead of time. In the case that an account is not opted into an asset, any transfers where payment is specified for that asset will fail until the account opts into the asset.
+
+
+##### Transfer: 
+
+    transfer(royalty_asset: asset, from: account, to: account, royalty_collector: account, royalty_asset_amount: uint64, payment: transaction)
```
fabrice102, line 123: What happens of not all the offered amount is used?
Can a second transfer take place?
Or does it rescind the offer?

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+    create_nft()uint64
+
+Creates an NFT with the appropriate parameters set. 
+
+Returns the Asset ID that was just created
+
+##### Move 
+
+    move(royalty_asset: asset, royalty_asset_amount: uint64, to: account)
```
fabrice102, line 188: Maybe `royalty_free_move`?

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+    "locals": [ 
+        {
+            "name": "{uint64 asset id as big endian bytes}",
+            "desc": "The ID of the asset that has been offered as key, authorizing address as value",
+            "type": "(address,uint64)"
+        }  
+    ],
+    "globals": [ 
+        {
+            "name": "0000000000000000",
```
fabrice102, line 297: Is it really this or the 0 byte to be consistent with the royalty policy for asset ID in key

fabrice102: See also https://github.com/algorandfoundation/ARCs/pull/75#discussion_r839006561

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+The `royalty_asset` is the asset being transferred without applying the Royalty Enforcement logic.
+
+The `royalty_asset_amount` is the number of units of this ASA ID that should be moved.
+
+The `to` parameter is the intended receiver of the asset. This account MUST be opted into the asset prior to calling this method or it will fail at the protocol level.
+
+
+### Storage
+
+#### Global Storage
```
fabrice102, line 201: Specifying global storage is good for two reasons:
1. allowing users to use it to get info (which requires to enter into the level of details you provide)
2. allowing users to know when to include a foreign account in the transaction (which only requires to specify which application call requires which local account).

2 is very important.
1 is useful for actual applications but I'm afraid that in a standard / interface it may prevent some optimizations that applications may do.
Instead, using read-only methods may allow more flexibility.

What do you think?

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+}
+```
+
+### Metadata
+
+The metadata associated to an asset MAY conform to [ARC-3](./arc-0003.md) or [ARC-69](./arc-0069.md) with an additional field in the `attributes` section specifying the Application ID for the Royalty Enforcer Contract. 
+
+```js
+//...
+"attributes":{
```
fabrice102, line 316: I think we should use `properties`, not `attributes`.
That would also be better re https://github.com/algorandfoundation/ARCs/pull/61/files

One issue is that ARC-69 only allows simple properties, but maybe we can convince them otherwise.

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+The use of a smart contract as a clawback address is a guaranteed way to know an asset transfer is only ever made when certain conditions are met, or made in conjunction with additional transactions. The Royalty Enforcer is responsible for the calculations required in dividing up and dispensing the payments to the respective parties. The present specification does not impose any restriction on the Royalty Collector distribution logic (if any), which could be achieved through a Multi Signature account, a Smart Signature or even through another Smart Contract.
+
+On Ethereum the EIP-2981 standard allows for ERC-721 and ERC-1155 interfaces to signal a royalty amount to be paid, however this is not enforced and requires marketplaces to implement and adhere to it.
+
+## Backwards Compatibility
+
+Existing ASAs with unset freeze and clawback roles will be incompatible with this specification.
```
fabrice102, line 335:
```suggestion
Existing ASAs with unset clawback address or unset manager address (in case the clawback address is not the application account of a smart contract that is updatable - which is most likely the case) will be incompatible with this specification.
```

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+## Reference Implementation
+
+https://github.com/algorand-devrel/royalty
+
+## Security Considerations
```
fabrice102, line 341: See also comments above.

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+### Royalty Policy 
+
+A Royalty Share consists of an `address` that should receive a Royalty payment and a `share` representing some share of the total payment amount. The Royalty Share MUST be specified in terms of [basis points](https://en.wikipedia.org/wiki/Basis_point) of the payment amount.
+
+A Royalty Policy is specified as a Royalty Collector and associated Royalty weighting. Additionally, if any assets besides Algos are acceptable, 
+
+
+```ts
+type RoyaltyPolicy {
+    share:              number,  // The share of the payment due, specified in basis points
```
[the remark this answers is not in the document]
cusma: Yes, I think we shouldn't specify a range for this at all. Maybe is not common but what if I want to set a royalty of 150%? I think we should stay general and not restrict this. Each implementation then must just make sure that the calculation is correct (does not overflow, etc.).

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+## Specification
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+
+[Royalty Policy](#royalty-policy) - The data structure that defines how royalties are dispersed
+
+[Royalty Enforcer](#royalty-enforcer) - The contract that enforces the royalty payments given the Royalty Policy and performs transfers of the assets.
+
+[Royalty Collector](#royalty-collector) - The account that collect the royalty and provides the royalty distribution logic. It could be set to a Single Signature, a Multi Signature, a Smart Signature or even another Smart Contract.
```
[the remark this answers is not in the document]
cusma: > One thing that we may want to require however is that it necessarily opted in to the allowed assets for payments.
> Also, we may want to specify what happens if it does not:
> does it mean the asset is no more transferrable? or that royalties are not collected?

[From my comment here](https://github.com/algorandfoundation/ARCs/pull/70#discussion_r838331343): I want to avoid (and be able to handle somehow) the "lock-in effect" by which people are no longer able to transfer and trade their ASA not because they refuse to comply with the `Royalty Policy` but because that Policy is broken by the creator ex-post.

If `Royalty Collector` opts-out all the `Share ASA` I would expect that the `Royalty ASA` is now “royalty free” (since the `Royalty Collector` doesn't want to receive Share ASA at all).

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+### Royalty Asset 
+
+The Royalty Asset is an ASA subject to royalty collection and MUST be created with the the appropriate roles and parameters. 
+
+> Because the protocol does not allow updating a role after it's been deleted, if the creator thinks she may want to modify them later, they must be set to some non-zero address.
+
+
+#### Asset Roles
+
+The Clawback role MUST be set to the Application Address.
```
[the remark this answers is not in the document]
cusma: As you say, this introduce complexity but maybe we can be permissive, like:

```
The Clawback role **MUST** be set to the Application Address of the Royalty Enforcer or to any account rekeyed to the Royalty Enforcer.

The first option **SHOULD** be always preferred when possible.
```

I'm wondering if this complicates or not the "discovery" phase for wallets/app that want to learn how to interact with the Royalty ASA.

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+#### Required
+
+
+##### SetPolicy
+    
+    set_policy(royalty_asset: asset, royalty_receiver: address, royalty_share: uint64, allowed_asset_1: asset, allowed_asset_2: asset, allowed_asset_3: asset, allowed_asset_4: asset)
+
+Sets the Royalty Policy for a given asset. 
+
+The `royalty_asset` is the ASA id of the asset the policy is set for.  If the `royalty_asset` is 0, the policy is defined for _any_ asset that has its freeze and clawback set to the application address.
```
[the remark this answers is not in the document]
cusma: > the policy is defined for _any_ asset that has its freeze and clawback set to the application address.

We should not mandate `FreezeAddress` AND `ClawbackAddress` set to Application Address, but just `ClawbackAddress` since the `FreezeAddress` could be set to `ZeroAddress`.

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+A Royalty Share consists of an `address` that should receive a Royalty payment and a `share` representing some share of the total payment amount. The Royalty Share MUST be specified in terms of [basis points](https://en.wikipedia.org/wiki/Basis_point) of the payment amount.
+
+A Royalty Policy is specified as a Royalty Collector and associated Royalty weighting. Additionally, if any assets besides Algos are acceptable, 
+
+
+```ts
+type RoyaltyPolicy {
+    share:              number,  // The share of the payment due, specified in basis points
+    royalty_collector:  string,  // The address that should collect this share of the payment
+    allowed_assets:     number[] // The list of asset ids that are acceptable for payment
```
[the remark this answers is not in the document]
barnjamin: Algo would not need to be specified, by default its accepted. This list is just for ASA ids that may also be accepted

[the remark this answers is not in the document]
nullun: Is it possible to only accept an ASA and not Algos?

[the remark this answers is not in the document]
barnjamin: Not as written, but I guess we could add it? I like the idea that there will always be one possible payment that will always be accepted though.

ARCs/arc-0018.md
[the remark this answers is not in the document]
barnjamin: Where would the extra 50% payment come from?

fergalwalsh:
> Wallets should have an easy way to determine if an asset is subject to royalty polices and some way to determine _which_ app is responsible for the authorization of transfers. We might consider something like arc3 where the `#arc18` is listed at the end of the URL and the metadata contains information about the Royalty Enforcement application. This would conflict with the current #arc3 suffix though, more thought needed.

This is an important issue that needs to be addressed. Currently I don't see how wallets/marketplaces will know that a given asset is adhering to this standard. The use of a suffix in the URL is not viable given that this is required by ARC3. Requiring a specific property in metadata is a possibility but it's hard to see how to handle all possible current and future metadata formats. But without a strict standard on this wallets won't have a way to identify these assets and users will have a poor experience when they try and fail to use normal AssetTransfer functionality.

ARCs/arc-0018.md
barnjamin: In the current implementation, its 8 0 bytes representing the uint64 zero value to be consistent. We could rename to just `default` or something?

barnjamin:
@fergalwalsh I agree. 

Considering repurposing the `reserve` address like arc19 for our own needs. 

In our case set the reserve address to some set of bytes that represents what spec/interface this asset adheres to as well as a uint64 representing the app id that is acting as the enforcer

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+> While its possible to enforce the group size limit, it is possible to circumvent the royalty enforcement logic by simply making an Inner Transaction application call with the appropriate parameters and a small payment, then in the same outer group the "real" payment. The counter-party risk remains the same since the inner transaction is atomic with the outers. 
+> Some ways to prevent a small royalty payment and larger payment in a later transaction of the same group might be by using an `allow` list that is checked against the `auth_addr` of the offer call. The `allow` list would be comprised of known and trusted marketplaces that do not attempt to circumvent the royalty policy. The `allow` list may be implicit as well by transferring a specific asset to the `auth_addr` as frozen and on `offer` a the balance must be > 0 to allow the `auth_addr` to be persisted.
+
+Upon each transfer the the Royalty Enforcer SHOULD assert that `AuthAddr == AssetSender` for the Seller (`from` parameter) and `AuthAddr == AssetReceiver` for the Buyer (`to` parameter). In other words Seller and Buyer accounts not have been Rekeyed. Exceptions can be handled with `Move` method.
```
fergalwalsh, line 147: I'm not clear exactly what the intention is here. Do you want to ensure that the Seller or Buyer has never been rekeyed? Or just not within the current group?

If it is a requirement that the accounts have never been required this seems unreasonable given that Rekeying is a very commonly used standard feature of Algorand.

fergalwalsh:
It's not clear to me yet what the standard flow of events is with this mechanism. I think this needs to be described in this document. i.e the parties involved and the sequence of transactions/app calls they make. It would be important to detail this both for peer-to-peer sales, marketplace listings and marketplace auctions.

ARCs/arc-0018.md
barnjamin: Right, this a non-required bit of the spec that calls out a possible way to effectively transfer the asset without going through whatever pre-requisites are required by the contract. 

A holder of this asset (or more generally any controlled asset, think of a security) might need to go through some KYC/AML that allows them to hold it. Using the rekey feature in an atomic transaction allows an account to atomically "sell" the asset.

I don't want to make it required here and think most implementers probably don't need to do this

barnjamin:
>It's not clear to me yet what the standard flow of events is with this mechanism. I think this needs to be described in this document. i.e the parties involved and the sequence of transactions/app calls they make. It would be important to detail this both for peer-to-peer sales, marketplace listings and marketplace auctions.

Good point, I'll add a swimlane chart. See the example here: https://github.com/algorand-devrel/royalty for a really basic implementation

fergalwalsh:
I'm also not clear how many implementations of this app need to exist. Could there be one global Enforcer that everyone uses or is it an app created per asset/ per creator? Or only different apps if people want slightly different functionality? This confuses me right now so probably it could be specified more clearly somewhere.

barnjamin:
>I'm also not clear how many implementations of this app need to exist. Could there be one global Enforcer that everyone uses or is it an app created per asset/ per creator? Or only different apps if people want slightly different functionality? This confuses me right now so probably it could be specified more clearly somewhere.

The spec allows for a default policy to be set for an application. If that suits your needs and you don't need to define a different policy per asset you need only 1 app. If you _do_ need to define a per asset policy you may store up to 64 in a single contract. Beyond that, you'd need to set up additional apps to handle specific policies.

One possible scenario is you have say 1000 different asset ids broken into groups of 5 common policies. In that scenario I think it'd make senses to have 5 apps with a default policy set for each and assign freeze/clawback/reserve appropriately to reference the application that holds the relevant default policy.

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+    create_nft()uint64
+
+Creates an NFT with the appropriate parameters set. 
+
+Returns the Asset ID that was just created
+
+##### Move 
+
+    move(royalty_asset: asset, royalty_asset_amount: uint64, to: account)
```
fergalwalsh, line 188: Is this optional or required? I think the heading levels are confused in this section.

The existence of this method seems to negate any concerns about rekeying or inner txns or allow lists given that there is an explicit method for royalty free transfers. 

If this is optional what is the intended flow with assets without this method for marketplaces that use escrows? Or for users moving between their own accounts? Is a 0 amount transfer allowed and recommended here?

barnjamin: This is optional and acts as an escape valve to transfer the asset without employing Royalty Enforcement logic.

It can be useful for "freeebies" or just to distribute the asset initially. Provided here as an implementation suggestion but not requirement

ARCs/arc-0018.md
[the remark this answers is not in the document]
cusma: The situation I'm describing here is the following: suppose for a given Royalty ASA I want to enforce a "fixed" royalty (or tax), regardless the amount of specific trades. Example: the price of the trade is 100 _units_ but the fixed "tax" on transfer is always 150 _units_.

I know it sounds uncommon, but should this spec accomodate this use-case too?

ARCs/arc-0018.md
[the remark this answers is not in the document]
barnjamin: @fabrice102 this would be default if no specific policy is defined for a given Asset ID

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+
+### ABI Methods
+
+#### Required
+
+
+##### SetPolicy
+    
+    set_policy(royalty_asset: asset, royalty_receiver: address, royalty_share: uint64, allowed_asset_1: asset, allowed_asset_2: asset, allowed_asset_3: asset, allowed_asset_4: asset)
```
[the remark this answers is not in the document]
barnjamin: I think we'd want to pass it in this case since we may want to check its parameters and opt the application into the asset.

ARCs/arc-0018.md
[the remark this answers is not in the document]
barnjamin: I don't think that'd be a common enough case to address in this spec

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+    rescind(royalty_asset: asset)
+
+Removes the asset from being offered by the sender.
+
+The `royalty_asset` is the ASA ID that is being rescinded.
+
+#### Optional
+
+#### Create NFT
```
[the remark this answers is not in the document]
barnjamin: I put it in the `optional` section but I can remove completely if you think it might be confusing

ARCs/arc-0018.md
barnjamin: In the current reference implementation the offered amount is updated to reflect the new offered amount but maybe we should be more explicit about it? It may also make sense to just wipe it completely.

I'm not sure what the correct behavior is, wdyt?

ARCs/arc-0018.md
```diff
@@ -0,0 +1,348 @@
...
+
+The `royalty_asset_amount` parameter is the number of units of this ASA ID to transfer. The amount MUST be less than or equal to the amount `offered` by the `from` account.
+
+The `payment` parameter is a reference to the transaction that is transferring some asset (ASA or Algos) from the buyer to the Application Address. 
+
+The `payment_asset` parameter is used in the case that the payment is being made with some asset rather than with Algos. It MUST match the Asset ID of the AssetTransfer payment transaction. 
+
+The transfer call SHOULD be part of a group with a size of 2 (app call + payment/asset transfer)
+
+> While its possible to enforce the group size limit, it is possible to circumvent the royalty enforcement logic by simply making an Inner Transaction application call with the appropriate parameters and a small payment, then in the same outer group the "real" payment. The counter-party risk remains the same since the inner transaction is atomic with the outers.
```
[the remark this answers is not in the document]
barnjamin: For the lower limit we'd have to have it specified for each Allowed Asset defined in the policy.  If some volatile asset is allowed the lower limit may be outside the perceived value of the Royalty Asset being transferred. 

In general I think I'd leave this to the implementer to decide _if_ a transfer should take place given some additional custom logic.

ARCs/arc-0018.md
barnjamin: I agree, I've added read only methods for both policy and offer

ARCs/arc-0018.md
barnjamin: I've updated the document to describe how the implementer might consider updating the offer after a transfer takes place.

ARCs/arc-0018.md
```diff
@@ -0,0 +1,449 @@
...
+
+> Since the Royalty Enforcer relies on using the Clawback mechanism to perform the transfer the Clawback should NEVER be set to the zero address. 
+
+The Freeze parameter **MUST** be set to the Application Address of the Royalty Enforcer if `FreezeAddr != ZeroAddress`, else set to `ZeroAddress`.
+
+> If the creator wants to allow an ASA to be Royalty Free after some conditions are met, it should be set to the Application Address 
+
+The Manager parameter **MUST** be set to the Application Address of the Royalty Enforcer if `ManagerAddr != ZeroAddress`, else set to `ZeroAddress`.
+
+> If the creator wants to update the Freeze parameter, this should be set to the application address
```
smonn, line 85: What happens if a marketplace needs to burn an NFT? If the manager address must be the Application Address, then the Royalty Enforcer needs to be able to burn the NFT as well, or I guess that could be an optional feature.

cusma: Yes this standard does not mandate what you could do as Manager through the Application, this is a free choice of the implementation (e.g. could let Royalty ASA "admins" act like a Manager using the Application as a proxy).

rssalessio:
> > It's not clear to me yet what the standard flow of events is with this mechanism. I think this needs to be described in this document. i.e the parties involved and the sequence of transactions/app calls they make. It would be important to detail this both for peer-to-peer sales, marketplace listings and marketplace auctions.
> 
> Good point, I'll add a swimlane chart. See the example here: https://github.com/algorand-devrel/royalty for a really basic implementation

This is similar to a royalty fee implementation that I made some time ago, if that could be of any help!

https://github.com/rssalessio/algorand-royalty-fees/blob/main/article.md

https://developer.algorand.org/tutorials/royalty-fees-in-algorand-using-pyteal/

ARCs/arc-0018.md
```diff
@@ -0,0 +1,571 @@
...
+
+The `payment_asset` parameter is used in the case that the payment is being made with some asset rather than with Algos. It **MUST** match the Asset ID of the AssetTransfer payment transaction. 
+
+The `curr_offer_amount` parameter is the current amount of the Royalty Asset offered by the `from` account. 
+
+The transfer call **SHOULD** be part of a group with a size of 2 (payment/asset transfer + app call)
+
+See [Security Considerations](#security-considerations) for details on how this method may be gamed.
+
+Upon each transfer the the Royalty Enforcer SHOULD assert that `AuthAddr == AssetSender` for the Seller (`from` parameter) and `AuthAddr == AssetReceiver` for the Buyer (`to` parameter). In other words Seller and Buyer accounts not have been Rekeyed. Exceptions can be handled with `Move` method.
```
barnjamin, line 185:
```suggestion
Upon each transfer the the Royalty Enforcer **SHOULD** assert that `AuthAddr == AssetSender` for the Seller (`from` parameter) and `AuthAddr == AssetReceiver` for the Buyer (`to` parameter). In other words Seller and Buyer accounts not have been Rekeyed. Exceptions can be handled with `Move` method.
```

ARCs/arc-0018.md
```diff
@@ -0,0 +1,571 @@
...
+
+The `royalty_asset` is the ASA id of the asset the policy is set for.  If the `royalty_asset` is 0, the policy is defined for _any_ asset that has its clawback set to the application address. 
+
+The `royalty_receiver` is the address of the account that should receive a partial share of the payment or asset transfer for any [transfer](#transfer) of the `royalty_asset` 
+
+The `royalty_share` is the share of the payment that should be sent to the `royalty_receiver`.  It is specified in basis points (e.g., 1% is 100).  
+
+The `allowed_assets_n` arguments represents the ASA ids that are acceptable for payment. The contract logic **MUST** opt into the set of assets specified in order to accept the payments as part of a transfer.
+
+The Royalty Collector **SHOULD** be opted into the full set of assets contained in this list.
```
barnjamin, line 135:
```suggestion
The Royalty Collector **MUST** be opted into the full set of assets contained in this list. 
```

fergalwalsh:
I have reviewed this proposal with my colleagues from Originals Only, Pera Wallet and some other projects. We definitely agree with the need for a standard for this but we have a few concerns about the proposed approach;

* The use and normalization of Clawback addresses
* The fact that this standard only specs an interface but not an implementation

### Clawback Concerns
Clawback functionality is generally seen as a unusual property for Assets and one that should be a cause for concern and detailed scrutiny. We feel that this is a healthy situation. We are concerned that using clawback functionality for scenarios as simple as enforcing royalties will lead to a 'normalization' of the clawback usage and users will learn to just accept it because so many assets use it. This would be unfortunate because assets with clawback enabled potentially put the user's investments at risk. This risk does not exist on assets without clawback so there is no need for users to understand details of of contracts to know that once they receive an asset it is theirs until they transfer it.

Expecting users to be able to look into the details of every contract that controls clawbacks to determine exactly what they do seems very unrealistic so we should limit their use as much as possible.

If there was a single Well Known contract address for royalty enforcement then this problem could potentially be dealt with by 'whitelisting' the address in wallets/explorers and explaining what exactly it does. However this standard does not propose a single well known contract.

### Specification of an interface only
The standard specifies an interface that an asset should use to be ARC18 compliant but it does not specify the implementation. We are concerned that this could easily lead to scams where assets are offered for sale with what looks like ARC18 restrictions but actually the contracts do something entirely different, potentially allowing unrestricted clawback. 

Additionally, genuine attempts at adhering to this standard could easily include bugs in their implementation that would put users' assets at risk at _all times_. 

Again, we feel it is unreasonable and unrealistic to expect users of these assets to review and verify the correctness of the contracts that control them, especially given that there will be many implementations and deployments of these.

These concerns could be alleviated if the standard could specify a single audited implementation. Ideally it would be a single deployment of an application or possibly it could be a template for an application with an onchain method of verification of that templated application.

### Alternative Solutions
We don't believe that this approach is the only viable approach to royalty enforcement. It seems that it should be possible to enforce royalties without the need to use Clawback but instead by using Freeze. It also seems possible to use local state of a single well known application to store policies rather than many deployments of apps (supposedly) implementing the same interface. I previously briefly described this as a potential solution on the Issue on this subject: https://github.com/algorandfoundation/ARCs/issues/27#issuecomment-946565466.

barnjamin:
> The use and normalization of Clawback addresses

I somewhat agree with this, it should not be used generally given the possible security issues. 

Without using clawback or some App controlled account that can act as escrow the seller must use some intermediate, app controlled, escrow account in order to allow async selling. That is, without some way for the app to move the asset, it'd force the seller to sign another transaction after some purchase is attempted in order to complete the transaction.  

I'm assuming here (maybe incorrectly?) that most folks would prefer a UX that involves setting a price and forgeting about it until they get a notification that it's sold.

>Expecting users to be able to look into the details of every contract that controls clawbacks to determine exactly what they do seems very unrealistic so we should limit their use as much as possible.

Fair point, forcing buyers to delve into contract logic isn't good. Approval programs do have a unique hash that we could/should expose better for this kind of comparison. 

Looking at the ERC-20 standard, they offer a couple "trusted" implementations which may be a reasonable way to provide something to compare a given royalty contract logic against?  https://eips.ethereum.org/EIPS/eip-20#implementation

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+The Asset Offer is the a data structure stored in the owners local state. It is keyed in local storage by the byte string representing the ASA Id. 
+
+```ts
+interface AssetOffer {
+    auth_address:   string // The address of a marketplace or account that may issue a transfer request
+    offered_amount: number // The number of units being offered
+}
+```
+
+This concept is important to this specification because we use the clawback feature to transfer the assets. Without some signal that the current owner is willing to have their assets transferred, it is be possible to transfer the asset without their permission.
```
JasonWeathersby, line 87: it is be possible - > it may be possible

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+    auth_address:   string // The address of a marketplace or account that may issue a transfer request
+    offered_amount: number // The number of units being offered
+}
+```
+
+This concept is important to this specification because we use the clawback feature to transfer the assets. Without some signal that the current owner is willing to have their assets transferred, it is be possible to transfer the asset without their permission. 
+
+In order for a transfer to occur, this field **MUST** be set and the parameters of the transfer request **MUST** match the value set.
+
+> A transfer matching the offer would require the transfer amount <= offered amount and that the transfer is sent by auth_address
```
JasonWeathersby, line 91: Does this transfer automatically clear the offer afterward?

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+A Marketplace may call the transfer request as long as the address initiating the transfer has been set as the `authorized_address` through the [offer](#offer) method in some previous transaction by the current owner.
+
+### ABI Methods
+
+The following is a set of methods that conform to the [ABI](arc-0004.md) specification meant to enable the configuration of a Royalty Policy and perform transfers.
+
+#### Set Policy  (REQUIRED):
+    
+```
+set_policy(royalty_basis: uint64, royalty_recipients: account)
```
JasonWeathersby, line 124: royalty_recipients to royalty_recipient

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+```
+
+Sets the policy for any assets using this application as a Royalty Enforcer.
+
+The `royalty_basis` is the percentage for royalty payment collection, specified in basis points (e.g., 1% is 100).  
+
+A Royalty Basis **MUST** be immutable, if an application call is made that would overwrite an existing value it **MUST** fail.
```
JasonWeathersby, line 131: I thought we were not going to force this and should offer language to put a max in here

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+> The immutability of a royalty basis is important since mutability introduces the possibility for a situation where, after an initial sale, the royalty policy is updated from 1% to 100% for example. This would make any further sales have the full payment amount sent to the royalty recipient and the seller would receive nothing.
+
+The `royalty_receiver` is the address of the account that should receive a partial share of the payment for any [transfer](#transfer) of an asset subject to royalty collection.
+
+The Royalty Collector **MUST** be opted into the full set of assets contained in this list of payment_assets. 
+
+> In the case that an account is not opted into an asset, any transfers where payment is specified for that asset will fail until the account opts into the asset. or the policy is updated.
+
+
+#### Set Payment Asset (REQUIRED):
```
JasonWeathersby, line 142: Why are we making this required? If they want to only accept algos

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+    to: account, 
+    royalty_receiver: account, 
+    royalty_asset_amount: uint64, 
+    payment: transaction, 
+    payment_asset: asset, 
+    current_offer_amount: uint64
+)
+```
+
+Transfers the Asset after checking that the royalty policy is adhered to. This call must be sent by the current owner (the `from` parameter)
```
JasonWeathersby, line 168: This call must be sent by the current owner - I thought the auth address

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+
+The `to` parameter is the account the ASA is transferred to.
+
+The `royalty_receiver` parameter is the account that collects the royalty payment. 
+
+The `royalty_asset_amount` parameter is the number of units of this ASA ID to transfer. The amount **MUST** be less than or equal to the amount [offered](#offer) by the `from` account.
+
+The `payment` parameter is a reference to the transaction that is transferring some asset (ASA or Algos) from the buyer to the Application Address of the Royalty Enforcer. It **MUST** be one of a Payment transaction or Asset Transfer transaction.
+
+The `payment_asset` parameter is used in the case that the payment is being made with some asset rather than with Algos. It **MUST** match the Asset ID of the AssetTransfer payment transaction.
```
JasonWeathersby, line 182: And the enforcer must be opted into the asset.

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+
+The `royalty_asset_amount` parameter is the number of units of this ASA ID to transfer. The amount **MUST** be less than or equal to the amount [offered](#offer) by the `from` account.
+
+The `payment` parameter is a reference to the transaction that is transferring some asset (ASA or Algos) from the buyer to the Application Address of the Royalty Enforcer. It **MUST** be one of a Payment transaction or Asset Transfer transaction.
+
+The `payment_asset` parameter is used in the case that the payment is being made with some asset rather than with Algos. It **MUST** match the Asset ID of the AssetTransfer payment transaction. 
+
+The `current_offer_amount` parameter is the current amount of the Royalty Asset [offered](#offer) by the `from` account. 
+
+The transfer call **SHOULD** be part of a group with a size of 2 (payment/asset transfer + app call)
```
JasonWeathersby, line 186: Are we specifying a specific order

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+
+> This requirement is meant to prevent a sort of race condition where the `authorized_address` has a `transfer` accepted before the `offer`-ing account sees the update. In that case the offering account might try to offer more than they would otherwise want to. An example is offered in [security considerations](#security-considerations)
+
+To rescind an offer, this method is called with 0 as the new offered amount.
+
+If a [transfer](#transfer) or [royalty_free_move](#royalty-free-move) is called successfully, the `offer` **SHOULD** be updated or deleted from local state. 
+
+> Exactly how to update the offer is left to the implementer. In the case of a partially filled offer, the amount may be updated to reflect some new amount that represents `offer_amount - amount transferred` or the offer may be deleted completely.
+
+#### Get Offer (OPTIONAL):
```
JasonWeathersby, line 236: I would suggest we remove this

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+
+    par Delist
+        Bob->>Royalty Enforcer: offer 0 
+        Bob->>Marketplace: delist
+    end
+```
+
+### Metadata
+
+The metadata associated to an asset **MAY** conform to any ARC that supports an additional field in the `properties` section specifying the specific information relevant for off-chain applications like wallets or Marketplace dApps.
```
JasonWeathersby, line 513: This should be immutable imo

ARCs/arc-0018.md
```diff
@@ -0,0 +1,601 @@
...
+
+The fields that should be specified are:
+
+  - `application_id` - The ID of the Royalty Enforcer application
+  - `application_address` - The Address of the Royalty Enforcer application
```
JasonWeathersby, line 518: Why do we have these

JasonWeathersby: i thought it was just application_id

ARCs/arc-0018.md
barnjamin: its specified as part of the abi method signature

ARCs/arc-0018.md
barnjamin: Just convenience for off chain things to make parsing easier. Otherwise its a couple api calls and parsing out encoded bytestrings.

I can remove but they seem useful?

ARCs/arc-0018.md
barnjamin: Only required to be implemented, not required to be used

cusma:
Hi @fergalwalsh I'll try to give my point of view about the point you raised.

> We are concerned that using clawback functionality for scenarios as simple as enforcing royalties will lead to a 'normalization' of the clawback usage and users will learn to just accept it because so many assets use it.

I think the same argument could be applied for the Freeze Address as well, so I do not consider this as a key differential between a Clawback-based approach vs a Freeze-based approach.

> Expecting users to be able to look into the details of every contract that controls clawbacks to determine exactly what they do seems very unrealistic so we should limit their use as much as possible.

Although I agree with the general principle we should recognize that the same applies whenever you interact with any Smart Contract. This is true for DeFi applications, DAOs and dApps in general: average users are not able to "audit" each contract by themselves. I think this particular use of Smart Contract is not different from the rest.

> If there was a single Well Known contract address for royalty enforcement then this problem could potentially be dealt with by 'whitelisting' the address in wallets/explorers and explaining what exactly it does. However this standard does not propose a single well known contract.

A standard reference is a necessary step before a best reference implementation, since a standard gives to developers a common ground to build and find out the best possible implementation that fits the standard.

> We don't believe that this approach is the only viable approach to royalty enforcement. It seems that it should be possible to enforce royalties without the need to use Clawback but instead by using Freeze.

I think that the usage of Freeze does not solve your concerns since, in my view, trusting a Clawback Address or trusting a Freeze Address is not that different. Owning an ASA that could be _maliciously or mistakenly_ frozen (maybe forever) is not much better than an ASA that could _maliciously or mistakenly_ clawback.

Moreover: if I understand your proposal correctly you suggest (please correct me if I'm wrong):

- Issue the Royalty ASA as DefaultFrozen
- Assign the Freeze Address to a Royalty Enforcer Smart Contract
- Issue a Group Transaction with the following transactions:
1. Application Call to the Royalty Enforcer -> InnerTxns: 1. AssetUnfreeze the Sender, 2. AssetUnfreeze the Receiver
2. AssetTransfer from Sender to Receiver
3. Payment from Sender to Royalty Collector
4. Application Call to the Royalty Enforcer - InnerTxns: 1. AssetFreeze the Sender, 2. AssetFreeze the Receiver

**Conclusion**
Seems to me that the usage of Clawback (that overrides the Freeze status) just makes the implementation much more simpler without really introducing more trust than the usage of Freeze, notice that the average user will not be able to audit the "clawback version" nor the "freeze version".

rssalessio:
> > We don't believe that this approach is the only viable approach to royalty enforcement. It seems that it should be possible to enforce royalties without the need to use Clawback but instead by using Freeze.
> 
> I think that the usage of Freeze does not solve your concerns since, in my view, trusting a Clawback Address or trusting a Freeze Address is not that different. Owning an ASA that could be _maliciously or mistakenly_ frozen (maybe forever) is not much better than an ASA that could _maliciously or mistakenly_ clawback.

Your point of view is understandable, but this is rather subjective. Correct me if I'm wrong, but the frozen feature is objectively better in terms of ownership, since ownership is retained (even if the asset may remain frozen forever...but at least you are still the owner). Even if there are bugs in the smart contract, one does not have to worry about losing ownership.

The Royalty Fee mechanism is quite important, so I would consider spending more time on it regardless of the implementation difficulty.

rssalessio:
> > Although I agree with the general principle we should recognize that the same applies whenever you interact with any Smart Contract. This is true for DeFi applications, DAOs and dApps in general: average users are not able to "audit" each contract by themselves. I think this particular use of Smart Contract is not different from the rest.
> 
> point_up This is spot on. No matter the service you're using, whether it be a DEX, a bridge or an NFT marketplace, there is still a level of trust between the user/customer and the service itself.

Indeed there is  a level of trust in using DeFI services. However, personally, if I don't like a DeFI app I always have the possibility to switch to another one.

If I don't like a blockchain because of the choices the devs take, people may consider switching blockchain (since there are still so many blockchains out there). If I want to release a new service that uses royalty fees, but I know that royalties are not well seen on Algorand, will I still choose Algorand? (rethorical question).


> Not to mention that this spec is more easily achieved using clawbacks vs the freeze address, and in the eyes of a purist they might as well be the same (undesired) thing. I don't think it's worth getting into an argument of transparency vs viability.

In my view it's not not only a "transparency vs viability" argument. Lack of transparency is just one of the possible risks, but we also need to consider other risks (such as bugs, loss of ownership, etc...).

ARCs/arc-0018.md
```diff
@@ -0,0 +1,696 @@
...
+> The `current_*` values for the offer are passed to prevent the race or attack described in [Security Considerations](#security-considerations).
+
+
+### Read Only Methods
+
+Three methods are specified here as `read-only` as defined in ARC22.
+
+#### Get Policy
+```
+get_policy()(address,uint64)
```
smonn, line 305: Should this be named to mirror the setter?

```suggestion
get_royalty_policy()(address,uint64)
```

Alternatively, renamed the setter to (the shorter) `set_policy`.

smonn:
Leaving a couple of thoughts here:

First, and apologies if I missed this in previous comments, what's the reason `current_authorized_address` is not required to be specified for the `transfer` method but is required for `royalty_free_move`? Shouldn't both need to verify the transfer is done based on the current offered state?

Second, to hopefully simplify things a bit, may I suggest normalizing the argument order and their names? For example, one could do this:

```
offer(
  royalty_asset: asset,
  royalty_asset_amount: uint64,
  auth_addr: account,
  offered_amount: uint64,
  offered_auth_addr: account,
)

transfer(
  royalty_asset: asset,
  royalty_asset_amount: uint64,
  from: account,
  to: account,
  offered_amount: uint64,
  offered_auth_addr: account,
  royalty_receiver: account,
  payment_txn: pay | axfer,
  payment_asset: asset,
)

royalty_free_move(
  royalty_asset: asset,
  royalty_asset_amount: uint64,
  from: account,
  to: account,
  offered_amount: uint64,
  offered_auth_addr: account,
)
```

barnjamin:
> First, and apologies if I missed this in previous comments, what's the reason current_authorized_address is not required to be specified for the transfer method but is required for royalty_free_move? Shouldn't both need to verify the transfer is done based on the current offered state?

For the `transfer` it should not be required since the sender of the transaction should match the `current_auth_addr`, for the `royalty_free_move` I think it should be possible to do the same so I'll remove it from that one.

>Second, to hopefully simplify things a bit, may I suggest normalizing the argument order and their names? For example, one could do this:

Good point, I'll take a look at doing this today

AustP:
Hi, I'm Aust from https://algoseas.io (building an NFT Marketplace) and wanted to share our team's thoughts on this.

Before ARC18 was proposed, we actually built a royalty enforcement system that was very similar to ARC18. ASAs were default frozen and a clawback mechanism from a smart contract was used to move the assets around. We also had a mechanism that mimicked the `royalty_free_move` endpoint of ARC18 to allow the movement of ASAs to escrow accounts or transferring amongst an individual's different wallets, etc.

After we built the system out and got it working with our marketplace, we ultimately ended up abandoning the system. The biggest problem is that by having a `royalty_free_move` method, it undermines the enforcement aspect of the royalty enforcement system. A marketplace could theoretically be built to exclusively use this endpoint for any "Royalty Enforced ASA" (REASA) and bypass the entire system.

It seemed to us that by having a system like this, we just enforce hardships on users (inability to transfer REASAs like standard ASAs, REASA incompatibility with existing marketplaces, etc.) without actually enforcing royalties 100% of the time. At the end of the day, a system like this boils down to: we rely on other developers to make sure royalty enforcement is happening.

Rather than fight that idea, we embraced it. We can relax the constraints (and complexity) of a royalty enforcement system if we are okay with the idea that it won't be enforced 100% of the time. REASAs can look and act exactly like ASAs to an end-user but if the user happens to sell a REASA on a marketplace that supports royalty enforcement, royalty payouts will automatically take place behind the scenes. We also get the benefit of having a royalty enforcement system that is backwards compatible with existing ASAs.

The royalty enforcement system that we are currently using can be viewed here: [AlgoSeas Royalty Enforcement System](https://docs.google.com/document/d/1x-cyLNMOm-E_ovnNdYs-B7UvRsA-bRMNqfQi_43IHXE/edit?usp=sharing). (This document was not created for this PR so pardon the shift in tone.) To summarize the document, we allow ASA creators to encode what the royalty information for the ASA should be as part of the ASA's metadata and we enforce it on-chain if it exists.

Maybe there's a better solution to the problem of enforcing royalties 100% of the time while supporting `royalty_free_move`, but this is where we landed... happy to discuss and help make Algorand the best that it can be!

barnjamin:
Hi @AustP, thanks for sharing the design!

>The biggest problem is that by having a royalty_free_move method, it undermines the enforcement aspect of the royalty enforcement system.

This spec defines this method as optional but if implemented it requires that it be performed _only_ by the administrator of the contract. That administrator I _expect_ should have some vested interest in the royalty being enforced (or not?) 

That said, there are still many ways to circumvent the royalty payment like using rekey to do a sort of atomic swap or making a small payment that gets split for royalties and a follow up payment with the real payment.  I think (😬) if the contract creator/owner wants they should be able to apply checks for rekey or limit the number of third party marketplaces that may be allowed to sell to only those that are known to have adherent contracts.

I think it'd be up to the implementer/royalty receiver to decide how stringent they'd want to be with these checks, understanding that the more strict they are, the less interested folks might be in the asset.

Is this spec something that algoseas would support if adopted?

AustP:
> This spec defines this method as optional but if implemented it requires that it be performed only by the administrator of the contract. That administrator I expect should have some vested interest in the royalty being enforced (or not?)

Ah yes I missed that part of the spec. I do think there will be use cases where having a locked down asset that enforces royalty payments will be needed. However, I think that for most NFTs, this level of enforcement is unnecessary.

> Is this spec something that algoseas would support if adopted?

Yes, AlgoSeas will support the standards established by the community.

smonn:
> No preference, I'll adapt ARC-20 to what you choose here (being camelCase if we don't change anything).

Sorry, my bad. Looks like ARC-18's ABI uses `snake_case`. @aldur / @barnjamin

Vytek:
When the final release?

barnjamin:
@aldur @cusma any issues with this going to draft and merging?

Vytek:
> @aldur @cusma any issues with this going to draft and merging?

Is it possible to consider a fourth figure that gets a small percentage on the sales like the platform?

PR #74, opened 2022-03-02
Extend arc-6 with enableNetwork and enableAccounts

DanBurton:
At Reach we have found it useful to split `enable` into `enableNetwork` and `enableAccounts` so that network queries can be performed without asking the user to select an account. We propose that this be included as optional in arcs 6, 10, and 11.

### DIFF
--- a/ARCs/arc-0006.md
+++ b/ARCs/arc-0006.md
@@ -11,12 +11,13 @@ status: Draft
 
 ## Summary
 
-An API for a function used to discover the addresses a wallet user is willing to use for a given Algorand DApp.
+An API for a function used to discover the addresses a wallet user is willing to use for a given Algorand DApp, and the network on which the user is willing to use them.
 
 ## Abstract
 
-A function, `enable`, which allows the discovery of accounts. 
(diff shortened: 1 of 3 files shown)


robbiemu:
are arc's mutable? I thought the pattern was to submit an informational arc that acts as an addendum to a previous one, in such a case.

SudoWeezy:
If this is still relevant, please consider merging the version in the repo before submitting it.

[ARC-6](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0006.md)

SudoWeezy:
@DanBurton, could you please tell me if this is still relevant?

SudoWeezy:
Should be added to [this PR](https://github.com/algorandfoundation/ARCs/pull/97) to create a new ARC extension for [ARC-6](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0006.md)

SudoWeezy:
PR merged with https://github.com/algorandfoundation/ARCs/pull/97

PR #75, opened 2022-03-07
Create arc-0024.md

FrankSzendzielarz:
ARC 20 - proposal to extend Arc4 to a general Application Specification discoverable file format

### DIFF
--- a/ARCs/arc-0020.md
+++ b/ARCs/arc-0020.md
@@ -1,0 +1,485 @@
+---
+arc: 20
+title: Algorand Application Metadata 
+status: Draft
+---
+
+# Algorand Application Metadata
+
+> This API is a draft.  Some elements may change.
+
+## Summary
+
+A standard for a general metadata format for Algorand
+_Applications_ (aka "smart contracts"), the _Application Specification_,
+supporting integration and discovery scenarios.
+
+## Specification
+
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**",
+"**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**",
+"**MAY**", and "**OPTIONAL**" in this document are to be interpreted
+as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+> Comments like this are non-normative.
+
+## Definitions
+
+* **Application:** an Algorand Application, aka "smart contract",
+  "stateful contract", "contract", or "app".
+* **Application Specification:** a metadata file describing the Application in both
+  human and machine readable ways.
+* **HLL:** a higher level language that compiles to TEAL bytecode.
+* **dapp (frontend)**: a decentralized application frontend, interpreted here to
+  mean an off-chain frontend (a webapp, native app, etc.) that interacts with
+  Applications on the blockchain.
+* **explorer**: an off-chain application that allows browsing the blockchain,
+  showing details of transactions.
+* **ABI**: a specification of the methods an Application offers.
+
+## Note
+
+The approach in this ARC document is to list goals or use cases, identify
+information and format for each goal, then collate them into a single format.
+
+The collated proposal is in the [Application Specification](#application-specification) section
+
+As the discussion on this ARC continues and new use cases are found, it would
+be helpful to list them here and extend the format in the same way.
+
+## Overview
+
+This document expands on [ARC-4](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md) to
+offer general Application metadata, supporting further use cases, and to offer
+ways of linking Applications to their metadata files to support their
+safe discovery.
+
+The goals of [ARC-4](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md) include support
+for HLLs and for dapp authors to integrate with Applications. IDEs offer
+the ability to generate proxy classes for different HLLs
+based on metadata about the remote API to interface with. Examples include
+SOAP WSDL, OpenAPI, and in future Algorand Application Specification metadata
+will be used in similar ways. Absent IDEs or similar code generation
+tools, dapp authors still also need a human readable document of the
+Application in order to integrate.
+
+An Application offers two types of interaction: application calls
+and read-only named state variables (global and local). ARC-4 provides
+information about the former but not the latter. It seems likely that
+there will also be other types of metadata needed by Application consumers.
+At the time of writing, for example,there are discussions around Application-
+based NFTs and their metadata.
+
+Whatever the content of those metadata files, consumers in the first
+place need some way of discovering them. They also need to know that
+the versions of both Application and Application Spec match.
+
+## Goals
+
+This section lists the goals and needs the Application Specification intends 
+to achieve and fulfill.
+
+1. Support tools and IDEs to automatically generate Application proxy classes, including ABI methods, base 'raw' methods and state variables.
+2. The Application Specification format must be human readable and machine readable.
+3. The Application Specification format must be extensible, to allow for inclusion of new ideas,
+such as discovery tags, NFT specifications, and so on.
+4. Support discovery by establishing a link between Applications and their corresponding
+Application Specification files.
+5. Composability: Applications may inherit behaviour from each other in different ways, such as
+implementing interfaces, composing and wrapping other application, direct single inheritance and so on.
+
+The below sections address the above goals, each adding some discussion
+and relevant specification details.
+
+### **1. IDE Support**
+
+There are two general ways of interacting with an Application. One is by making an
+application call transaction and the other is by reading global or local state.
+The state variables are named.
+
+An imaginary Application called SugarSupplierContract offers
+a price for refined sugar to the public. This can be read from the
+global state variable "SugarPrice." A retailer negotiates a discount
+wholesale price. This is offered to the customer via a local state variable
+called "YourPrice." The Application also offers a method to calculate 
+the estimated shipping price for a given volume of sugar.
+
+The retailer would like to be integrate with this SugarSupplierContract
+using the development tool of their choice.
+
+The retailer, in this scenario already has the Application Specification
+available, so they give this to their IDE and the following proxy class is
+generated. The below is just a C# example.
+
+```cs
+   public abstract class SugarSupplierContract : SmartContractReference
+    {
+        [Storage(StorageType.Global)]
+        public ulong SugarPrice;
+
+        [Storage(StorageType.Local)]
+        public ulong YourPrice;
+
+  
+        public abstract ulong ShippingCost(ulong VolumeInKilos);
+    } 
+```
+
+Proxies can support server or dapp (frontend) calls to an Algorand node
+or even contract to contract invocations for compilers.
+
+Even without the IDE to tooling support, a dapp author
+still needs to understand how the SugarSupplierContract is intended
+to be used, so information must be provided somehow.
+
+To achieve this goal, this ARC document proposes that the
+Application Specification include a single ARC-4 Contract
+description, and that the ARC-4 Contract section be updated to
+include locals and globals:
+
+```typescript
+interface Contract {
+  /** A user-friendly name for the contract */
+  name: string;
+  /** Optional, user-friendly description for the interface */
+  desc?: string;
+  /**
+   * Optional object listing the contract instances across different networks
+   */
+  networks?: {
+    /**
+     * The key is the base64 genesis hash of the network, and the value contains
+     * information about the deployed contract in the network indicated by the
+     * key
+     */
+    [network: string]: {
+      /** The app ID of the deployed contract in this network */
+      appID: number;
+    }
+  }
+  /** All of the methods that the contract implements */
+  methods: Method[];
+
+  /** All local variables intended for consumption as part of this contract **/
+  locals: Variable[];
+
+  /** All global variables intended for consumption as part of this contract **/
+  globals: Variable[];
+
+}
+```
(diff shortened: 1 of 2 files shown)


runvnc:
I appreciate that you created this. Thank you.

My suggestions:

- Consider "modules" as a use-case also, for example a `random()` function intended to be integrated into a more complex application
- New versions will reference different contract JSON/asset IDs.
- Leverage the ASA infrastructure to identify application specifications: `#pragma appspec [assetid]`
- Require a note prefix for the ASA creation/update or app creation/update. This means that tools/sites can index the app spec data by querying an Indexer against that note prefix rather than having to look at every single application or relying on some central website where everyone is required to submit. 

The question has been brought up repeatedly "how much of this discovery stuff is in scope for Algorand rather than the community" -- standardizing on a note prefix or any other decentralized mechanism for finding this data will enable multiple tools and curation options. **Not doing so will default to whatever solution pops up and becomes popular becoming the de facto centralized monopoly solution, likely controlled by a small company**. Based on the comments so far, it seems that this is what the community prefers. This seems very short-sighted and really leaves up the outcome up to chance.
- One core requirement for interfacing with an application may be to implement a particular interface as the other half of a producer/consumer protocol.
- App spec ideas, stored in IPFS referenced by the ASA in the pragma:
```json
{
  "appid": 123434,
  "description":"",
  "website": "",
  "contact":"",
  "version": "",
  "keywords": ['voting','DAO'],
  "implements-arc4": [ 234536, 345665 ] // list of ASAs pointing to interface JSON
  "requires-arc4": [467467] // list of ASAs pointing to interface JSON that the _consumer_ must implement
  "globals": [
     { "rand": "uint64 with the next random number (0-2,147,483,647)"}
  ]
}

```

FrankSzendzielarz:
Is the modules idea the same as the Interfaces idea I mentioned in the Composability section? Arc-4 talks about Interfaces as distinct from the Contract concept, but with Interfaces I am not sure how the relationship between Application Spec is maintained, especially when Interface creation occurs after Application Spec creation and by 3rd parties. (eg: 10 Applications all do similar things, and Development Team 11 extrapolates an interface post hoc). Regardless, the Interfaces mention is in the Composability section, so if your modules idea is different, please elaborate and I will add it in as a talking point.

I know the ARC process is supposed to separate the discussion from the ARC itself but in this early stage, to stimulate creativity and give new readers the benefit of the discussions we had in Discord, I will add new ideas directly into the ARC as TBD.

I will go through the rest of your suggestions and make updates later, as I am focusing on the compiler tasks today.

runvnc:
I mean, as you mention in 'composability', the case where the contract referenced by an application specification is actually not generally used as a standalone application, but rather integrated as part of another more comprehensive application. Yes, it would implement a specific interface in that case. For example, generating a random number is not useful for the end user, but very useful when used as part of a more elaborate application, which might also take advantage of other contracts that implement bits of other common functionality.

Obviously if someone creates an interface after an application, the application and app spec would need to be updated if it was desired to reference that.

barnjamin:
This is interesting and seems well thought out. 

From a quick read through the only thing I'll mention is that if/when we get another storage type for extensible storage (hopefully soon?) we'd want a third option besides global/local storage.

FrankSzendzielarz:
Good shout. I'll get to updating the Arc in due course. Hopefully today. In fact, come to think of it, there might be cases where scratch vars should be known too.

barnjamin:
@fabrice102 is this good to merge as draft?

FrankSzendzielarz:
to all above: i will definitely update all your suggestions into the ARC. My first compiler milestone due tomorrow, so focusing on that. thanks!

barnjamin:
How about for programmatically determined keys? for example using the `Itob(asa_id)` as the key like:
```json
{
  "locals": [ 
        {
            "name": "{uint64 asset id as big endian bytes}",
            "desc": "The ID of the asset that has been offered as key, authorizing address as value",
            "type": "address"
        }  
    ],
    "globals": [ 
        {
            "name": "0000000000000000",
            "desc": "The default royalty policy for any ASA unless its ASA ID is specified",
            "type": "(address,uint64,uint64,uint64,uint64)"
        }  
        {
            "name": "{uint64 asset id as big endian bytes}",
            "desc": "Royalty policy for asset ID in key ",
            "type": "(address,uint64,uint64,uint64,uint64)"
        }  
    ]
    }
    ```

**Comment by FrankSzendzielarz** [comment]
off the top of my head, and without deeper consideration/understanding, it looks like something I'd consider as trying to abuse the key as value. i think if the consumer of the spec file cannot guarantee that the key is there, it might as well not be, but like i say this is just a quick off the cuff comment. i am thinking of the contract is literally being a class with 2 public method (approval and clear state) and the local/global state as properties or fields with fixed names, a-la OO. yes algorand contracts allow for field names to be dynamic , keys to be deleted, but for contract interoperability that might be something out of scope?

File: ARCs/arc-0020.md [hunk 1]
```diff
@@ -0,0 +1,485 @@
...
+}
+```
+
+There needs to be a _Variable_ specification too, following similar style as
+the _Method_ entity:
+
+```typescript
+interface Variable {
+  /** The name of the variable **/
+  name: string;
```
fabrice102, line 178: Sorry I haven't looked at the proposal in details.
But one question is how do you encode `name`, especially for non-utf-8 names.
string is indeed limited to utf-8, but not keys.

FrankSzendzielarz: Hi there. Thanks. This Arc is about a metadata format that allows a contract to be expressed as a 'class' where methods (abi signatures) and fields (state vars) are named as utf8 . in both cases the underlying byte array can hold anything but having method and property names as anything other than utf8 strings is unlikely to be directly supported by HLLs. Yes the code generators implementing the proxy classes on the basis of these metadata files to call the contracts could add mapping code to write or read from byte array keys but this would, it seems to me, be superfluous complexity. Though am happy to be told why here non utf8 property names should be supported.

aldur:
@FrankSzendzielarz, following up from your chat with @cusma, could you swap ARC-20 for ARC-24 references?

FrankSzendzielarz:
> @FrankSzendzielarz, following up from your chat with @cusma, could you swap ARC-20 for ARC-24 references?

Nil problemo. Will do, though it'll be tomorrow if there's no urgency.

cusma:
Adding https://github.com/algorandfoundation/ARCs/issues/118#issuecomment-1210521958 here as well for visibility.

ARCs/arc-0024.md
```diff
@@ -0,0 +1,485 @@
...
+metadata emerge, users may be exposed to phishing-type attacks where
+they are misdirected to fake application clones (eg: T1nyman, spelt
+with a 'one').
+
+At a minimum then some kind of veriable link between Author, Application and
+Specification needs to exist.
+
+This ARC proposes the following:
+
+1. Store the SHA-512/256 hash of the current Application Specification file
```
SudoWeezy, line 305: We might need to synchronize with PR for [ARC-23](https://github.com/algorandfoundation/ARCs/pull/80).

ARCs/arc-0020.md
jannotti: I suspect a common use case would be using a key that is the raw 32 bytes of an address, or is a hash of something.

Looking ahead to boxes, I think those will be very common names for boxes.

ARCs/arc-0024.md
jannotti: Yes, please.  This is evidence for having a general approach in ARC23 - one that allows specifying several pieces of metadata.  I think the `bytecblock` approach would would well, perhaps with a first key that specifies "ARC24", and then pairs of bytes that describe what is being annotated.

`bytecblock "ARC24" "source" "4HSF68GH" "ARC30spec" "63GF53KU" ...`

ARCs/arc-0024.md
```diff
@@ -0,0 +1,485 @@
...
+in the Application itself.
+2. To save storage, declare a constant byte array, using _bytecblock_
+that is of the format "APPSPEC{HASH}" where HASH is the 32 byte
+hash of the file.
+3. The first opcode in the ApprovalProgram program is where the
+_bytecblock_ MUST be placed.
```
jannotti, line 311: There are some downsides to forcing this `bytecblock` to appear as the first instruction.  It means that a user of TEAL can't place that instruction directly in their TEAL, because of interactions with the `byte` pseudo-op. The ARC23 spec detail this in more detail, but at the very least, I would expect you might want to start with:
```
b skip
bytecblock "THE STUFF YOU'RE TALKING ABOUT"
skip:
```

I'd be in favor of this being a standard that describes app state _only_ and leave ARC23 the job of binding that schema to the app.

ARCs/arc-0024.md
```diff
@@ -0,0 +1,485 @@
...
+  version: uint32;
+  
+
+}
+```
+
+### **5. Composability**
+
+It may be desirable to build Contract elements through inheriting from another Contract
+or implementing Interfaces.
```
jannotti, line 340: I would be in favor of "flattening" any such composition mechanism. So, if an app contains some state variables because some tool or HLL has created it as "inheriting" from another app or class, that need not be described here. Instead, document the field directly. The app consumer does not need to know _why_ there is a state variable, it juts needs to know that it's there, and the format of the dat in it.

ARCs/arc-0020.md
jannotti: It will also be very reasonable for a HLL to compile a variable down to a single letter key, in order to make a full 127bytes available in the value. So perhaps there should be a `name` and a `key`.  The name is human readable and might be the variable name in the HLL, even if the Algorand state key is called something smaller and less understandable.  Perhaps `key` is defaulted to `name` to simplify a common case, but `key`, if specified, is specified as b64 encoded bytes.

ARCs/arc-0024.md
```diff
@@ -0,0 +1,485 @@
...
+to be used, so information must be provided somehow.
+
+To achieve this goal, this ARC document proposes that the
+Application Specification include a single ARC-4 Contract
+description, and that the ARC-4 Contract section be updated to
+include locals and globals:
+
+```typescript
+interface Contract {
+  /** A user-friendly name for the contract */
```
jannotti, line 142: This section seems to show its heritage from when it was being proposed as a change to ARC4.  I'd rather see this presented without all of the details around `Contract`, and instead something like, "This ARC proposes two optional additional keys to the Contract" schema of ARC4: `globals` and `locals`.  Or, perhaps a cleaner why to plan for the future is to only propose one new key, `ARC24`, which would itself have subkeys `global` and `local`.

I think it's reasonable (though perhaps not explicitly stated in ARC4) to expect that adding new keys to an existing format is fine. So this would be a compatible addition to the Contract Schema, and ARC24 compliant explorers/etc could use it.

ARCs/arc-0024.md
```diff
@@ -0,0 +1,485 @@
...
+
+Extensions to this specification MUST NOT ever include machine-only or binary formats.
+
+### **3. Extensibility**
+
+It is difficult to know what the Application Specification will be expected
+to contain in the future. One thing that is reasonably certain though
+is that everyone would benefit if the information about the Application
+was not scattered across different places in different formats.
+
```
jannotti, line 237: I think this can be removed if, instead, we extend `Contract` in the same way, in future ARCs.  A future ARC-88 would add another optional key "ARC88" to the `Contract` schema. Old code would ignore it, new code would use it.

But we avoid the complexity of this new envelope which is currently empty.  JSON is a pretty goo envelope for unknown future data already.

ARCs/arc-0024.md
```diff
@@ -0,0 +1,485 @@
...
+}
+```
+
+using:
+
+```typescript
+interface Application {
+
+  /** The corresponding Application id  **/
+  appID: uint64;
```
jannotti, line 379: I think this is handled by the `networks` field of the ARC4 `Contract` object.

ARCs/arc-0024.md
```diff
@@ -0,0 +1,485 @@
...
+MUST include the following extension:
+
+```typescript
+interface Application {
+
+  /** The corresponding Application id  **/
+  appID: uint64;  
+
+  /** The Application version  **/
+  version: uint32;
```
cusma, line 331: What about a `string` for the version so that you can adopt also a `major.minor.fix` format?

ARCs/arc-0024.md
```diff
@@ -0,0 +1,485 @@
...
+
+```typescript
+interface Variable {
+  /** The name of the variable **/
+  name: string;
+  /** Optional, user-friendly description for the variable **/
+  desc?: string;
+  /** The type of the variable **/
+  type: string;
+}
```
cusma, line 183: Are we going to allow ABI types for `Variable`? E.g.: I can specify `bool` so that a client can properly read the App state interpreting a `uint64` as a `bool`.

FrankSzendzielarz: Yes this ARC is quite old now. Since it was written there are a few things I've learned. Yes probably the variable type should be from an agreed list, in ARC4. Though right now arc4's types are far from acceptable.

FrankSzendzielarz: Also the Globals/Locals needs to change as Jannotti suggests above, that global/local not be explicit - as we will have boxes, etc.

cusma: Yes, the _current_ state schema can not be known on App creation due to Boxes, but I still think would be useful to have a standard way to declare the state schema needed on App create in addition to methods' ABI specs. For example: imagine a client wants to "install" a TEAL dependency, deploy the App and call it. ABI Interface JSON is useful to encode the calls, but the client should still know the initial state schema to deploy the App correctly.

PR #77, opened 2022-03-09
ARC-4: Remove constraints on reference types

StylishTriangles:
With this change it will be possible to create flexible ABI-compliant interfaces without resolving to dirty hacks.

Example use-cases in which this change is handy:
- `optin(asset[])void`
- `distributerewards((account,uint64)[])void` - or any batch operation

In my opinion, such methods are useful and are likely to be used by application creators. Their current incompatibility with the ABI will result in non-standardized approaches being taken.

### DIFF
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -520,7 +520,7 @@ any leading zeros, _unless_ `N` is zero, in which case only a single 0 character
 * `<type>[]`: A variable-length array. `type` can be any other type.
 * `string`: A variable-length byte array (`byte[]`) assumed to contain UTF-8 encoded content.
 * `(T1,T2,…,TN)`: A tuple of the types `T1`, `T2`, …, `TN`, `N >= 0`.
-* reference types `account`, `asset`, `application`: only for arguments, in which case they are an
+* reference types `account`, `asset`, `application`: For encoding purposes they are an
 alias for `uint8`. See section "Reference Types" below.
 
 Additional special use types are defined in [Reference Types](#reference-types)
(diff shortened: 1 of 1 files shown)


barnjamin:
I like this, and I don't immediately see why they wouldn't be allowed to begin with. @jannotti @jasonpaulos ?

jannotti:
I think this is ok, though I too would like Jason's input.  Our reasoning for the original limitation is that it is probably(?) a bad idea to use `[]asset` as an argument type, as it implies that you could send 10 assets.  But that objection doesn't apply to, say, `[4]asset`, or using an asset type in a tuple.  Further, it's not as though the ABI limits your ability to say `[800]uint64` even though that can't be fit in a txn either.

 If Jason can't think of any further problems, I would be in favor of accepting this change.

jasonpaulos:
I think this change makes sense.

The only thing I don't like about it is how it interacts with inner transaction resource creation in TEAL 6+. For example, if my app creates 100 assets using inner asset create txns, then I wanted to call a function `registerAssets(asset[])void` on another app with these 100 new assets, the maximum I would be able to pass in is actually 8, since we're limited by the size of the transaction's foreign assets array. But since the assets were created in the same transaction group, it's technically possible for the `registerAssets` inner app call to reference all of them, we just can't encode them all.

I guess this is really a weird thing about the AVM, it just gets exposed more with this ABI change. (Perhaps raising the foreign resource limit to 256 for inner txns would solve it?)

Anyway, I still think this is a good idea despite this one limitation, though if accepted it will certainly take some time before ABI implementations in the SDK and goal can work with it.

jannotti:
That's a nice solution to that problem. I only noticed the issue yesterday
as well. But I think your solution works - it's ok to raise the size limits
on inner foreign arrays because the caller can only add things it has
available anyway.  We have to be careful about allocbound though.  This
would produce txns in ApplyData with big foreign arrays, so those
allocbounds must be big enough for them. (None of this should prevent this
merge, it's just a note to keep in mind if we expand inner foreign arrays.)

ARCs/arc-0004.md
```diff
@@ -581,18 +581,13 @@ For any ABI value `x`, we recursively define `enc(x)` to be as follows:
...
+- `byte` is an alias for `uint8`
+- each of the reference types is an alias for `uint8`
 
 ### Reference Types
 
-Three special types are supported _only_ as the type of an
-argument. They **MUST NOT** be embedded in arrays or tuples, as 
-their encoding is unusual. (They **MAY** implicitly be part of
-a tuple if they are the type of the 15-th argument or an argument
-further away.)
```
jasonpaulos, line 594: Could you still mention that reference types are valid only in arguments (not in return values), and could you explicitly state that embedding these types in tuples or arrays is allowed?

StylishTriangles: I misinterpreted the part that said that it can only be a type of the argument, meaning it cannot be a return type. Will fix.

jannotti: I'm on board after noting that somewhere that they cannot be return types.

StylishTriangles:
Added requested changes for the reference assets

jasonpaulos (approved):
Thanks for making the changes, looks good to me

fabrice102 (approved):

PR #79, opened 2022-03-16
Create arc 22

ori-shem-tov:
Extend arc-4 to add `read-only` annotation to methods

### DIFF
--- a/ARCs/arc-0022.md
+++ b/ARCs/arc-0022.md
@@ -1,0 +1,65 @@
+---
+arc: 22
+title: Extend arc-4 to add `read-only` annotation to methods
+status: Draft
+---
+
+# Extend arc-4 to add `read-only` annotation to methods
+
+## Summary
+
+The following document introduces a convention for creating methods (as described in [ARC-4](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md)) which don't mutate state.
+
+## Abstract
+
+The goal of this convention is to allow smart-contract developers to distinguish between methods which mutate state and methods which don't by introducing a new property to the `Method` descriptor.
+
+## Specification
+
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+> Comments like this are non-normative.
+
+## Read-only functions
+
+A `read-only` function is a function with no side-effects. In particular, a `read-only` function should not include:
+- local/global state modifications
+- calls to non `read-only` functions
+- inner-transactions
+- sending logs using the `Log` opcode
+
+It is recommended for a `read-only` function to not access transactions in a group or metadata of the group.
+
+> The goal is to allow algod to easily execute `read-only` functions without broadcasting a transaction
+
+In order to support this annotation, the following `Method` descriptor is suggested:
+```typescript
+interface Method {
+  /** The name of the method */
+  name: string;
+  /** Optional, user-friendly description for the method */
+  desc?: string;
+  /** Optional, is it a read-only method (accroding to ARC-22) */
+  readonly?: boolean
+  /** The arguments of the method, in order */
+  args: Array<{
+    /** The type of the argument */
+    type: string;
+    /** Optional, user-friendly name for the argument */
+    name?: string;
+    /** Optional, user-friendly description for the argument */
+    desc?: string;
+  }>;
+  /** Information about the method's return value */
+  returns: {
+    /** The type of the return value, or "void" to indicate no return value. */
+    type: string;
+    /** Optional, user-friendly description for the return value */
+    desc?: string;
+  };
+}
+```
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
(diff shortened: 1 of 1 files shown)


ARCs/arc-0022.md
```diff
@@ -0,0 +1,64 @@
...
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+> Comments like this are non-normative.
+
+## Read-only functions
+
+A `read-only` function is a function with no side-effects. In particular, a `read-only` function **SHOULD NOT** include:
+- local/global state modifications
+- calls to non `read-only` functions
+- inner-transactions
```
jasonpaulos, line 28: Does this mean a read-only function cannot call another read-only function from another contract? I can see arguments for and against this, but on the whole I think it would be a lot more powerful if read-only functions could do this

ARCs/arc-0022.md
```diff
@@ -0,0 +1,64 @@
...
+arc: 22
+title: Extend arc-4 to add `read-only` annotation to methods
+status: Draft
+---
+
+# Extend arc-4 to add `read-only` annotation to methods
+
+## Summary
+
+The following document introduces a convention for creating methods (as described in [ARC-4](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md)) which don't mutate state.
```
jasonpaulos, line 11: In addition to read-only, something like Solidity's pure methods also seem like a good idea. The difference between read-only and pure is that pure methods cannot read any on-chain data.

jasonpaulos (commented):
I left some feedback. Additionally, to make sure we're on the same page, from my perspective the primary benefit of having method distinctions like this would be so off-chain consumers can confidently "call" read-only/pure methods without actually sending a transaction, and the app developer only has to implement the logic once in the on-chain contract.

For this to be widely used, algod would likely need to implement a way of processing and returning read-only/pure methods without adding any transactions to the pending pool, similar to the existing dryrun endpoint.

ARCs/arc-0022.md
fabrice102: I agree we should allow that. Maybe, we can make it clearer this way:

Here is a proposal: replace 

```
- calls to non `read-only` functions
- inner-transactions
```

with

```
- inner transactions, except inner transactions that are application calls to `read only` methods of this smart contract or another smart contract
```

FrankSzendzielarz: What is the use case for either pure or readonly? Why would developers or IDE developers need have this expressed as a specific boolean instead of the information being available in the description? 

My thoughts are if there's a real need for this then it would be made more meaningful if the app call transaction readonly effect was enforced by the protocol (eg. write to global would fail).  Then in that case you might also want ambiguous methods, where write side effects are driven by arguments, then you might want to be able to call the method in 'read only mode' by specifiying that in a new property of the app call txn

fabrice102: The reason for read-only would be allow DApps to query a smart contract without having to pay transaction fees.

FrankSzendzielarz: How does the read only flag help with that?

ARCs/arc-0022.md
ori-shem-tov: We thought about adding pure functions but couldn't find an actual benefit for that. Any ideas?

FrankSzendzielarz: Could someone please explain to me what exactly is the benefit of a readonly flag or a pure flag? I am asking because I feel like I am missing something. AFAIK these ABIs are post-hoc descriptive entities and not used for driving compilation (eg: the pure flag in solidity), so how are you expecting these flags to be used and by whom?

jannotti: I think @jasonpaulos summed it up in the first comment, but I'll elaborate. If a function is marked "read-only" then an off-chain evaluator could confidently run an avm evaluator against block chain state, and get a correct answer.  This is like a view function in Ethereum.  You're not looking to send transaction to the block chain with the attendant costs and latency - you just want to run the code to get an answer where the code is really just interpreting the contracts internal state.

A pure function must the same, except running the code is even easier - it won't look at blockchain state.  So your IDE could have an AVM interpreter, or perhaps your unittest framework. And it wouldn't even need access to blockchain state to test your functions.

Both are useful for dapps that are trying to display information or build txns using code from the contract rather than interpret bytes of state on their own.

FrankSzendzielarz: Maybe I will leave this until tomorrow as my brain must have failed or something. I totally absolutely don't get it. 

So in scenario 2 above I completely ignore if someone has set a readonly flag, send my txn to the 'sandbox endpoint' or whatever it is, and if there's a state access it needs to fail anyway. Why not just have metadata in the contract that it's readonly and enforce it in the AVM? Anyway, I guess I'll leave the convo for now, I probably need a break.

jannotti: I'm having this discussion with @barnjamin right now, so I want to point out that the key benefit of an explicit "read-only" flag is that decision to do number 1 or number 2 from Jason's example can be automated.  dapp writer writes a method invocation and he gets an answer. But if it's read-only, he doesn't pay for the answer. He doesn't consult the description, read it, and decide "Oh, since it says read-only in the prose, I can invoke it the special cheap way". That happens for him.

FrankSzendzielarz: Perhaps there are also benefits to making it a property of the app call instead. Perhaps an abi method can be read only or not depending on a switch property of the txn with the app called in a read only mode. 

By the way what would be the behavkour  for group transactions including these?

SudoWeezy (approved):
Merging ARC-0022,  Extend ARC-0004 with read-only

jannotti:
Because it's machine readable

PR #83, opened 2022-04-08
Update README.md

smonn:
Add index of current ARCs to help with discoverability.

No strong opinions here, just had a hard time tracking which ARC is for what. Having a list in the main README seemed helpful.


jannotti:
Great idea. I almost wished we changed the .md filenames to be descriptive as well.

fabrice102:
Thanks! We need to automate this in the future but that's great right now.

PR #87, opened 2022-04-17
Update arc-0069.md

Vytek:
This is a proposal to modify some options of the ARC69 to meet the needs and the use of 3D Mesh for any immersive and non-immersive virtual worlds.


ARCs/arc-0069.md
```diff
@@ -31,9 +31,9 @@ The ASA parameters should follow the following conventions:
 * *Asset Name* (`an`): no restriction.
 * *Asset URL* (`au`): a URI pointing to digital media file. This URI:
     * **SHOULD** be persistent.
-    * **SHOULD** link to a file small enough to fetch quickly in a gallery view.
+    * **SHOULD** link to a file small enough to fetch quickly in a gallery view for files different from 3D Mesh. For 3D Mesh files it is **RECOMMENDED** that a client application downloads and renders them.
```
ChrisAntaki, line 34: I like the original wording on this line, it could apply to 3D meshes already

ARCs/arc-0069.md
```diff
@@ -138,7 +138,7 @@ Clients **SHOULD** render an ASA's latest ARC69 metadata. Clients **MAY** render
 
 ##### Updating ARC69 metadata
 
-If an ASA has a manager address, then the manager **MAY** update an ASA's ARC69 metadata. To do so, the manager sends a new `acfg` transaction with the entire metadata represented as JSON in the transaction's `note` field.
+If an ASA has a manager address, then the manager **MAY** update an ASA's ARC69 metadata. To do so, the manager sends a new `acfg` transaction with the entire metadata represented as JSON in the transaction's `note` field. The metadata inserted in `note`**MAY** also be serialized using MessagePack (see: https://github.com/msgpack/msgpack/blob/master/spec.md).
```
ChrisAntaki, line 141: This is a breaking change, I would be against it

ARCs/arc-0069.md
```diff
@@ -31,9 +31,9 @@ The ASA parameters should follow the following conventions:
 * *Asset Name* (`an`): no restriction.
 * *Asset URL* (`au`): a URI pointing to digital media file. This URI:
     * **SHOULD** be persistent.
-    * **SHOULD** link to a file small enough to fetch quickly in a gallery view.
+    * **SHOULD** link to a file small enough to fetch quickly in a gallery view for files different from 3D Mesh. For 3D Mesh files it is **RECOMMENDED** that a client application downloads and renders them.
     * **MUST** follow [RFC-3986](https://www.ietf.org/rfc/rfc3986.txt) and **MUST NOT** contain any whitespace character.
-    * **SHOULD** specify media type with `#` fragment identifier at end of URL. This format **MUST** follow: `#i` for images, `#v` for videos, `#a` for audio, `#p` for PDF, or `#h` for HTML/interactive digital media.  If unspecified, assume Image.
+    * **SHOULD** specify media type with `#` fragment identifier at end of URL. This format **MUST** follow: `#i` for images, `#v` for videos, `#a` for audio, `#p` for PDF, `#h` for HTML/interactive digital media, or `#m` for 3D Mesh format.  If unspecified, assume Image.
```
ChrisAntaki, line 36: Adding #m seems simple enough. Would limiting the kinds of meshes make sense though? We would want meshes a browser could render effectively

ARCs/arc-0069.md
```diff
@@ -97,7 +97,7 @@ However, contrary to the Asset URL, the `external_url` does not need to link to
 
 #### MIME Type
 
-In addition to specifying a data type in the ASA's URL (`au` field) with a URI fragment (ex: `#v` for video), the JSON Metadata schema also allows indication of the URL's MIME type (ex: `video/mp4`) via the `mime_type` field.
+In addition to specifying a data type in the ASA's URL (`au` field) with a URI fragment (ex: `#v` for video), the JSON Metadata schema also allows indication of the URL's MIME type (ex: `video/mp4`) via the `mime_type` field. For URL MIME 3D Mesh format **SHOULD** follow: https://www.iana.org/assignments/media-types/media-types.xhtml#model
```
ChrisAntaki, line 100: Sticking to meshes browsers can render effectively seems smart

ARCs/arc-0069.md
Vytek: It is true that the original wording could apply to 3d meshes. But it depends heavily on the purpose the mesh was made. A low poly could fit in a gallery.  A 3d mesh  in high definition made for historical simulations or architecture purpose could weigh up to GB.

ARCs/arc-0069.md
Vytek: The MessagePack choice would be an option (may) to optimize the 1000 bytes available for notes and therefore saving bytes for meaningful data. Besides, if I remember right, Algorand uses MessagePack to serialize and transmit the note.

ARCs/arc-0069.md
Vytek: Rendering in browsers would be easy  using the GLB format (https://www.khronos.org/gltf/) which is widely used. 
However, there are other renderers besides browsers. We must not lose the possibility of using other  clients and formats such as USDZ (https://graphics.pixar.com/usd/release/spec_usdz.html) or 3d engines not based on webGL.

ARCs/arc-0069.md
Vytek: ARC69  as a standard is meant for Digital Media Tokens. It is not limited to browsers.  Using Iana standard can ensure  a wide application in  different devices, not only mobile devices, and a wider set of applicable internet standards and protocols.

ARCs/arc-0069.md
ChrisAntaki: I've had good experiences with https://modelviewer.dev/, which supports glTF and GLB

ARCs/arc-0069.md
ChrisAntaki: ARC69 has 69%+ marketshare because it works so well in browsers

ARCs/arc-0069.md
ChrisAntaki: I'm against this breaking change

ARCs/arc-0069.md
Vytek: In this case it is not good. What if I wanted a high definition mesh NFT indicating the original binary blob in the IPFS hash? This especially makes sense when using NFT for Virtual Worlds / Metaverse

ARCs/arc-0069.md
Vytek: There is no marketshare problem. ARC 69 with these changes would continue to apply perfectly to everyone, giving the possibility of other types of uses.

ARCs/arc-0069.md
Vytek: Perfect! This does not exclude the use of other standard formats already in use and supported by other platforms.

PR #91, opened 2022-04-28
ARC-0020: Smart ASAs

aldur:
Quoting the first line of the ARC to provide quick context:

> A "Smart ASA" is an Algorand Standard Asset (ASA) controlled by a Smart Contract
 that exposes methods to create, configure, transfer, freeze, and destroy the
 asset.


brandonaaskov:
There's already quite a few commits on this branch. Any chance we can get a description of what ARC-0020 is all about?

ARCs/arc-00xx.md
```diff
@@ -0,0 +1,484 @@
...
+{
+  "name": "AssetCreate",
+  "args": [
+    { "type": "uint64", "name": "Total" },
+    { "type": "uint32", "name": "Decimals" },
+    { "type": "bool", "name": "DefaultFrozen" },
+    { "type": "string", "name": "UnitName" },
+    { "type": "string", "name": "AssetName" },
+    { "type": "string", "name": "URL" },
+    { "type": "[]byte", "name": "MetaDataHash" },
```
smonn, line 83: To follow the ABI spec[^1], this needs to be changed:

```suggestion
    { "type": "byte[]", "name": "MetaDataHash" },
```

Similar for other `[]byte` usages. Otherwise PyTEAL will break while parsing the ABI JSON schema.

[^1]: https://developer.algorand.org/docs/get-details/dapps/smart-contracts/ABI/#data-types

ARCs/arc-00xx.md
```diff
@@ -0,0 +1,484 @@
...
+  "name": "AssetConfig",
+  "args": [
+    { "type": "asset", "name": "ConfigAsset" },
+    { "type": "uint64", "name": "Total" },
+    { "type": "uint32", "name": "Decimals" },
+    { "type": "bool", "name": "DefaultFrozen" },
+    { "type": "string", "name": "UnitName" },
+    { "type": "string", "name": "AssetName" },
+    { "type": "string", "name": "URL" },
+    { "type": "[]byte", "name": "MetaDataHash" },
```
smonn, line 127:
```suggestion
    { "type": "byte[]", "name": "MetaDataHash" },
```

smonn, line 127:
```suggestion
    { "type": "byte[]", "name": "MetaDataHash" },
```

ARCs/arc-00xx.md
```diff
@@ -0,0 +1,484 @@
...
+  },
+  {
+    "name": "GetURL",
+    "args": [{ "type": "asset", "name": "Asset" }],
+    "returns": { "type": "string" }
+  },
+  {
+    "name": "GetMetaDataHash",
+    "args": [{ "type": "asset", "name": "Asset" }],
+    "returns": { "type": "[]byte" }
```
smonn, line 314:
```suggestion
    "returns": { "type": "byte[]" }
```

smonn, line 314:
```suggestion
    "returns": { "type": "byte[]" }
```

smonn (commented):
ABI type fix.

Also a reminder to rename methods and arg names to `snake_case` to match ARC-18 as dicussed here: https://github.com/algorandfoundation/ARCs/pull/70#issuecomment-1121154912

cusma:
#### Asset Configuration
```json
{
  "name": "asset_config",
  "args": [
    { "type": "asset", "name": "config_asset" },
    ...
  ],
  "returns": { "type": "void" }
}
```

I have some doubts here: I'm not convinced that the ARC should mandate `config_asset` to be `asset`. Since we are not performing _look-up_ operations on the underlying ASA a generic `uint64` could be sufficient.

🤔

aldur:
> #### Asset Configuration
> ```json
> {
>   "name": "asset_config",
>   "args": [
>     { "type": "asset", "name": "config_asset" },
>     ...
>   ],
>   "returns": { "type": "void" }
> }
> ```
> 
> I have some doubts here: I'm not convinced that the ARC should mandate `config_asset` to be `asset`. Since we are not performing _look-up_ operations on the underlying ASA a generic `uint64` could be sufficient.
> 
> 🤔

I think a _reference implementation_ might not need to loop-up for the asset, but another implementation could (think, maybe you are storing the "reserve" address directly at the ASA level instead of keeping it at the SC level). So I think this is correct

ARCs/arc-0020.md
```diff
@@ -0,0 +1,869 @@
...
+{
+  "name": "asset_create",
+  "args": [
+    { "type": "uint64", "name": "total" },
+    { "type": "uint32", "name": "decimals" },
+    { "type": "bool", "name": "default_frozen" },
+    { "type": "string", "name": "unit_name" },
+    { "type": "string", "name": "name" },
+    { "type": "string", "name": "url" },
+    { "type": "byte[]", "name": "metadata_hash" },
```
barnjamin, line 83: Shouldn't this be byte[32] since its a hash?

ARCs/arc-0020.md
```diff
@@ -0,0 +1,869 @@
...
+
+#### Asset Transfer
+
+```json
+{
+  "name": "asset_transfer",
+  "args": [
+    { "type": "asset", "name": "xfer_asset" },
+    { "type": "uint64", "name": "asset_amount" },
+    { "type": "account", "name": "asset_sender" },
```
barnjamin, line 168: The `asset_sender` field of the protocols `AssetTransferTransaction` is only for clawbacks, preferring `sender` for sending the asset. Should we keep the same naming convention here?

ARCs/arc-0020.md
```diff
@@ -0,0 +1,869 @@
...
+
+#### Asset Configuration
+
+```json
+{
+  "name": "asset_config",
+  "args": [
+    { "type": "asset", "name": "config_asset" },
+    { "type": "uint64", "name": "total" },
+    { "type": "uint32", "name": "decimals" },
```
barnjamin, line 122: I'm confused about how these will work in practice, if an asset was created with total 10 and decimals 1, then this method was called setting total to 100 and decimals 0, what _actually_ happens? The underlying ASA can't have its supply increased to 100 units since that field is immutable

ARCs/arc-0020.md
```diff
@@ -0,0 +1,869 @@
...
+>
+> The controlling Smart Contract SHOULD perform an asset destroy operation on
+> the ASA with ID `destroy_asset`. The operation will fail if the asset is still
+> in circulation.
+
+#### Circulating Supply
+
+```json
+{
+  "name": "circulating_supply",
```
barnjamin, line 307: Should we have a naming convention for read-only methods that they start with `get_*`?

ARCs/arc-0020.md
aldur: It _could_ be, but the interface can remain flexible. An implementation can then decide to reject `metadata_hash` if longer than 32 bytes.

ARCs/arc-0020.md
aldur: Other `get_*` methods might do math as well, so I'd be ok with renaming this `get_circulating_supply`.

ARCs/arc-0020.md
aldur: I _think_ here we want this semantic:

- `sender` is always the user triggering the call, i.e. issuing the transaction
- Assuming permissions and roles, the call moves asset from A (`asset_sender`) to B (`asset_receiver`)
- `asset_sender`  COULD be `sender` and will probably be in most calls

cusma: I agree. The `sender` is by default the issuer of the ABI call transaction. If `sender` is different from `asset_sender` (which should always be specified since ABI standard doesn't have optional arguments) the this is `asset_transfer` is interpreted as `clawback` and the consequent checks must follow.

ARCs/arc-0020.md
cusma: I fine with standardise `get_` naming convention.

nullun (commented):
Is this a TEAL reference implementation in progress anywhere?

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+A "Smart ASA" is an Algorand Standard Asset (ASA) controlled by a Smart Contract
+that exposes methods to create, configure, transfer, freeze, and destroy the
+asset.
+
+This ARC defines the ABI interface of such Smart Contract, the required
+metadata, and suggests a reference implementation.
+
+## Motivation
+
+The Algorand Standard Asset (ASA) is an excellent building block for on-chain
```
robdmoore, line 28: Not sure if worth adding, but something I really appreciate about ASAs: As a first class mechanism it also supports a standard, reliable way of representing and querying ownership and history, which is very powerful.

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+applications. It is battle-tested and widely supported by SDKs, wallets, and
+dApps.
+
+However, the ASA lacks in flexibility and configurability. For instance, once
+issued it can't be re-configured (its unit name, decimals, maximum supply).
+Also, it is freely transferable (unless frozen). This prevents developers from
+specifying additional business logic to be checked while transferring it (think
+of royalties or [vesting](https://en.wikipedia.org/wiki/Vesting)).
+
+Enforcing transfer conditions require freezing the asset and transferring it
```
robdmoore, line 38: ... and it's also not supported by any marketplaces and also requires a great deal of trust to be placed in the creator of the asset who could easily rug pull the assets.

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+through a clawback operation -- which results in a process that is opaque to
+users and wallets and a bad experience for the users.
+
+The Smart ASA defined by this ARC extends the ASA to increase its expressiveness
```
robdmoore, line 42: Minor suggestion (speaks to the purpose of it being a standard and the flow on effect of having confidence supporting it across the ecosystem - people could do this today, but nothing supports it).

From:

> Developers can now introduce additional business logic around its operations. Wallets, dApps, and SDKs can recognize Smart ASAs and adjust their flows and user experiences accordingly.

To:

> By introducing this as a standard, marketplaces, wallets, dApps, and SDKs can confidently and consistently recognize Smart ASAs and adjust their flows and user experiences accordingly.

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+  ],
+  "returns": { "type": "void" }
+}
+```
+
+Calling `asset_config` configures an existing Smart ASA.
+
+> Upon a call to `asset_config`, a reference implementation SHOULD:
+>
+> - Fail if `config_asset` does not correspond to an ASA controlled by this smart
```
robdmoore, line 141: Presumably this is done by checking clawback is set to the address of the current app? Should this explicitly be mentioned?

aldur: I would consider this an implementation detail: your suggestion works, but you could even hardcode the list of assets you want to manage in the TEAL, for instance.

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+  ],
+  "returns": { "type": "void" }
+}
+```
+
+Calling `asset_transfer` transfers a Smart ASA.
+
+> Upon a call to `asset_transfer`, a reference implementation SHOULD:
+>
+> - Fail if `xfer_asset` does not correspond to an ASA controlled by this smart
```
robdmoore, line 179: Should there be mention here of looking up the frozen statuses set by `asset_freeze` and `account_freeze`?

ARCs/arc-0020.md
robdmoore: +1 on `get_`

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+```
+
+Calling `asset_create` creates a new Smart ASA and returns the identifier of the
+ASA. The [metadata section](#metadata) describes its required properties.
+
+> Upon a call to `asset_create`, a reference implementation SHOULD:
+>
+> - Mint an Algorand Standard Asset (ASA) that MUST specify the properties
+>   defined in the [metadata section](#metadata). In addition:
+>   - The `manager`, `reserve` and `freeze` addresses SHOULD be set to the
```
robdmoore, line 100: This would preclude using ARC-19 in conjunction with ARC-20, but maybe that's ok since ARC-20 allows you to have essentially a dynamic URL anyway? In which case, it raises the question of what `SHOULD` the smart contract add for the URL field in the ASA itself, and furthermore, if you wanted to support a metadata standard with traits for a smart ASA which standard(s) do we want to support? Should the ABI for `asset_create` and `asset_config` support taking transaction note, or should we say that the smart ASA tracked URL field should always be used (forcing you to store metadata off-chain e.g. IPFS). When box storage comes along that maybe gives a third option that's completely on-chain?

If we do say that the URL field should always be looked up via the smart contract rather than the ASA then possibly we can repurpose the URL field on the ASA to indicate it's a smart ASA (e.g. `algorand://arc-20` or `algorand://{app id of managing contract}#arc-20` or something), rather than the unit name / name as indicated below (which I personally think would be a bit messy given it's the default bit of text you would see about any ASA while viewing it and waiting for other data to load about the ASA (e.g. while looking up IPFS / indexer etc.).

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+> It is RECOMMENDED for calls to this method to be permissioned, e.g. to only
+> approve transactions issued by the controlling Smart Contract creator.
+
+#### Asset Configuration
+
+```json
+{
+  "name": "asset_config",
+  "args": [
+    { "type": "asset", "name": "config_asset" },
```
robdmoore, line 120: If you only want to set one of these you can't really do that can you? You always have to provide everything yeah? This should probably be mentioned.... That makes it hard to do idempotent partial updates (but you can use the `lease` field with the app call to achieve it and it's not something that ASAs support anyway).

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+>
+> - Fail if `asset` does not correspond to an ASA controlled by this smart
+>   contract.
+> - Return the circulating supply of `asset`, defined by the difference between
+>   the ASA `total` and the balance held by its `reserve_addr` (see [Asset
+>   Creation](#asset_creation)).
+
+#### Getters
+
+```json
```
robdmoore, line 326: Should there be a getter that returns all of them in one payload too for convenience, or is the intention these are always going to be called from within smart contracts?

Making an ABI call to a `get_*` method is stored on the blockchain, you can't do that as an offline operation so it costs ALGO to execute a get?

Related: has any thought been given to how smart ASAs can be indexed without clogging the blockchain with heaps of get calls?

I guess you could index the app calls and make assumptions that the values passed in are persisted if the call succeeded. Not as clean as being able to call algod api to return current state of param values for an ASA (for no cost) since you now have to have an indexer. I guess the Algorand indexer could be modified to detect and support indexing smart ASAs though?

barnjamin: I think the get_* methods should be marked as `read-only` as in ARC22 https://github.com/algorandfoundation/ARCs/pull/79

If they are, we should be able to call against dryrun or the in progress `simulate` endpoint which should be at no cost https://github.com/algorand/go-algorand/pull/4322

ARCs/arc-0020.md
robdmoore: Hmm, just read the "Specifying the controlling Smart Contract" section. Given this contract controls the creation of the smart contract my point about supporting `note` stands since if you want to use ARC-69 (unsure if the contract trusts that the identification of the smart contract is included in the ARC-69 payload or not (I guess it has no choice but to trust it for ARC-3 anyway).

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+
+> Smart ASAs SHOULD NOT require users to opt in to the controlling Smart
+> Contract, unless the implementation requires storing information into their
+> local schema (for instance, to implement [freezing](#asset-freeze); also see
+> [security considerations](#security-considerations)).
+>
+> Clients MAY inspect the local state schema of the controlling Smart Contract
+> to infer whether opt in is required.
+>
+> If a Smart ASA requires opt in, then clients SHOULD prevent users from closing
```
robdmoore, line 816: Hmm, this one gets dicey if you use the contract to create multiple ASAs.

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+> out the controlling Smart Contract unless they don't hold an ASA balance.
+
+## Rationale
+
+This ARC builds on the strengths of the ASA to enable a Smart Contract to
+control its operations and flexibly re-configure its configuration.
+
+The rationale is to have a "Smart ASA" that is as widely adopted as the ASA both
+by the community and by the surrounding ecosystem. Wallets and dApps:
```
robdmoore, line 825: ... and marketplaces (I guess you could say they are a dApp, but I think they are important enough to be worth mentioning.

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+  transactions, which call the ABI methods of the controlling Smart Contract.
+
+With this in mind, this standard optimizes for:
+
+- Community adoption, by minimizing the [ASA metadata](#metadata) that need to
+  be set and the requirements of a conforming implementation.
+- Developer adoption, by re-using the familiar ASA transaction reference in the
+  methods' specification.
+- Ecosystem integration, by minimizing the amount of work that a wallet, dApp or
+  service should perform to support the Smart ASA.
```
robdmoore, line 840: Suggestion (ASAs handle privacy (opt-in), first-class history and first-class ownership):

> - End user experience, by wrapping the user friendly ASA construct

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+TODO.
+
+## Security Considerations
+
+Keep in mind that the rules governing a Smart ASA are only in place as long as:
+
+- The ASA remains frozen;
+- the `ClawbackAddr` of the ASA is set to a controlling Smart Contract, as
+  specified in the [metadata section](#metadata);
+- the controlling Smart Contract is not updatable, nor deletable, nor
```
robdmoore, line 875: Suggestion(?):

> - The controlling Smart Contract is a superset of the ABI defined by this standard

ARCs/arc-0020.md
robdmoore: @aldur reminded me that you can actually issue a dry run to call a get operation. That's pretty cool, but should definitely be mentioned / recommended (and definitely means there should be a `get_all_fields` method so you can achieve it in a single HTTP request to algod).

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+
+If your controlling Smart Contract implementation writes information to a user's
+local state, keep in mind that users can close out the application and (worse)
+clear their state at all times. This requires careful considerations.
+
+For instance, if you determine a user's [freeze](#asset-freeze) state by reading
+a flag into their local state, you should:
+
+- Set the flag to `frozen` at opt in;
+- explicitly verify that a user's `frozen` flag is `0` before approving
```
robdmoore, line 888: Minor: Should be capital `E`

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+### Local State
+
+If your controlling Smart Contract implementation writes information to a user's
+local state, keep in mind that users can close out the application and (worse)
+clear their state at all times. This requires careful considerations.
+
+For instance, if you determine a user's [freeze](#asset-freeze) state by reading
+a flag into their local state, you should:
+
+- Set the flag to `frozen` at opt in;
```
robdmoore, line 887: For consistency with above `frozen` in this section should be `asset_frozen`.

Also, how does this work when there's multiple ASAs associated with a single contract?

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+> Upon a call to `asset_create`, a reference implementation SHOULD:
+>
+> - Mint an Algorand Standard Asset (ASA) that MUST specify the properties
+>   defined in the [metadata section](#metadata). In addition:
+>   - The `manager`, `reserve` and `freeze` addresses SHOULD be set to the
+>     account of the controlling Smart Contract.
+>   - The remaining fields are left to the implementation, which MAY set `total`
+>     to `2 ** 64 - 1` to enable dynamically increasing the max circulating
+>     supply of the Smart ASA.
+>   - `name` and `unit_name` MAY be set to `SMART-ASA` and `S-ASA`, to
```
robdmoore, line 105: Most wallets, explorers, etc. are going to display these names so I'm not sure if this is a good recommendation (most end users aren't going to care). Maybe this can be like the ARC-3 recommendation where you add `@arc20` to name or `#arc20` to URL and clients shouldn't display it?

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+#### Asset Configuration
+
+```json
+{
+  "name": "asset_config",
+  "args": [
+    { "type": "asset", "name": "config_asset" },
+    { "type": "uint64", "name": "total" },
+    { "type": "uint32", "name": "decimals" },
+    { "type": "bool", "name": "default_frozen" },
```
robdmoore, line 123: What does this field mean / how is it used for a smart ASA? Noting the recommendation in the security section - maybe it should be used in conjunction with that recommendation if you care about freezing?

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+> - Mint an Algorand Standard Asset (ASA) that MUST specify the properties
+>   defined in the [metadata section](#metadata). In addition:
+>   - The `manager`, `reserve` and `freeze` addresses SHOULD be set to the
+>     account of the controlling Smart Contract.
+>   - The remaining fields are left to the implementation, which MAY set `total`
+>     to `2 ** 64 - 1` to enable dynamically increasing the max circulating
+>     supply of the Smart ASA.
+>   - `name` and `unit_name` MAY be set to `SMART-ASA` and `S-ASA`, to
+>     denote that this ASA is Smart and has a controlling application.
+> - Persist the `total`, `decimals`, `default_frozen`, etc. fields for later
```
robdmoore, line 107: There are a number of places in this spec that talk about persisting information:

- These parameters (`total`, `decimals`, etc.)
- The `manager_addr` for asset config
- `asset_freeze` (recommends global)
- `account_freeze` (recommends local)

If you are controlling multiple ASAs (e.g. hundreds / thousands) from the same contract (which I think is an important implementation option) then this is non-trivial to implement. I think there should be a section about how to handle multiple assets that talks through this. #bringonboxstorage

ARCs/arc-0020.md
robdmoore: Yeah, that's fair. Is it worth listing a couple of them as examples or as a `MAY`?

robdmoore:
Hi all,

Sorry for spamming with so many comments, wanted to get really deep into this and understand it fully since I'm sure I'll be implementing it at some point 🤣 

For context: I'm one of the creators of [Data History Museum](https://museum.datahistory.org/) and have created a general purpose dynamic / high scale minting platform, [onRAMP](https://www.onramp.makerx.com.au/) (that now runs the Data History Museum) so I have a vested interest in this standard since I'm likely going to want to create a bunch of automation around smart ASAs in the future.

ARCs/arc-0020.md
robdmoore: (sorry, missed your comment there @barnjamin - great callout!)

robdmoore:
@aldur you missed `circulating_supply` in your commit to rectify the `get_` stuff FYI

ARCs/arc-0020.md
cusma: To:

> By introducing this as a standard both developers, users (marketplaces, wallets, dApps, etc.) and SDKs can confidently and consistently recognize Smart ASAs and adjust their flows and user experiences accordingly.

ARCs/arc-0020.md
cusma: >This would preclude using ARC-19 in conjunction with ARC-20, but maybe that's ok since ARC-20 allows you to have essentially a dynamic URL anyway?

@robdmoore ARC-20 is so general that you can dynamically change whatever `field` you want, not just the `URL`, also the `name`, the `unit_name` and so on.

The binding between the "underlying ASA" and the Smart ASA App can be done in several ways, personally I'm support the option to use an AVM-accessible field to programmatically verify the binding, for example: the `URL` field of the "underlying ASA", not to be confused with the `URL` of the Smart ASA, which could be set to whatever value you want (even bigger than 96 kB) and changed over time.

With respect to traits: in addition to regular Smart ASA field you may add to the Smart ASA state whatever traits you want and even have a specific on-chain logic to change each trait, which is accessible to the AVM (just like any variable).

ARCs/arc-0020.md
cusma: The ABI requires all the arguments to be passed, so you can't have an "optional" with default value. This being said the ARC-20 is general, it's then a matter of the implementation (or event the client) to automate the process to handle default value.

robdmoore: That's fair, although potentially has implications on indexing (per my other comment about that).

ARCs/arc-0020.md
robdmoore: Fair enough, and that would make sense to be a separate ARC standard in the future so marketplaces have a standard way of determining traits.

robdmoore: (but to enable ARC-69 support this interface needs transaction note I think).

ARCs/arc-0020.md
robdmoore: (and for clarity - agree they should be marked readonly, but also I think we need an endpoint that returns everything so dApps are more performant).

ARCs/arc-0020.md
cusma: The ARC-20 does not prevent addition of a `note` along with the `InnerTxn` resulting from its ABI calls. So it's maybe more a matter of the ARC-69 to specify that in case the asset is created as Smart ASA the implementation must take care of assembling the `note` of the `InnerTxn` according to the ABI call that generates the `AssetCreate` or `AssetConfig`. WDYT?

ARCs/arc-0020.md
aldur: Marked read-only methods, will also add a unique endpoint to get the config 👍

ARCs/arc-0020.md
```diff
@@ -0,0 +1,870 @@
...
+([ARC-3](./arc-0003.md), [ARC-69](./arc-0069.md)), then it MUST include a
+`arc-20` key and set the corresponding value to a map, including the ID of the
+controlling Smart Contract as a value for the key `application-id`. For example:
+
+```javascript
+{
+  //...
+  "properties": {
+    //...
+    "arc-20": {
```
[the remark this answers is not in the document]
aldur: Is there a conflict of sorts? Otherwise, I'd keep things separated just to not block this/arc-16.

ARCs/arc-0020.md
aldur: True and also opt-in/opt-out is tricky. Eventually, we'll just use boxes. For now I have updated the prose.

ARCs/arc-0020.md
aldur: I am conflicted here; even if the controlling SC implements the ABI, that doesn't mean that rules won't change (e.g. if you updated it, etc).

ARCs/arc-0020.md
aldur: Non-native English speaker, but: even after a semicolon? 🤔

ARCs/arc-0020.md
aldur: The _how_ is implementation-dependent. The spec mimics the ASA, so a Smart ASA that wants to maximise similarities with ASAs should make sure that, if `default_frozen` is set, then users are frozen by default unless unfrozen explicitly (per asset).

The security recommendation section states that if you use local storage and you don't find a flag, expect the worse (in case of freezing, that user is frozen).

aldur: I have updated the prose around the security consideration, hope it helps.

ARCs/arc-0020.md
aldur: True. 

I think that section should NOT be normative, but I can't write it before the design around boxes crystallises.

One thing we could write is how to handle multiple ASAs that share the same configuration -- meaning that require constant storage.

ARCs/arc-0020.md
aldur: I'd prefer keeping this as light as possible, but I am open to discussion.

What I'd like to stand from the prose is "if the controlling SC created / configured the ASA, then it controls it". Also, this means that when creating it or configuring it, it should ensure control and right syntax/semantics.

If this doesn't show from the prose, then yes, we need to add examples. :)

ARCs/arc-0020.md
aldur: Added a unique endpoint as well (replacing all the others to prevent redundancy).

aldur:
@robdmoore thank you for the comments!

I think I have now addressed most of them. There's still something missing related to handling multiple ASAs with a single SC. That is a helpful part to have but tricky to write without boxes. Since it will be NON-normative, I would be inclined to merge this and then have a second PR to add that.

ARCs/arc-0020.md
aldur: I am open to follow this. The trouble is that if you later update the name (smart ASA allows that), what's specified in the ASA will NOT be updated. How do we feel about that?

ARCs/arc-0020.md
robdmoore: Yeah if it's a bullet list I'd expect all of the items to start with a capital

robdmoore: So there's three styles of punctuation for bullet lists (and actually we were both wrong :P).

https://www.instructionalsolutions.com/blog/bulleted-list-punctuation

If using semi colons then all the bullets should be lowercase (i.e. the `Set` should be `set`) and there should be a full stop at the end of the last bullet and the second last should end in `; and`.

ARCs/arc-0020.md
robdmoore: My gut feel is that the name here should be a helpful placeholder.

If you have a UI that asynchronously loads the information it has to hand as it has then it can:

* Show the asset name (from the ASA) first (potentially stripping out `@arc20` if present?)
* Make a call to the underlying smart contract (e.g. by reverse lookup from clawback, or by using the URL field maybe? I note: the reference implementation I found hidden in Beaker used URL field to store a lookup of the app id)
* Once that call came back the UI could cache that value and show the proper name

I think that's a much better user experience for users to see a relevant "holding name" than seeing "SMART-ASA".

ARCs/arc-0020.md
robdmoore: Makes sense.

As a suggestion, can we at least have a section that discusses the fact that this standard is intended to support issuing multiple ASAs, but aspects of that implementation may be harder to implement in lieu of box storage given the limited amount of global storage available? That way it's clear to people this standard is open to supporting that pending people figuring out how to implement it in the future.

ARCs/arc-0020.md
```diff
@@ -116,24 +116,35 @@ ASA. The [metadata section](#metadata) describes its required properties.
...
+      { "type": "string", "name": "url" },
+      { "type": "byte[]", "name": "metadata_hash" },
+      { "type": "address", "name": "manager_addr" },
+      { "type": "address", "name": "reserve_addr" },
+      { "type": "address", "name": "freeze_addr" },
+      { "type": "address", "name": "clawback_addr" }
+    ],
+    "returns": { "type": "void" }
+  },
+  {
```
robdmoore, line 138: <3

ARCs/arc-0020.md
```diff
@@ -645,149 +596,6 @@ Calling `get_FIELD` reads and returns the `asset`'s `FIELD` as specified in:
       "returns": {
         "type": "uint64"
       }
-    },
-    {
-      "name": "get_total",
```
robdmoore: Love the new asset config endpoint! Wondering if these are needed as well or not? Namely, for situations where you may want to call these within a stateful smart contract? Will a stateful smart contract be able to pull out individual tuple values from `get_asset_config` or not (I've never tried that before)?

SudoWeezy:
As ARC-18 has been added to the repo.  
I am requesting a review to know if we can merge it too.

ARCs/arc-0020.md
```diff
@@ -0,0 +1,747 @@
+---
+arc: 0020
+title: Smart ASA
+description: An ARC for an ASA controlled by an Algorand Smart Contract
+author:
+  - Cosimo Bassi (@cusma)
+  - Adriano Di Luzio (@aldur)
+  - John Jannotti (@jannotti)
+discussions-to: <URL>
```
cusma, line 9:
```suggestion
discussions-to: #109 
```

ARCs/arc-0020.md
```diff
@@ -0,0 +1,747 @@
...
+>
+> If a Smart ASA requires opt in, then clients SHOULD prevent users from closing
+> out the controlling Smart Contract unless they don't hold a balance for any of
+> the ASAs controlled by the Smart Contract.
+
+## Rationale
+
+This ARC builds on the strengths of the ASA to enable a Smart Contract to
+control its operations and flexibly re-configure its configuration.
+
```
cusma, line 665:
```suggestion

ARC-20 is meant to be an "extension" of ASAs, rather than a "replacement" of the existing ASA framework with an approach purely based on Smart Contract, _a-la Ethereum_. Regular ASAs should always be intended as _go-to_ solutions if they cover all the requirements of the specific use-case. The Smart ASA "extension" aims to piggyback on ASAs battle-tested security, their users and tooling adoption as well as preserving their backward compatibility with ASA-based ARCs (e.g. ARC-3, ARC-69, etc.).

```

cusma (approved):

PR #95, opened 2022-05-10
ARC-0001 & ARC-0005-ARC-0011 & ARC-0025: Pre finalizing

ori-shem-tov:
trying to finalize all the wallet related ARCs


ARCs/arc-0007.md
```diff
@@ -28,43 +28,43 @@ This ARC uses interchangeably the terms "throw an error" and "reject a promise w
...
   data?: any;
-  successTxnIds: (TxnId | null)[];
+  successTxnIDs: (TxnID | null)[];
 }
 ```
 
 A `PostTxnsFunction` with input argument `stxns:string[]` and promised return value `ret:PostTxnsResult`:
 * expects `stxns` to be in the correct string format as specified by `SignedTxnStr`.
-* **MUST**, if successful, return an object `ret` such that `ret.txId` is in the correct string format as specified by `TxId`.
+* **MUST**, if successful, return an object `ret` such that `ret.txID` is in the correct string format as specified by `TxID`.
```
fabrice102, line 51: I think it is `txnIDs`?

fabrice102, line 51:
```suggestion
* **MUST**, if successful, return an object `ret` such that `ret.txID` is in the correct string format as specified by `TxID` (defined below).
```

ARCs/arc-0007.md
```diff
@@ -28,43 +28,43 @@ This ARC uses interchangeably the terms "throw an error" and "reject a promise w
...
 export interface PostTxnsError extends Error {
   code: number;
   data?: any;
-  successTxnIds: (TxnId | null)[];
+  successTxnIDs: (TxnID | null)[];
 }
 ```
 
 A `PostTxnsFunction` with input argument `stxns:string[]` and promised return value `ret:PostTxnsResult`:
 * expects `stxns` to be in the correct string format as specified by `SignedTxnStr`.
```
fabrice102, line 50:
```suggestion
* expects `stxns` to be in the correct string format as specified by `SignedTxnStr` (defined below).
```

ARCs/arc-0007.md
```diff
@@ -28,43 +28,43 @@ This ARC uses interchangeably the terms "throw an error" and "reject a promise w
...
-Defined as in [ARC-0001](arc-0001.md):
+Defined as in [ARC-0001](arc-0001.md#interface-signedtxnstr):
 
 > [`SignedTxnStr` is] the base64 encoding of the canonical msgpack encoding of the `SignedTxn` corresponding object, as defined in the [Algorand specs](https://github.com/algorandfoundation/specs).
 
-### String specification: `TxnId`
+### String specification: `TxnID`
 
-A `TxnId` is a 52-character base32 string (without padding) corresponding to a 32-byte string;
+A `TxnID` is a 52-character base32 string (without padding) corresponding to a 32-byte string;
```
fabrice102, line 63:
```suggestion
A `TxnID` is a 52-character base32 string (without padding) corresponding to a 32-byte string.
For example: `H2KKVITXKWL2VBZBWNHSYNU3DBLYBXQAVPFPXBCJ6ZZDVXQPSRTQ`.
```

ARCs/arc-0007.md
```diff
@@ -86,11 +86,11 @@ Regarding a call to `postTxns(stxns)` with promised return value `ret`:
...
+* If successful, `postTxns` **MUST** resolve the returned promise with the list of transaction IDs `txnIDs` of the posted transactions `stxn`.
 * If unsuccessful, `postTxns` **MUST** reject the promise with an error `err` of type `PostTxnsError` such that:
-  * `err.code=4400` if there was a failure sending the transactions or a code as specified in [ARC-0001](arc-0001.md) if the user or function disallowed posting the transactions.
+  * `err.code=4400` if there was a failure sending the transactions or a code as specified in [ARC-0001](arc-0001.md#error-standards) if the user or function disallowed posting the transactions.
   * `err.message` **SHOULD** describe what went wrong in as much detail as possible.
-  * `err.successTxnIds` **MUST** be an array such that `err.successTxnId[i]` is the transaction ID of `stxns[i]` if `stxns[i]` was succesfully committed to the blockchain, and `null` otherwise.
+  * `err.successTxnIDs` **MUST** be an array such that `err.successTxnID[i]` is the transaction ID of `stxns[i]` if `stxns[i]` was succesfully committed to the blockchain, and `null` otherwise.
```
fabrice102, line 93:
```suggestion
  * `err.successTxnIDs` **MUST** be an array such that `err.successTxnID[i]` is the transaction ID of `stxns[i]` if `stxns[i]` was successfully committed to the blockchain, and `null` otherwise.
```

fabrice102 (approved):
Looks good! Make a few minor suggestions.

ARCs/arc-0007.md
```diff
@@ -28,43 +28,43 @@ This ARC uses interchangeably the terms "throw an error" and "reject a promise w
...
+export type TxnID = string;
 export type SignedTxnStr = string;
 
 export type PostTxnsFunction = (
   stxns: SignedTxnStr[],
 ) => Promise<PostTxnsResult>;
 
 export interface PostTxnsResult {
-  txnIds: TxnId[];
+  txnIDs: TxnID[];
```
barnjamin, line 39: The struct defined in go-algorand has this as `TxID` not `TxnID`, should they match?


https://github.com/algorand/go-algorand/blob/bd18d0497f4a89b28fdd54811f95444adde901a4/daemon/algod/api/server/v2/generated/types.go#L704

ARCs/arc-0007.md
barnjamin: I'm not sure which is preferable but go-algorand has this as TxID

https://github.com/algorand/go-algorand/blob/bd18d0497f4a89b28fdd54811f95444adde901a4/daemon/algod/api/server/v2/generated/types.go#L704

barnjamin:
This looks good pending resolution of `TxnID` vs `TxID`

ARCs/arc-0007.md
DanBurton: I have no strong opinions on which it should be, but the doc should at least be consistent with itself.

DanBurton (approved):
These changes lgtm. I only had one minor remark about `txnID` vs `txID` (pick one).

ori-shem-tov (commented):
I suggest we use `TxnID` even though go-algorand uses `TxID`, since all the related ARCs use `T\txn` and not `T\tx`

fabrice102:
@SudoWeezy One thing I realized we need to add in ARCs 7-9 is a "Security Considerations" section with a text like:

> In case the wallet uses an API service that is secret or provided by the user, the wallet **MUST** ensure that the URL of the service and the potential tokens/headers are not leaked to the dApp. 
> > Leakage may happen by accidentally including too much information in responses or errors returned by the various methods. For example, if the nodeJS `superagent` library is used without any filtering of errors and responses, errors and responses may include the request object which itself include the potentially secret API service URL / secret token headers.

PR #97, opened 2022-05-16
Create ARC for extension of ARC-06 with account switch functionality

robert-zaremba:
Wallet is a gateway for dapps. For descent UX and wallet integration with a dapp we need to be able to instrument a wallet to execute a switch as well as register hooks for:
* network switch
* account switch

### Rationale

Very often webapps have their own state and remember the network and user account. For example a webapp can use an indexer to show all pools where a user has a stake in a given Dapp. The webapp will use network genesis hash and user account address to query the pools.


SudoWeezy:
If this is still relevant, please consider merging the version in the repo before submitting it.

[ARC-6](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0006.md)

SudoWeezy:
@DanBurton, could you please give me your opinion on this?

DanBurton:
I think this is good but should be its own follow-on ARC. Maybe we should combine it with #74.

SudoWeezy:
@DanBurton, I merged your previous PR #74 into it.
Do you want this to be an extension rather than updating the current ARC-6?

DanBurton:
Sorry for the delayed response. I am ok with updating the current ARC-6.

SudoWeezy:
@DanBurton & @robert-zaremba 
If the changes I made to merge the PR properly are ok for you, we can go like this.

DanBurton:
👍 Looks good to me, @SudoWeezy.

PR #102, opened 2022-05-18
ARC-0027 :  Conventions for Digital 3D Media Tokens

Vytek:
New ARC 27 for Community Algorand Standard Asset Parameters Conventions for Digital 3D Media Tokens (Metaverse, Virtual Words, Games, etc...)


Vytek:
We are studing the draft for modify and update.

Vytek:
I arrived at this demo by seeing this ad that was relaunched by the Algorand Foundation account on Twitter:
https://twitter.com/myalgo_/status/1590722388613660672

You can analyze how it works with these explorers:
https://algoexplorer.io/asset/472783898
https://app.dappflow.org/explorer/asset/472783898/transactions

An analysis of the ARC69 can be found here:

https://app.dappflow.org/arc-portal/arc/69/workspace?asset_id=472783898

Here is the data present in the NFT:

```
{
  "index": 472783898,
  "params": {
    "creator": "METASWXOZB3CFFNWD6BDWU7CG5E42HNWFJZMM6IWR7MCT4P7NDW6755IMM",
    "decimals": 0,
    "default-frozen": false,
    "manager": "METASWXOZB3CFFNWD6BDWU7CG5E42HNWFJZMM6IWR7MCT4P7NDW6755IMM",
    "name": "Metapunk #415",
    "name-b64": "TWV0YXB1bmsgIzQxNQ==",
    "total": 1,
    "unit-name": "META415",
    "unit-name-b64": "TUVUQTQxNQ==",
    "url": "https://arweave.net/I10X177mZUZbVSXPRMRJDHiW_cHXgfhtpkMrYN8Zxwo",
    "url-b64": "aHR0cHM6Ly9hcndlYXZlLm5ldC9JMTBYMTc3bVpVWmJWU1hQUk1SSkRIaVdfY0hYZ2ZodHBrTXJZTjhaeHdv"
  }
}
```

As you can see the URL reference is towards an image which is the one that is displayed in the gallery and also inside the mobile phone for example Pera Wallet.

Let's now pass to the METADATA obviously present in the Notes field of the NFT creation transaction (Total 1, Decimal 0):

```
{
  "standard": "arc69",
  "external_url": "https://app.metapunks.world/metapunk/415",
  "media_url": "https://arweave.net/3YspqAa51DEsIsCMDLs5_C52BtHPoulLev3rXkINlII",
  "mime_type": "model/gltf-binary",
  "attributes": [
    {
      "trait_type": "Lighting",
      "value": "Only Ton Forgives"
    },
    {
      "trait_type": "Background",
      "value": "Solid Look"
    },
    {
      "trait_type": "Face",
      "value": "Light #2"
    },
    {
      "trait_type": "Torso",
      "value": "Stylish Puffer"
    },
    {
      "trait_type": "Legs",
      "value": "Jeans #2"
    },
    {
      "trait_type": "Shoes",
      "value": "Cosmic"
    },
    {
      "trait_type": "Extra #2",
      "value": "Headphones"
    },
    {
      "trait_type": "Eyewear",
      "value": "None"
    },
    {
      "trait_type": "Extra #1",
      "value": "None"
    },
    {
      "trait_type": "Backpack",
      "value": "None"
    },
    {
      "trait_type": "Head",
      "value": "None"
    }
  ],
  "description": "",
  "properties": {}
}
```

The media url refers to the three-dimensional MESH in glTF format, external URL refers from a landing page for the 3D Mesh: https://app.metapunks.world/metapunk/415 (Created by Metapunk)

Problems encountered

1) No control hash of the different media is defined. The use of platforms such as IPFS and Arweave are used for the calculation of the HASH used directly in the reference URL. It is therefore not possible to make a modification to the object without changing the URL reference that is entered and notarized in a secure transaction on the BC Algorand. In my opinion it is wrong, the Asset Metadata Hash field must be mandatory and must:

1) Be the HASH SHA256 of MESH 3D so you can check it on any platform once downloaded
2) At the limit, a HASH SHA256 must be reported for each URL present in the METADATA (Notes field of the transaction creating the Algorand) in order to verify all the elements and not rely only and only on the URLs generated by IPFS and arweave. This is to avoid that the same file can be used in two different NFTs not giving the possibility to the platform (I assume any platform but in particular a Virtual World and in the future the Metaverse) to have a trivial control through a HASH control. Obviously there remains the problem of MESH modify to which a HASH cannot give feedback, it will probably be necessary to use and add further Metadata always within the limit of 1000 Bytes.

So much so that the field is defined optional:
Asset Metadata Hash (am): the SHA-256 digest of the full resolution media file as a 32-byte string (as defined in NIST FIPS 180-4)
OPTIONAL

2) The specification: https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0069.md defines that:
}, "mime_type": {"type": "string", "description": "Describes the MIME type of the ASA's URL (` au` field). "

So mime_type refers to the AU field which is the address of the image NOT of the 3D mesh object instead contained in the METADATA present in the notes. And even if the mime_type field is fine, which field does media_url or external_url refer to? It is not clear and even contrary to the specification.

3) The specification says that:
SHOULD specify media type with # fragment identifier at end of URL. This format MUST follow: #i for images, #v for videos, #a for audio, #p for PDF, or #h for HTML / interactive digital media. If unspecified, assume Image.

And in the case of 3D Meshes defined Model and therefore a new #m according to the specification? And why is it used only and only in the case of IPFS addresses and not for arweare references? What if other decentralization systems were used? There is little generalization and indeed standardization that remains not general and abstract.

PR #107, opened 2022-06-18
Moving ARC19 status to Final.

pbennett:
ARC19 should realistically be considered final at this point as almost 14K ARC19 NFTs have already been minted - all completely immutable.  
Any changes to the URL spec of ARC19 at this point will likely need to be an independent ARC.


barnjamin (approved):
@fabrice102

barnjamin:
@fabrice102 can we merge this?

PR #113, opened 2022-07-18
ARC-0028 : Events

DanBurton:


ARCs/arc-0028.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Deriving the 4-byte prefix from the Event Signature
+
+To derive the 4-byte prefix from the Event Signature, perform the `sha512/256` hash algorithm on the signature, and select the first 4 bytes of the result.
+
+This is the same process that is used by the [ABI Method Selector as specified in ARC 4](./arc-0004.md#method-selector).
+
+### Sample interpretation of Event log data
+
+An exchange dapp might emit a `Swapped` event with two `uint64` values representing quantities of currency swapped. The event signature would be: `Swapped(uint64,uint64)`.
```
barnjamin, line 39: should we stick with lowercase names by default?

ARCs/arc-0028.md
```diff
@@ -0,0 +1,81 @@
...
+
+First, we can determine the expected 4-byte prefix by following the spec above:
+
+```js
+> { sha512_256 } = require('js-sha512')
+> sig = 'Swapped(uint64,uint64)'
+'Swapped(uint64,uint64)'
+> hash = sha512_256(sig)
+'1ccbd9254b9f2e1caf190c6530a8d435fc788b69954078ab937db9b5540d9567'
+> prefix = hash.slice(0,8) // 8 nibbles = 4 bytes
```
barnjamin, line 55: is this nibbles or just the stringified hex encoding?

barnjamin:
I like it, couple changes maybe but seems good to me

ARCs/arc-0028.md
DanBurton: For this particular JS library, it returns the stringified hex encoding, where each character represents 1 nibble, which is why `.slice(0,8)` represents 8 nibbles.

ARCs/arc-0028.md
DanBurton: I'm not sure what you mean when you say "by default". Do you simply mean that this example should use a lower-case name? Or that something more should be written into the ARC specifically recommending lower-cased names?

Note that HumbleSwap v2 contracts conform to this ARC, and its event names are upper-cased, which inspired this example.

barnjamin: fair enough, for some reason i had in my head it was lowercase names only but A-Za-z are allowed 👍

ARCs/arc-0028.md
```diff
@@ -0,0 +1,81 @@
...
+
+In short: to emit an Event, a dapp calls `log` with ABI formatting of the log data, and a 4-byte prefix to indicate which Event it is.
+
+## Specification
+
+Each kind of Event emitted by a given dapp has a unique 4-byte identifier. This identifier is derived from its name and the structure of its contents, like so:
+
+### Event Signature
+
+An Event Signature is a utf8 string, comprised of: the name of the event, followed by an open paren, followed by the comma-separated names of the data types contained in the event, followed by a close paren. This follows naming conventions similar to ABI signatures, but does not include the return type.
```
jannotti, line 29: Do you want to explicitly say that the data types allowed are the same as ARC-4? Perhaps explicitly _disallowing_ the reference types like `asset` and `txn`?

jasonpaulos (commented):
Events are certainly a good idea. This is something we've thought about but it didn't make it into ARC-4.

The spec in this PR makes sense to me, but I think it'd be nice to see the following as well:
* Explicitly stating that methods which have a return value **MUST NOT** emit an event after they log their return value. This is [already specified in ARC-4](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#standard-format) but I think it's good to call out here too.
* Extending one/multiple of the ARC-4 [method](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#method-description)/[interface](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#interface-description)/[contract](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#contract-description) description objects to contain a list of events that the code might emit. I'm not sure what makes the most sense yet, but if this information were included somewhere in an ABI description object, I think this would make it easier for contract authors to define and describe events and for callers to identify and parse them.

jeapostrophe:
Re: description objects --- I think that Events are a new category inside of interface or contract, in order to match how people talk about things in Ethereum. Sometimes people write in prose descriptions that a particular method emits an event, but it is not a mechanical thing.

DanBurton:
Here is an example of parsing HumbleSwap's event logs:
https://gist.github.com/DanBurton/d11a6c8a99b33f28ea7fc83a82ccaef8

ARCs/arc-0028.md
```diff
@@ -0,0 +1,81 @@
...
+> This API is a draft.
+> Some elements may change.
+
+## Summary
+
+A methodology for structured logging by Algorand dapps.
+
+## Abstract
+
+Algorand dapps can use the `log` primitive to attach information about an application call. This ARC proposes the concept of Events, which are merely a way in which data contained in these logs may be categorized and structured.
```
robdmoore, line 19: As a suggestion rather than "dapps" should this say smart contracts to make it clearer (on first read I thought it was talking about `console.log` in a web app, which is normally what I think when hearing "dapp").

Also suggest linking `log` to <https://developer.algorand.org/docs/get-details/dapps/avm/teal/opcodes/#log>.

ARCs/arc-0028.md
```diff
@@ -0,0 +1,81 @@
+---
+arc: 28
+title: Algorand Event Log Spec
+status: Draft
+discussions-to: https://github.com/algorandfoundation/ARCs/issues/113
+---
```
SudoWeezy, line 6: Please fill this part as explained in the new [ARC-0](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0000.md)

---
arc: <to be assigned>
title: <The ARC title is a few words, not a complete sentence>
description: <Description is one full (short) sentence>
author: <a comma-separated list of the author's or authors' name + GitHub username (in parenthesis), or name and email (in angle brackets).  Example, FirstName LastName (@GitHubUsername), FirstName LastName <foo@bar.com>, FirstName (@GitHubUsername) and GitHubUsername (@GitHubUsername)>
discussions-to: <URL>
status: Draft
type: <Standards Track, Meta, or Informational>
category (*only required for Standards Track): <Interface, or ARC>
created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
requires (*optional): <ARC number(s)>
---

DanBurton, line 6:
```suggestion
---
arc: <to be assigned>
title: Algorand Event Log Spec
description: <Description is one full (short) sentence>
author: Dan Burton (@DanBurton)
discussions-to: https://github.com/algorandfoundation/ARCs/issues/113
status: Draft
type: Standards Track
category: ARC
created: 2022-07-18
requires: ARC-0004
---
```

DanBurton: 👍 Updated

SudoWeezy:
@DanBurton Please give me access to your PR, I can't make any changes.

SudoWeezy:
@DanBurton Please give me access to your PR, I cannot make any changes.

SudoWeezy:
I can not update the PR, so I created a new one. PR https://github.com/algorandfoundation/ARCs/pull/141

PR #116, opened 2022-08-01
ARC-0019: final call

fabrice102:
The main change is the clarification of the interaction with ARC-3
(there is a need for clients to bypass Asset Metadata validation
in that case.)

Other changes are mostly cosmetic. This includes Markdown fixes
(including text immediately after \`\`\` without newline
breaks markdown)


robdmoore:
Given that there's an `Interactions with ARC-3` section, I'm curious on thoughts on an `Interactions with ARC-69` section.

I know that Patrick isn't a fan of this (we've discussed it before and have differing opinions on this), but in the same way that ARC-20 deliberately doesn't couple itself to any metadata standards I actually think that ARC-19 should be the same. I'm not if that's how the various marketplaces etc. that have supported ARC-19 have implemented it, but I think it makes sense to not assume that presence of ARC-19 implies presence of ARC-3. This gives future flexibility where may there are other metadata standards that get used, but have a mutable URL requirement.

The selfish reason I'd like to see an explicit mention of interaction with ARC-69 is I think the two standards together would be powerful for a sue case like Data History Museum. We deliberately want to store a permanent record of history and so are not comfortable including metadata on IPFS (note: IPFS is immutable, but not permanent) and want to ensure it's on chain. We do have a mutable media requirement though (we use https:// url pointing to cloudflare in front of S3) so having ARC-19 for the media and ARC-69 for the metadata is the perfect fit. Acknowledging that the media would then by my same logic not necessarily be permanent, in our case the media is less important than the metadata. There are other benefits to ARC-69 in some use cases like much easier / performant ability to get historical values (also important for Data History Museum).

robdmoore:
Outside of that, I think this standard is brilliant and made full use of it for https://algoanna.makerx.tech/.

TommyAlex79:
> I think it makes sense to not assume that presence of ARC-19 implies presence of ARC-3.

I agree, arc19 allows you to mutate the effective url and that could point to anything: arc3 json, directly to media file or why not a new arc standard that does not yet exists.

For example, if one is not interested about traits and json, why not pointing to the media file directly and reduce number of in directions? 

Far more importantly, If a new ARCx standard comes out, why not be able to point to a file using that new format. 

I agree arc19 should be far more flexible than just arc3.  Arc19 is that good, it should not be limited to just arc3. 

In terms of NFTs, the two biggest NFT projects are using different approaches. One is pointing to arc3 and then from there to an image. The other is pointing straight to an image in ipfs.

Nftexplorer supports both approaches, and so do major markeplaces.

pbennett:
I prefer it all be to be resolvable solely by the hash (the reserve address) - so just fetching only the asset (from an algorand node endpoint) is all that's necessary - a single fetch.  You then instantly know if its changed - and can cache solely based on that base key.  Whether it's just an image or a link to metadata that then has a link to an image, any change in either would change the base hash - and that's all visible from a single fetch.  Cache the results - whatever they are - image / metadata, etc. based off that resulting CID or hash - and you're done.   If sites want to store histories - you can store cached results by their hashes.

So for ARC19, to "also" use the note technique of arc69 - devs are certainly welcome to do it but I think requiring to *also* search transaction history on every asset is a horrible requirement - and it should always be VERY explicit that this is required.  So if people do this for arc19 assets then people will now have to assume others 'might' do it and now the super clean and super efficient algorand node (non-archival node as well!) asset fetch has to now also be an algorand *indexer* transaction search(!) to check transactions against it ?   Honestly, I would say these aren't arc19 assets.  If I read an ARC19 asset I would never expect to ALSO search the transaction history to grab the latest note - ARC19 is partly to get rid of this requirement!  ARC19 only requires a single asset fetch from a node and nothing else - very explicitly and deliberately.

I think if people really want to mix the two it should be different ARC and be labeled as such - possibly with a #arcXX fragment on the URI.
So a new ARCXX could just be notes describing the use case for the combining of the two - and then we can figure out a way to tag them so implementers know they'll need to search transaction history using an indexer as well.

ARCs/arc-0019.md
```diff
@@ -42,61 +44,107 @@ An Asset conforming this specification **MUST** have:
...
 
 The parameters to the template ipfscid are:
-1. IPFS _version_, **MUST** be '0' or '1' representing the IPFS CID version.
-2. _Multicodec content-type name_, **MUST** be an IPFS multicodec name of 'raw' or 'dag-pb'.  Other codecs **SHOULD** be supported but are beyond the scope of this proposal.
-3. _Field name_ **MUST** be 'reserve' to represent the reserve address is used for the 32-byte hash but is specified here so future iterations of the specification may allow other fields or syntaxes to reference other mutable field types.
-4. _Hash type_ **MUST** be 'sha2-256' but is explicitly specified so as to allow future versions the option of using other hashes. 
+1. IPFS `<version>`, **MUST** a valid IPFS CID version. Client implementation **MUST** support '0' or '1' and **SHOULD** support future version.
```
pbennett, line 67: ..must 'be' a valid.. - be was removed accidentally here.
All other changes are excellent, thanks Fabrice!

robdmoore:
That makes sense @pbennett  - I agree it should be clear what metadata standard is being used in conjunction. That's why I've been using `#arc3` on my arc-19 NFTs. ARC-69 has this with the `#i` / `#v` etc. already baked in.

pbennett:
@robdmoore 
My point is that using ARC19 *and* ARC69 is not an ARC19 asset and shouldn't be considered one.  An ARC19 assets current state should be be able to be fetched entirely from a single asset information fetch from an Algorand node.
ARC69 requires a separate transaction search using an indexer.
If you want to *also* use ARC69 on an ARC19 asset, then that combination should be a different ARC and that asset then be ARCXX instead.   Then that ARCXX asset could be #arcXX and integrators would know it was an ARC19+ARC69 hybrid.

fabrice102:
@robdmoore @pbennett The main reason why we added the ARC-3 section is that ARC-19 is actually formally incompatible with ARC-3 because of the asset metadata hash.
However, ARC-19 is fully compatible with ARC-69 as ARC-69 does have the asset hash optional.
I think we can wait and see how things go. We can always add non-normative comments or create a new standard depending on use of ARC-19+ARC-69.

I'm merging this and will mark the full spec final in a couple of weeks.

PR #117, opened 2022-08-06
ARC-0029: Soul Bound NFTs

Hathoriel:
Adding a line of the ARC to provide context

> A "Soul Bound NFT" is an Algorand Standard Asset (ASA) controlled by a Smart Contract
> that exposes methods to create, configure and destroy the
> asset.

This PR is issued in the context of the [Hackathon: Algorand GreenHouse Hack#1 Soul Bound NFTs ARC](https://gitcoin.co/issue/29133)

### DIFF
--- a/ARCs/arc-0029.md
+++ b/ARCs/arc-0029.md
@@ -1,0 +1,102 @@
+---
+arc: 0029
+title: Algorand Standard Asset Parameters Conventions for Soul Bound NFTs
+author:
+- Lukas Kotol (@hathoriel)
+  discussions-to: https://github.com/algorandfoundation/ARCs/pull/117
+  status: Draft
+  created: 2022-08-06
+---
+
+## Summary
+
+This document introduces conventions for the parameters of Algorand Standard Assets (ASAs) for Soul Bound non-fungible tokens (NFTs).
+
+## Abstract
+
+A "Soul Bound NFT" is an Algorand Standard Asset (ASA) defined by configuration of parameters preventing
+them to be transferred from an original address.
+
+## Motivation
+Non-fungible tokens (NFTs) are digital tokens on the blockchain network.
+Each NFT has its unique token ID. This is just a number that identifies the token,
+and no two NFTs minted from the same smart contract can have the same token ID.
+This proves that every NFT is unique. NFTs can be traded or given away, which
+means that they are not irreversibly tied to only one entity.
+
+But imagine one who wants after completing a challenge to get honour or
+status which is bounded to the person and certifying such a person has finished the challenge.
+This scenario can be mapped to real-life achievements like obtaining a university degree,
+winning a club competition or earning a professional certification.
+
+Described use cases are also widely used in gaming.
+In some MMORPG games, after slaying rare monsters, game characters get legendary items
+which are bounded only to the hero who took part in monster destruction.
+These legendary items are characterized by the fact that they cannot be sold
+or traded with other players, which means they are "Soul Bounded" to the game character.
+The consequence of binding items to the character is that other players could be sure their
+owner had slain such a monster and it can give the slayer special status among other players.
+Once players dont want to keep "Soul Bounded" items, they can trash them to permanently remove them from their account.
+
+However, the ASA lacks the possibility of proving that the owner of the address finished some task, mission or job,
+because the ASA can be transferred between addresses.
+
+The Soul Bound NFTs defined by this ARC prevent ASA to be transferred and permanently binding NFTs to the owner's address.
+This means that not only user cannot transfer ownership, but the minter also cannot withdraw/transfer/change ownership as well.
+By introducing this as a standard users can confidently and consistently prove or certify
+that a project, competition or event has been completed by themselves.
+
+## Specification
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC
+2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+### ASA Parameters Conventions
+
+The ASA parameters of Soul Bound NFTs should follow the conventions for NFT tokens defined in [ARC-0003](https://arc.algorand.foundation/ARCs/arc-0003)
+except for the following parameters:
+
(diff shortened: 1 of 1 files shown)


cusma:
Hi @Hathoriel, isn't this just a special case os ARC-20?

Hathoriel:
Hi, I would say yes. Do you propose any changes how to explain Soul Bound NFTs in more elegant way? If you will head me to any structure which would explain it better it would be great.

cusma:
I would say that this use case could be completely derived as a specific application of ARC-20.

Hathoriel:
And do you have any example of already derived ARC from some other ARC for me to take inspiration? 😊

aldur:
> @cusma it is special case of ARC-20, but for instance interface for creating asset or getting config is different.

Why does it have to be different?

Hathoriel:
> @cusma it is special case of ARC-20, but for instance interface for creating asset or getting config is different.
> 
> Why does it have to be different?
> 
Because for instance for Soul Bound NFTs it does not makes sense to have params [freeze_addr or clawback_addr](https://github.com/algorandfoundation/ARCs/pull/91/files#diff-c08e64a43dec1f1b41e618fabc47655cc5997c13292140d6738324d006f643ffR88) when NFT is irreversibly bounded to the address. It does not make sense to keep this information about freezing when you have it frozen permanently.

aldur:
> Because for instance for Soul Bound NFTs it does not makes sense to have params [freeze_addr or clawback_addr](https://github.com/algorandfoundation/ARCs/pull/91/files#diff-c08e64a43dec1f1b41e618fabc47655cc5997c13292140d6738324d006f643ffR88) when NFT is irreversibly bounded to the address. It does not make sense to keep this information about freezing when you have it frozen permanently.

One way to do that without having a new spec (which I think is a BAD idea in this case and agree with @cusma this is a specific application) is to mandate it in the implementation: any invalid call (e.g. specifying `clawback_addr` other than `ZeroAddr`) will be rejected.

robdmoore:
A soulbound NFT shouldn't even need a smart contract necessarily yeah?

You could create the NFT as default frozen with a clawback, issue it to the user (after opt-in) using clawback and then clear clawback. At that point it is bound to that address forever without needing a smart contract at all?

The one caveat is if you want to allow the asset to be destroyed. In that case you would need a smart contract.

I feel like both options should be mentioned in any ARC standard discussing soulbound tokens.

Hathoriel:
> > Because for instance for Soul Bound NFTs it does not makes sense to have params [freeze_addr or clawback_addr](https://github.com/algorandfoundation/ARCs/pull/91/files#diff-c08e64a43dec1f1b41e618fabc47655cc5997c13292140d6738324d006f643ffR88) when NFT is irreversibly bounded to the address. It does not make sense to keep this information about freezing when you have it frozen permanently.
> 
> How about this?
> 
> Instead of having a new spec, enforce this in the implementation but maintain ARC-20: any invalid call (e.g. a configuration specifying `clawback_addr` other than `ZeroAddr`) will be rejected.
> 
> I think is a BAD idea to provide a new spec for this and agree with @cusma this is a specific application of ARC-20.

@aldur Ok thanks for your idea. Will try to make it close as possible.

Hathoriel:
> > A soulbound NFT shouldn't even need a smart contract necessarily yeah?
> > You could create the NFT as default frozen with a clawback, issue it to the user (after opt-in) using clawback and then clear clawback. At that point it is bound to that address forever without needing a smart contract at all?
> > The one caveat is if you want to allow the asset to be destroyed. In that case you would need a smart contract.
> > I feel like both options should be mentioned in any ARC standard discussing soulbound tokens.
> 
> Careful @robdmoore and @Hathoriel - users **can** still move frozen assets - they can close the asset out - which sends it back to its creator. Even when its frozen into their account.
> 
> NFDs are basically soul-bound NFTs in almost all ways - so I'm already doing what ARC20 covers and the 'user can close' the asset, even when frozen, is a non-obvious gotcha that has to be handled. As a result, if you truly want it covered, a contract needs to govern it. (because the contract can put it back later for eg).
> 
> I do agree with other posters though and what's being described here seems basically to be a particular _implementation_ of what ARC-20 already covers.

@pbennett Thanks for noticing this, I will try to include it into the solution. 👍

Hathoriel:
Hi, @robdmoore @pbennett @aldur, I have changed the whole ARC according to your comments, so hopefully, we will continue on the work. :-)

ARCs/arc-0029.md
```diff
@@ -0,0 +1,101 @@
...
+Upon voluntary pickup and craft, the sword permanently binds to a character’s “soul,” making
+it impossible to trade, sell or even swap it between a player’s characters.
+
+In other words, “Thunderfury”’s price was the aggregate of all social costs
+related to completing the difficult quest line with friends and guild members.
+Other players spotting Thunderfuries could be sure
+their owner had slain “Ragnaros,” the blistering firelord.
+
+World of Warcraft players could trash legendary and soulbound items like the Thunderfury to permanently remove
+them from their account.
```
fionnachan, line 38: I understand this is a submission for the gitcoin bounty https://gitcoin.co/issue/29133 and so it uses the term Soulbound. This paragraph is trying to explain the terminology "Soulbound", which is needed to give context. However, the highlighted part is a blatant copypaste from [EIP-4973](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4973.md#motivation) to this proposed ARC. I'm sure the concept could be explained with other examples and wordings instead of a direct copy of words.

Hathoriel: OK, thanks for the review! I will try to reformulate it better.

Hathoriel: Hi @fionnachan, I think I have fixed this issue with new wording and examples.

fabrice102: Thanks @fionnachan for flagging this!
Since EIP are CC0, in theory, I guess we could use directly without citation.
However, I would strongly recommend that EIP-4973 is credited if anything is taken from there (even actually if not strictly copy-pasted).

SudoWeezy:
Hello,

We will ask you not to submit a PR directly on the ARC repository to ease the process during an ARC bounty hackathon.
After judging if the winner is deemed to be a good starting point for an ARC, a PR can be made.

Please submit your work directly here:   https://gitcoin.co/issue/29133

If you have any questions, please contact joe@algorand.foundation

PR #122, opened 2022-08-16
ARC-0030 : Collection Metatdata as an ASA

whereisrysmind:
NFT collections are currently only defined off chain, and many sites have their own index of collections that require independent verification. This standard allows anyone to define an NFT collection with optional verification. Anyone can assemble their own collection list from data on chain without relying on an intermediary or gatekeeper, representing a common reference. Creators can mint a cASA to group their own creations, and any user can mint a collection from multiple creators for arbitrary categorization. An ASA can be part of multiple collections or none at all. Using ARC19 as the base format allows cASAs to be mutable: new assets can be added to or removed from the list dynamically. A consistent unit name ('CLXN') allows for easy asset search by unit name. Allowing collections of applications or addresses provides further flexibility and new functionality. Using this specification applies no additional requirements to the creation of assets, or the existing standards for doing so, which allows for backwards compatibility with existing minted items.


pbennett:
This seems like a really nice idea actually.  ARC19 allows it to mutate as what is/isn't in a collection can change over time (or be 'revealed' later).  The asset will always be in the creators control, and it becomes easy to locate within a particular set of creator accounts.

For NFDs at least, I was thinking creators could just set a 'collection' property and that would be the collection name used for all NFTs linked to the NFD linked to their creation accounts.  However, for multiple collections created in the same accounts (not the best IMO, but it's not uncommon), this 'account(s)->NFD->collection' model breaks down.
Segments in the future could be used for independent collections for particular projects but again, it still kind of presumes distinct accounts per collection.

IOW: I'm definitely in favor of this one !   I'll add support to NFDomains NFT tracker right away if this catches on.

runvnc:
Normally the collection is defined by the wallet address so listing all of the assetids seems like it's just an alternative for people that don't want to query an Indexer. Also this requires an extra two steps when collections change such as new ones being minted: pin the new collection ASA and then the asset config transaction.

But I have discussed before with one of the NFT Explorer guys about this basic idea of having some decentralized collection description. When I brought it up it was really because many people were putting multiple wallets together in a collection. Now with unlimited ASAs per wallet it's less of a big deal.

But still, this type of thing could be useful for verification. Take for example the NFT explorer verification process. How would (or could if expanded) this ASA be used to replace that verification process? I mean let's break down what we are supposed to be verifying and see how this ASA can accomplish that.

I feel like we should also consider a way to take advantage of NFDs. Maybe somehow share a signature or some proof with the NFD? Not sure.

gabrielkuettel:
> How would (or could if expanded) this ASA be used to replace that verification process? I mean let's break down what we are supposed to be verifying and see how this ASA can accomplish that.

Maybe an `addresses` property would be useful, which includes the address(es) that minted the collection. If not present, it could default to the token's manager address. I believe a lot of marketplaces / explorers verify based on algorand address, so this would provide a convenient way for them to verify a collection's addresses by looking at the CLXN token.

runvnc:
Maybe just make sure the asset config for this ASA is issued by the "real" owner's address that has an NFD. Then you would just give the assetid of this ASA to NFT Explorer or whatever and they can display the owner info for that NFD. If that NFD/wallet has already undergone some verification process by NFT Explorer or the marketplace then it will not need to for subsequent collections.

Also please add a field that just states that everything in the one wallet is in the collection. Unless is it really critical to have every single asset listed when it's just the one wallet?

pbennett:
It seems like this proposal definitely comes from the perspective of an NFT marketplace.  They want to show NFTs in terms of 'collections.'  There's nothing in the NFTs that defines this, and there's nothing about the creators accounts that really defines this either, as creators will sometimes use the same account for multiple collections.
Since marketplaces/aggregators (asalytic/nftx/..) don't want to be in the business of having to maintain various internal private databases of which specific assets are, or aren't in a particular collection - putting that burden on the creator themselves seems reasonable.

On one hand, it's a big 'heavy' to have this extra file that has to be updated as assets are added, but given that there are a LOT of NFTs that can't be changed in any way, there has to be some way of grouping those legacy assets (possibly comingled) all together, but in a way the creators still have control over.  The new 'asset' way of doing this is a possible mechanism I think.

In regard to NFDs (since others mentioned them again) - NFDs are super powerful in that a single NFD can be verifiably linked to specific accounts owned by that NFD - and thus all NFTs 'created' by those accounts can link back to that creator's NFD, and properties in that single NFD can apply to all of those creations instantly.  It was always my intention that the 'creator' of an NFT being able to be tied back to an NFD and that be a verification model, but collections gets interesting.
A 'collection' property for eg would handle that like I mentioned - however - I've gotten the impression that it's not uncommon for creators to 'mix' collections a bit, creating multiple collections from the same account.  
The community could say, hey, don't do that and some of those existing collections get a little comingled, being labeled as one account?  So, then the guidance was 1 NFD per collection - then it would be simple.

However, because they are a perfect place for on-chain metadata location, there's no reason a property in NFDs couldn't just be an IPFS link itself.  Let it be kind of the meta CLXN(s) asset ? ie: a property called collections that is an ipfs link (updatable by the owner at will with a single update_field NFD contract call) that lists the multiple collections in the linked accounts and the asset breakdown?

Hopefully I'm not diverging too much from this specific ARC - but it does feel like we're all still trying to feel around a little bit on how best to handle it, and a little spitballing seems in order.

Vilijan:
Really interesting proposal. I have a couple of questions regarding it.

### What is stopping malicious actors to add assets that are not part of their collection in their cASAs?

Let's say we have two collections created by different authors, A and B. Is something stopping B from adding ASA_IDs in the `assets` property in their cASA that belongs to collection A?

With this, each cASA can potentially hold invalid data which will need to be validated by the platforms that want to aggregate the collections such as Asalytic & NFTExplorer.

### Can the same result be achieved with ARC69?

ARC69 enables you to add configuration to each asset using AssetConfigTransaction. If an optional `collection_name` property is reserved in the ARC69 for the name of the collection, then for each ASA we will know to which collection it belongs. Additionally, there will be no room for manipulation, since only the `creator_address` can issue those AssetConfigTransactions.

TommyAlex79:
I think this approach could work well for **personal favourite lists**. I could create a list of all NFTs that I like, use that collection NFT to then view my favourite items somewhere. These can be from any collection, owned or not. Beyond this, I have **some doubts** (which can be addressed probably, but which are currently not addressed in the current PR as of today).

Beyond "personal fav lists", I see this as a starting point for discussions. I am not sure how a website like asalytic or nftx could use this with any degree of safety, I am also not sure how a marketplace that has not issued this **would be able to trust it**, as is. Would we basically end up ignoring all cASAs unless issued by a trusted source? So, ignoring the ones issues by artists and individuals, and by marketplaces and shuffle sites which might not yet have earned trust? This **could end up in the vast majority of cASAs being ignored by most parties**? (unless we address this issue somehow). Also to trust some parties, we need to know who issued it and what sort of checks were performed. Having the issuer as a string in the json would not be sufficient, that could be forged, it would either need to be minted by that issuer or contain some sort of signature of that issuer as proof.

I also share @Vilijan concern that **there is currently nothing stopping someone adding (even at a later time) assets that are not part of their collection** to the increase perceived quality of the artwork in general and attract more buyers (also for their own items). I am sure there are multiple other potential exploits. 

Also, for large collections which support minting on demand, this would potentially be **very wasteful of resources**... at 10k items, minting a new one on demand would result in 10,001 files having been pinned so far. The last file would be again a brand new immutable JSON file with 10k + 1 items, where the only difference would be the addition of the last element (a popular collection could also be doing this at a fast pace). 

While the PR says the unit-name "CLXN" must be used, there is nothing preventing any ASA that is not a cASA from using "CLXN" unit-name (i.e. **it's not a special reserved unit-name**). We would need to be conscious of that and introduce something else in the JSON to make it clear what we are dealing with.

Also, the PR seems to be suggesting a general approach, rather than something standardise that everyone knows how to interpret. Custom fields can always be added, sure, but **main ones need to be defined as part of the standard, and with their exact format and data type**, etc....  For example a PFP image for the collection is clearly missing. Details about the artist is missing also, but, for example, should this be a string? Or a list of strings? (multiple artists) Should it be a complex type which can include website links? or just a reference that points to a aASA (Artist ASA) or maybe an NFD? 

I think now that there is no upper-limit on the number of NFTs that can get created, a simple approach would be to use and dedicate an algorand account to a single collection. If someone wants to give it a nice name and some properties, could then use NFDs, for example. **This what most artists seem to be doing now**. While very raw, at least this does not permit malicious asset injection: all assets created by this account will be by definition created by this account and the blockchain guarantees that.

I am of course in favour of finding a solution that can be more flexible than the raw approach above, and would also work for some legacy collections (which are over multiple accounts or where the same account is reused for multiple collections), but I think a bit more thought and work is required or if we rush something through it will be ignored by most aggregator and marketplaces.

For me, the most important thing is that we are discussing this, I totally agree with the need to try and find a more decentralised approach, whether it is this or not. 

Current version at time of this message: https://github.com/algorandfoundation/ARCs/blob/f2ce1853c477b76d43e3ff34da3ff5061a64c349/ARCs/arc-0030.md

runvnc:
I agree with a lot of those points but I still see the verification of owner as relevant to collection metadata and NFD has verification built in so it can serve as a decentralized ID. It would be nice if there was a way to just not necessarily have to put everything inside of the actual NFD. Some things might be more loosely connected (like a collection ASA sharing the same wallet address).
Guess I am kind of repeating myself.

pbennett:
> What is stopping malicious actors to add assets that are not part of their collection in their cASAs?

I think it's kind of assumed this wouldn't be allowed but it should still be clearly spelled out as a MUST.  Any marketplace will already have all assets created by the creator from one ALGOD fetch.  Only those asset ids should be valid for anything in CLXN assets found by that same creator in the same account.  An asset id not created by the account the CLXN asset was created by should be excluded.

> Can the same result be achieved with ARC69?

It would at least let 'legacy' assets get a collection name added, but personally I just want to see the 'fetch asset info, oh, and ALSO search all acfg transactions to 'possibly' get the metadata' approach die.  You can get all the information you need to fetch initially (or the prior cached results) for arc3 and arc19 by just one 'asset' fetch of an algod node [no indexer required].  It also means having to 'touch' every single NFT instead of doing it all in one place.

evanmaltz:
Agree with @pbennett, this approach is more efficient and cleaner than using ARC69 because fewer txns/fetches are required and does not sacrifice security. Anyone can verify if the cASA creator also created the assets it lists, and there are still use cases for cASAs with assets created by others (e.g. favorite lists) which cannot be achieved by ARC69. Other ARC30 functionalities that cannot be implemented via ARC69 are lists of addresses. 

Another concern was wasting resources... I don't think anyone minting a 10k collection would mint them one at a time and pin a new collection each time. They would just update the cASA in batches. Also pinned files can be unpinned if it really becomes an issue.

True that anyone can use the "CLXN" unit name for any purpose, people can also mint a cASA that is a list of ARC30 compliant collections so that solves itself. There will always be a problem of non-compliant implementations (purposeful or not) cluttering the chain. 

I will include an addresses field to include multiple creators. And because the cASA has 10^19 units, every creator's approval can be verified by distributing units of the cASA to all the creators in the list. This allows for safe verification that all assets in the list have been approved by their verified creators.

Also @pbennett, a custom collections field would work, but constrains the number and types of collections that people might make. I can see people buying an NFD to represent proper NFT collections, but not favorite lists for example. I think it would be better to have a collection primitive that can use NFDs for verification if desired, people will get cost limited if they have to mint a new NFD for each collection. Plus asset IDs should consume significantly less space on the NFD than the ipfs hash (right? 32 bytes vs 30 bits for asset ids <1 billion in decimal if I'm not missing something), so users could fit many more cASAs on one NFD. I would like to brainstorm how cASAs can better interoperate with NFDs, so far it seems like the best utility is verification.

runvnc:
I might be missing something else but if we are just literally going by a
big list of assetids we could end up multiplying that times 1000 again if
they can be coming from random wallets and we have to look up each one.

evanmaltz:
@runvnc I think the lookup could be done efficiently by getting all the assets from the creator address then comparing the lists, rather than looking up the creator of each asset independently. But even so, that verification would only have to be done once.

pbennett:
@evanmaltz 
I'm confused about the favorite list mention.  Favorite lists sounds like more of a "per-user" preference, how does that fit into the definition of a creator-controlled list of assets in a 'collection' ?

Re. the NFD and multiple collections - it's easy to point to off-chain but decentralized metadata (IPFS) that's fairly large - so a top-level 'creator' account could reference all of their collections via one property.  A collections property linking to IPFS content that references all their collections - arguably even an IPFS unixfs tree - where each collection was a distinct json virtual 'file' within the virtual directory tree.

Another option, and it comes later, but NFD "segments" will also likely be a cheaper way for someone to have NFDs per-collection.  Root level creator name with collection as segment - and then data in each standalone nfd for that collection.
I think there's a lot of options here.  One part of me wants to come up with some ideas and just run with it, but another wants me to see what the community (and this is part of that IMO) comes up with and innovates on independently.

Because NFDs allow owned-controlled linkage and that is something that can be used for NFT->creator->NFD associations there's already a lot that can be done there.  The fact that a CLXN asset is in a linked account still automatically kind of falls into the same bucket IMO - NFTs / Collection Definitions - owned by X account(s) - tied to x NFD with same owner.  IOW: We have a lot of options here.

evanmaltz:
@pbennett favorite lists would be outside the scope of creator-controlled collections. 
Would you be able to handle verification multi-creator collections? I liked the idea of using the ASA for that where each creator could verify their contribution to a collection by owning a unit of the cASA.

pbennett:
I'm not sure the multi-creator approach really matches though.   A given NFT can only have one creator.   If that's truly multi-creator and truly multi-ownership, than that creation account could be multisig but it would still be a single account.
So if the cASA was distributed across multiple accounts, there's still only one manager of that cASA that would have to modify it.  So then what does it being in another account really represent?  
Do you see this more for things like collabs where a single-owner cASA (only 1 project) is responsible for the collection and references asset-ids created by a completely different account, yet the presence of a cASA with an asset-id contained in that other account is what allows that other account and asset to be considered 'valid' for that collection?

If this is the use-case you're thinking then I could see it.

TommyAlex79:
> And because the cASA has 10^19 units, every creator's approval can be verified by distributing units of the cASA to all the creators in the list. 

**I cannot see any reference to this in the ARC PR**. 
Is this how they must be minted? I can see this is how the current implementation mints them though, which seemed peculiar.

In any case, it starts getting fairly complex. Are you proposing that on top of knowing who the creator of the cASA is, one should also know all the holders of the cASA and should verify the list of assets against either the creator or the holders?


Also, the tweet (since there seems to be more detail in the tweet than in this ARC) mentions: 

> Asset lists can include other ASAs, smart contracts, wallet addresses, or even other collections.
[https://twitter.com/KinnDAO/status/1559639199971778561]

**Again, this is not in the ARC**, the ARC is a lot more vague about this:
> "`assets` property **MUST** contain an array with at least one index"
So, it's not clear if the ARC matches the tweet (index can be any index) or if the index must be an ASA ID.

So, if the ID does not seem to have been created by the account creating the cASA or by one of the <=10^19 accounts holding the cASA, then it might still be valid as it could be an application ID instead? 
Or is this not something that is really intended to be supported? I think the ARC needs to be very clear about this.


> But even so, that verification would only have to be done once.

not once per cASA, but once per modification though.


> I'm confused about the favorite list mention. 

The way this has been introduced to the community indicated that this can be used for that purpose also (and in fact the current implementation allows that). This would mean failing "verification" would not be the exception, but the norm potentially. 


Version discussed:
[https://github.com/algorandfoundation/ARCs/blob/f2ce1853c477b76d43e3ff34da3ff5061a64c349/ARCs/arc-0030.md]

evanmaltz:
@TommyAlex79 The ARC does mention in a couple locations what counts as an index: "Indices can be application IDs, ASA IDs, or addresses", and also in the examples. I will include a more explicit definition in the requirements section as well.

I will also include the number of units in the requirements to support the ability to handle verification of multi-creator collections. That idea is new. And yes, I am proposing the option to do one additional query to check the holders of the cASA and then compare that to the list of creators, should not add much overhead. Anyone can perform this verification check and publish their list of verified assets for use as a reference in case some don't want to do all the checks themselves (Kinn Foundation will do this).

> not once per cASA, but once per modification though.
Not once per modification. At best only once per address list update, and at worst once per asset list update.

TommyAlex79:
Ah, it did in the abstract, you are right. 
Should it be a single list like that? Or would it be better to have 3 lists? What is the expected meaning when adding an address? 

Yes, my concerns with regards to verification is obviously not the individual or the marketplace minting, but another entity watching the blockchain for new cASAs and having to do their own independent verification, since anyone can mint them (another marketplace or aggregator). The concerns are outlined above in the other posts. 

I can see how minting them for own use would be a lot less problematic, you trust yourself, whether you are an individual, a marketplace or anything else. This is **not** the perspective from which I am looking at this. 


>  And yes, I am proposing the option to do one additional query to check the holders of the cASA and then compare that to the list of creators
>Not once per modification. Only once per asset list update. Once verified, you would only need to check the assets that were added instead of the entire list so it can still be done as efficiently as if the new assets were in their own collection.

**The asset creator is immutable, but holders can change. So actually, you would probably need to re-verify at any asset transfer of any quantity of the cASA too.** 


> This is an issue with ARC19 in general and the community doesn't seem to mind.

No, this is a completely different use case. It's a **feature** of ARC19, and it could be an **issue here**.

evanmaltz:
@TommyAlex79
It may make sense to have 3 lists, not opposed to the idea. There should be a creators field for an array of addresses, distinct from using the asset list for addresses. The asset list of addresses could potentially be useful for rankings or other forms of information about groups of addresses. from just holding a unit of the cASA which would indicate permission to use derivative works. 

> This would mean failing "verification" would not be the exception, but the norm potentially.

There can be a field added that indicates the cASA creator is intending for this to be an official asset collection as opposed to some other use case. This will minimize unnecessary attempts to verify of assets that do not require it. Default will be false.

Default frozen can be used to prevent unauthorized cASA unit transfers. But in theory there is no issue with new holders of the cASA units, checks would only need to occur when a holder returns their unit (as this would indicate that they have revoked their permission). So it is possible that a holder may not be the creator of an asset in the collection, because that asset may simply be a derivative work of theirs. It would be up to the collection owner to ask for permission to use IP by exchanging units and the IP holder to monitor usage of their IP. If no unit holders are listed as a creator of any assets in the cASA, then checking unit holders isn't necessary for verification. 

Assuming absolutely no trust, verification would only require independent entities to:
1. Check that the collection is intended to be verified (i.e. 'official': true), if false verification is not required.
2. Check the asset list and assemble a list of creators. Though ideally it will be correct in the creatorAddresses field metadata. I suppose if the lists are found not to match, verification could fail right there. This step could be optimized to simply check if the assets list is a subset of the cASA creator's created assets (which I imagine will be the norm), in which case no subsequent steps are required.
3. Check that all creators are holding >0 units of the cASA. No other addresses really need to be checked because it will be up to the cASA creator to manage the distribution of the units. 

It is only a handful of requests. And at the end of the day, there are <2000 collections on Algorand so I don't think this verification will be prohibitively difficult. The vast majority of collections have only a single creator, so these considerations are worst case. Most of these collections won't be changing regularly and only the updates will need to be checked, which can be done efficiently by comparing asset lists of the different versions and also watching for specific asset transfer txns of cASA units from holder --> cASA creator. Even in the completely trustless case, I think this is pretty reasonable.

> No, this is a completely different use case. It's a feature of ARC19, and it could be an issue here.

If someone mints an original NFT as ARC19 then updates the image to copy someone else's work (or deletes the traits, or any other change a holder may take issue with), it would be the same issue for the holder. ARC19 (and really just mutability in general) is the source of the issue you are describing. The problem you describe could be solved by using ARC3 or deleting/changing the manager address.

ARCs/arc-0030.md
```diff
@@ -0,0 +1,71 @@
...
+
+This ARC introduces a standard for defining a collection ASA (cASA) linked to ARC19 metadata. This standard defines a collection as a mutable set of indices, providing flexibility. By linking to ARC19 metadata on IPFS, the set can be arbitrarily large, even including the complete set of all indices. Indices can be application IDs, ASA IDs, or addresses. 
+
+## Specification
+
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+> Comments like this are non-normative.
+
+To be considered a collection, the cASA **MUST** use the unit name `CLXN` and **MUST** contain a property `assets`. The `assets` property **MUST** contain an array with at least one index. A collection name and description are **RECOMMENDED**. All other fields are **OPTIONAL** and additional, custom fields **MAY** be included. The cASA **SHOULD** be minted using ARC19 to allow for immutability, but **MAY** be minted in the ARC3 format if immutability is preferred.
```
SudoWeezy, line 29: It might be helpful to add `"url"` as **RECOMMENDED**. 
Maybe a `"social_media": {}` field can make sense to.  
What do you think?

xxiled-plastic-cat:
From a platform point of view - the proposal is really good - we currently do have a self seeding database of NFT collections based partly on the name of NFTs and the creator address(s). Allowing us to query to BC and have one asset with a list would be a great help - then if we want to hold things in our DB for easier/quicker access then we can, but we don't need to - would hopefully open up the space for more innovation.

From a user standpoint - either through NFDs or somewhere else - creator lists with your own cASAs that can be minted on demand would be really nice - can see a lot of applications for that - not least if that cASA were in a way sellable - therefor creating a bundle of ASAs in the list which could be sold in one go. Very interesting and looking forward to watching it progress.

headline-design:
> since only the `creator_address`

This is a very good point. We would propose requiring individual assets to also have a collection id for security purposes. So collection objects are invalidated if they contain assets that are not self-identifying as part of the collection.

pbennett:
> This is a very good point. We would propose requiring individual assets to also have a collection id for security purposes. So collection objects are invalidated if they contain assets that are not self-identifying as part of the collection.

Having to update every single asset to change what collection they're in is an issue, and either forces the asset to only be in a single collection (or the asset would need a collection array).  Most importantly, requiring any kind of mutation *on every single asset* kind of goes against the entire point of this proposal (ignoring legacy or truly immutable assets).  
Also requiring indexer searches just to get 'current state' is a horrible precedent to require - particularly since many assets purposely chose not to for a reason.

runvnc:
1. Why does the artwork need to go in the verification section?
2. Can we consider requiring (MUST) some kind of image, name and description?
3. Please recruit a few developers such as from NFT Explorer or whatever marketplace and go through the verification workflow with them using this ASA. Maybe the NFD plays into that somehow?
4. There needs (MUST) to be a list of associated wallets. Otherwise you need to query each asset individually, or query one asset and hope that if you query that wallet you will get all of them. Besides the associated wallets are important information. The wallet and unit names may need to be connected also.

runvnc:
Also this is probably highly relevant to what we are trying to do here, if you take a step back: https://docs.ipfs.tech/concepts/file-systems/#mutable-file-system-mfs

For example if the hash of a directory was the same then you would know that you didn't need to update any cached asset data for a collection.

runvnc:
After discussing this with another developer, the verification seems useless, and validation seems unworkable and will cause implementation issues. Also as currently implemented the deployed collections are generally invalid.
Also it is quite strange that the name is optional. I don't see how a collection could practically be useful to anyone if it doesn't have a name.

TommyAlex79:
Anyone can mint anything they want obviously, but here we are proposing a standard for the whole community to adopt and as such we need to make sure all devs, creators, aggregators, marketplaces, etc.. are happy and fully satisfied by this approach. Is this the case? I am not sure we can say this. @SudoWeezy, what is the process to try and understand the sentiment?

Ideally, this should have been discussed in an interactive forum before the PR.
It's very hard to **shape** something when it's already a PR with an ARC number, and already being minted (despite no adoption, other than: "[We’re giving away 100 algo to the address that mints the most collections](https://tinyurl.com/2xsku8cx)").

Anyway, I am still worried by how complex it is for people to be able to use and trust a cASA. The validation is complex, but also you need to watch reserve address changes live and asset transfers live and immediately invalidate and revalidate.
Holding a cASA is also too weak, the manager of the cASA signs any changes, but holders are unaware of changes, they are not asked to sign off anything. Conversely, any holder (collab, etc..) could cause problems by sending the cASA back to creator and invalidating the whole collection. 

I could point out small things that are not clear in the spec, or small suggestions, etc... but somehow I feel that's a distraction as I would like to go back a few steps and try and simplify this and leverage more blockchain guarantees.

SudoWeezy:
@TommyAlex79 I agree with you that this kind of new concept needs to be approved by the vast majority of the community before approval.
I want a proper way to bring every party involved in the same meeting to discuss ARCs. In that case, mainly explorer, marketplace, and significant project.

SudoWeezy:
This arc will be discussed the 21/09/2022 in this [discord channel](https://discord.com/channels/491256308461207573/1021420568341397504) 
- 1pm NY
- 5pm UTC
- 7pm PARIS

runvnc:
Well, kind of like being stuck in a time loop and don't want to rehash any
old arguments, but again, just pointing out, unless I am crazy.. with arc69
I can get all of the assetids _and_ metadata from one Indexer query (per
1000 assets) whereas if we just have a random list of assetids from
multiple wallets then that means we may need multiple queries just to get
all of the reserve addresses and then to actually get the metadata it is 1
fetch per asset, or 1000 fetches per 1000 assets.

Anyway Arc19 is well established at this point and working towards support
in my programs.

SudoWeezy:
Box storage will be released soon, as we can use them to store information on-chain. 
We can also consider this approach.

emg110:
I love the idea but can I ask why supporting ARC3 is not tried or rationalized out! Is there a rational reason ARC3 is excluded because ARC3 collections as well can enjoy this proposal IMHO!

PR #123, opened 2022-08-30
ARC4 - adding boxes

barnjamin:
An upcoming release of Algorand allows Applications to create and manage `Boxes`, a named set of bytes owned by the Application that created it.

In order to read or write to a box, it _MUST_ be specified in the ApplicationCallTransaction much like the current requirement for Assets/Apps/Accounts.

This PR introduces boxes as a reference type to be encoded in the same way as other reference types.


JasonWeathersby (commented):
We definitely need to add boxes here.

PR #126, opened 2022-09-05
Arc4

FrankSzendzielarz:


FrankSzendzielarz:
@SudoWeezy PR suggesting some ways of addressing issue https://github.com/algorandfoundation/ARCs/issues/112

barnjamin:
I'd like to suggest we make this into several PRs to isolate the concepts and have more focused discussion.

The ABI type changes should be on its own for example

FrankSzendzielarz:
OK Will separate the PR into 2 parts a) stuff related to types and fields b) the contracts.json related metadata

Also, the proposals around state should be clarified: the state named variables are those 'guaranteed' by the contract across all instances of the contract. Any instance specific or dynamically generated state keys should be accessible via a dictionary in the HLL client. Those instance specific or dynamically generated keys don't need any metadata definition.

ARCs/arc-0004.md
```diff
@@ -527,12 +573,21 @@ any leading zeros, _unless_ `N` is zero, in which case only a single 0 character
...
 and [Transaction Types](#transaction-types).
 
+### Type hints
+The following hints to further describe an argument or return type are available. 
+Hints are arrays of functions in an argument description:
+
+* Max(N) - the maximum decimal or hexadecimal ('0x...') value,N, this integer can contain
+* Min(N) - the minimum decimal or hexadecimal ('0x...') value,N, this integer can contain
+* ...
+
```
barnjamin, line 590: I am v hesitant to add these as any kind of mini DSL for hints, id rather just put the hint in text in the description

jannotti: I agree. The ABI should only describe how to encode a value, not information about the semantics of the values. For example, maybe a method only takes odd numbers. That information should not be encoded in the ABI.

FrankSzendzielarz: Yes. This information should belong in the Arc24 extension to arc4 that serves for client generarion. Aliases can be established. For example uint16 witha Precision hint of a positive value works far better than a FixedNxM value. By the way disagree on semantics. Arc4 specifies a fixed point arithmetic type but *should* mandate santics around the type

PR #130, opened 2022-09-20
ARC-0012: Standard for Claimable Algorand Standard Assets (ASA)

joe-p:
The goal of this standard is to establish a standard in the Algorand ecosytem by which ASAs can be sent to an intended receiver even if their account is not opted in to the ASA.  

If integrated into ecosystem technologies including wallets, explorers, and dApps, this standard can provide enhanced capabilities around ASAs which are otherwise strictly bound at the protocol level to require opting in to be received.  This also enables the ability to "burn" ASAs by sending them to the claimable ASA vault associated with the global Zero Address.


nullun:
Nice work @joe-p and @silentrhetoric!

What's the reason behind returning the 0.2 Algo to the ASA creator address on redemption of the ASA? If I send USDC to someone who has yet to optin to it, why should Circle get 0.2 Algo when it's claimed? I understand that sending it back to the original sender isn't possible without state, but I'd have thought hardcoding the intended receiver of the ASA would be easier? I think this might also eliminate the need for a smart contract?

Additionally would the 2*MBR requirement only be for the first time the smart signature is used, and subsequent uses only require 0.1 Algo? If I'm nice I'd like to already prefund my address with 0.1 Algo so that all senders would only need to pay 0.1 Algo for the ASA they're choosing to send.

nullun:
I noticed in the TEAL for the subroutines (`handle_optin` and `handle_claim_axfer`) that they both contain a duplicate check as their first assert. It's the same condition that was used to enter in the first place. Is this for readability or pre-optimising? Great stuff 👍

nullun:
Nice work @joe-p and @SilentRhetoric!

What's the reason behind returning the 0.2 Algo to the ASA creator address on redemption of the ASA? If I send USDC to someone who has yet to optin to it, why should Circle get 0.2 Algo when it's claimed? I understand that sending it back to the original sender isn't possible without state, but I'd have thought hardcoding the intended receiver of the ASA would be easier? I think this might also eliminate the need for a smart contract?

Additionally would the 2*MBR requirement only be for the first time the smart signature is used, and subsequent uses only require 0.1 Algo? If I'm nice I'd like to already prefund my address with 0.1 Algo so that all senders would only need to pay 0.1 Algo for the ASA they're choosing to send.

ARCs/arc-0012.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Abstract
+
+The goal of this standard is to establish a standard in the Algorand ecosytem by which ASAs can be sent to an intended receiver even if their account is not opted in to the ASA.  
+
+TEAL stateless smart signature contract accounts corresponding to any Algorand account serve as a claimable ASAs account with logic that permits opting into incoming ASAs and subsequently for only the intended receiver to claim the ASA into their normal account.  A single associated stateful application on the network enables a cost--the minimum balance requirement to create an account and hold an ASA--to be refunded to the ASA creator when the asset is claimed by the receiver.
+
+If integrated into ecosystem technologies including wallets, epxlorers, and dApps, this standard can provide enhanced capabilities around ASAs which are otherwise strictly bound at the protocol level to require opting in to be received.  This also enables the ability to "burn" ASAs by sending them to the claimable ASAs account associated with the global Zero Address.
```
barnjamin, line 25:
```suggestion
If integrated into ecosystem technologies including wallets, explorers, and dApps, this standard can provide enhanced capabilities around ASAs which are otherwise strictly bound at the protocol level to require opting in to be received.  This also enables the ability to "burn" ASAs by sending them to the claimable ASAs account associated with the Zero Address.
```

ARCs/arc-0012.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Motivation
+
+Algorand requires users to opt in to receive any ASA, a fact which simultaneously:
```
barnjamin, line 29:
```suggestion
Algorand requires accounts to opt in to receive any ASA, a fact which simultaneously:
```

ARCs/arc-0012.md
```diff
@@ -0,0 +1,200 @@
...
+
+1. Grants account holders fine-grained control over the contents of their account by allowing them to select what assets to allow and preventing any unwanted tokens from appearing in their balance.
```
barnjamin, line 31:
```suggestion
1. Grants account holders fine-grained control over their holdings by allowing them to select which assets to allow and preventing receipt of unwanted tokens.
```

ARCs/arc-0012.md
```diff
@@ -0,0 +1,200 @@
...
+2. Annoys users when transactions fail and must be resubmitted after completing an opt in transaction--or asking someone else to do so.  This routine is jarring to people who are new to Algorand or are accutomed other blockchains without this requirement.
```
barnjamin, line 32:
```suggestion
2. Frustrates users and developers when accounting for this requirement especially since other blockchains do not have this requirement.
```

ARCs/arc-0012.md
```diff
@@ -0,0 +1,200 @@
...
+- **Explorer**: An off-chain application that allows browsing the blockchain, showing details of transactions.
+- **Wallet**: An off-chain application that stores secret keys for on-chain accounts and can display and sign transactions for these accounts.
+- **Mainnet ID**: The ID for the application that should be called upon claiming an asset on mainnet
+- **Testnet ID**: The ID for the application that should be called upoin claiming an asset on testnet
+- **Minimum Balance Requirement (MBR)**: The minimum amount of Algos which must be held by an account on the ledger, which is currently 0.1A + 0.1A per ASA opted into.
+
+### TEAL Smart Contracts
+
+There are two smart contracts being used for ARC-0012. A [smart signature](../assets/arc-0012/claimable_lsig.teal) and a [stateful application](../assets/arc-0012/claim_app.teal). In summary, the smart signature is the account used for holding the ASA and handling opt-in and claim logic. The stateful application is used atomically upon claim to ensure the creator of the ASA is paid 0.2 ALGO.
```
barnjamin, line 71:
```suggestion
There are two smart contracts specified for ARC-0012:
1. A [smart signature](../assets/arc-0012/claimable_lsig.teal) is the account used for holding the ASA and handling opt-in and claim logic. 
2. A [stateful application](../assets/arc-0012/claim_app.teal).  is used atomically upon claim to ensure the creator of the ASA is paid 0.2 ALGO.
```

ARCs/arc-0012.md
```diff
@@ -0,0 +1,200 @@
...
+
+Logic for determining whether the logic should branch to `handle_claim_axfer` or `handle_optin`.
+
+##### main
+
+Logic that contains necessary assertions, such as verifying the account is not being closed or rekey. Ultimately routes to `handle_claim_payment`, `handle_axfer` or `err`.
+
+#### Smart Contract (Stateful App)
+
+The application is used to ensure the creator is payed upon ASA claiming. A stateful application must be used to get the creator of the ASA. The application logic is broken down into multiple sections.
```
barnjamin, line 99:
```suggestion
The application is used to ensure the creator is refunded upon ASA claiming. A stateful application must be used to get the creator of the ASA. The application logic is broken down into multiple sections.
```

barnjamin (commented):
Started taking a pass and looks good.

Most of the comments here are tweaking working and removing any words we can without losing clarity.

joe-p:
One concern with the current implementation is that a claimable account could get spammed with unwanted assets. The problem is that if one wanted to get rid of the assets at the protocol-level, they would need to claim the asset which would send the 0.2 ALGO back to the creator meaning there isn't a lot of risk involved in sending spam assets, assuming people would like to keep their claimable ASA account clean. 

To counter this, we could implement an alternative action to claiming that the receiver can take. Instead of just being able to claim an ASA to their own personal account, we could make it so the smart signature can burn the token with the ALGO going to the FeeSink/RewardsPool instead of back to the creator. This would add more risk to sending spam assets, because spammers could loose that 0.2 ALGO forever. Thoughts?

PFunk420:
> One concern with the current implementation is that a claimable account could get spammed with unwanted assets. The problem is that if one wanted to get rid of the assets at the protocol-level, they would need to claim the asset which would send the 0.2 ALGO back to the creator meaning there isn't a lot of risk involved in sending spam assets, assuming people would like to keep their claimable ASA account clean.
> 
> To counter this, we could implement an alternative action to claiming that the receiver can take. Instead of just being able to claim an ASA to their own personal account, we could make it so the smart signature can burn the token with the ALGO going to the FeeSink/RewardsPool instead of back to the creator. This would add more risk to sending spam assets, because spammers could loose that 0.2 ALGO forever. Thoughts?

I like this idea. Perhaps there should be 3 options for the receiver:

1. Accept - ASA goes to receiver, fee goes to creator
2. Deny and return to creator - ASA and fee both go to creator
3. Deny and burn - ASA goes to burn, fee goes to FeeSink/RewardsPool

PFunk420:
> Let's say Charlie is new to Algorand and has a 0 ALGO wallet. Alice sends Charlie 1 Alice-Coin for 0.2 ALGO via the claimable ASA account. Then Bob sends Charlie 1 Bob-Coin for 0.1 ALGO via the claimable ASA account. If Charlie wants to claim Alice-Coin, but not Bob-Coin, then he will claim Alice-Coin but only 0.1 ALGO can be sent to Alice since Charlie's claimable ASA account still needs 0.2 ALGO (0.1 ALGO for account MBR, 0.1 ALGO for Alice-Coin). This basically means Alice paid 0.1 ALGO for Charlie to hold Bob-Coin in the claimable account. Since the order of claiming can't be guaranteed, it makes the most sense to make every sender send 0.2 ALGO.

Doubling the fee for every transaction to be fair to the first person to send to a wallet seems overkill IMO. I think it would be more desirable in the long run to lower overall fees for everyone even if it means "first senders" pay more. The "first sender" .1 Algo only gets paid once per wallet. If I'm doing an airdrop and have to send to 1000 wallets, 800 of which have their contract primed already, and 200 of which have not, I'm still paying overall less for my airdrop.

jannotti:
> > Deny and return to creator - ASA and fee both go to creator
> 
> The problem is that the creator could be opted out by the time the ASA is returned ...

I'm not sure I followed the entire conversation, but I wanted to point out that creators can not opt out of an ASA they created.

joe-p:
> > > Deny and return to creator - ASA and fee both go to creator
> > 
> > 
> > The problem is that the creator could be opted out by the time the ASA is returned ...
> 
> I'm not sure I followed the entire conversation, but I wanted to point out that creators can not opt out of an ASA they created.

Ah ok yeah wasn't sure on that one. In which case I like @PFunk420 idea of having three options

1. Claim: ASA -> claimer & 0.2 ALGO -> creator
2. Reject: ASA -> creator & 0.2 ALGO -> creator
3. Burn: ASA -> burn address & 0.2 ALGO -> FeeSink/RewardPool

ARCs/arc-0012.md
```diff
@@ -0,0 +1,199 @@
...
+2. Deriving the connected account's *corresponding claimable ASAs account* using the standard TEAL smart signature code
+3. Displaying any ASAs in the claimable ASAs account with the ability to claim them into the connected wallet account
+4. A smart sending tool to transfer ASAs from the connected wallet account to an intended receiver, dynamically diverting the transfer to the receiver's corresponding claimable ASAs account if the receiver is not opted in to the ASA.*
+
+With this toolset it is possible to view one's own claimable ASAs account, claim ASAs from it, and automatically utilize other people's claimable ASAs account to send them any ASA.
+
+### Burning Tokens
+
+Additionally, the smart sending tool can be used to burn ASAs by entering the global Zero Address `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ` as the intended recipient.  The app will automatically re-route the ASA to the unique "UN-claimable" ASAs account corresponding to the Zero Address.  An ASA sent here can never be claimed by anyone, rendering the transferred quantity permanently irretrievable and effectively "burned" in a verifiable way.
```
cusma, line 190: So there will 1 Burner Account for 1 ASA?

Reasoning on "ASA irreversible burn" I was inclined to have 1 single "ASA skin", well known in the ecosystem, that could only opt-in into ASA with no Clawback Address, so that anyone would now that is ASA is sent to the known App ID `123` is burned for sure (without having to check the specific burner logic).

WDYT?

joe-p: There will be one smart signature that can be used to burn any ASA. To verify how much of an ASA is burned, one would need to simply check the balance of that one account. The logic for this contract will show it is claimable by the zero address `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ`, but since this account cannot send transactions it is considered burnt.

cusma: Can the bruner address opt-in ASAs with non-null Clawback Address?

SilentRhetoric: No, but this intent of this ARC is not to neutralize Clawback as a protocol-level feature.

cusma: I'm not proposing to neutralize it. I'm just trying to understand if the proposed solution guarantees an _irreversible_ burning of ASAs.

SilentRhetoric: We can add some language to clarify explicitly that it is not possible to permanently burn a clawback-enabled ASA under any circumstances (as far as I am aware). This should help avoid any misunderstanding about the fact that this logicSig mechanism cannot override clawback.

TTogle:
What about a workaround to ASA opt-in if the ASA was created by an App opt-in to the app and opt-in to all assets the escrow creates?

fernando-deka:
What would happen if the asset is a Smart Asset ARC-0020 which is frozen by default?

SilentRhetoric:
> What would happen if the asset is a Smart Asset ARC-0020 which is frozen by default?

I think that will require a different solution, as ARC-0020 assets really only exist in the state data of an application.  What we are contemplating here would not solve for that; it would probably require a separate proposal.

joe-p:
Just to provide an update: We are currently working on a new proposal using smart contracts and boxes to allow the MBR to be sent back to the *sender*.

Basic idea is that a single master contract is called when sending someone an ASA. The master contract then either calls or creates a child contract that holds the ASA(s) for the intended receiver while recording who the sender is and how much the MBR was. The master contract is called when claiming an ASA and the master contract will call the child to trigger an inner transaction that sends the ASA to the end-user.

spencercap:
this assumes that the ASA receiver has a bit of $ALGO and is not a completely new acct, yes? 

min bal needed being... 0.201 = 0.1 (for existing on-chain) + 0.1 (for holding the asset) + 0.001 (for making the txn)

robdmoore:
Huge +1 for this, I'm building an experience where users get issued certificates and it would be really handy if I can airdrop them into a "claiming account". It'll make the UX and code way simpler.

What's the current status of this one? It's not entirely clear from the PR.

joe-p:
@robdmoore The ARC is in need of a rewrite. I have ditched the lsig-based implementation in favor of a box-based implementation. One implementation of it can be seen [here](https://github.com/joe-p/arc12/blob/master/arc12.rb) which is written in TEALrb, a Ruby-based DSL for writing smart contracts. Another implementation in TypeScript can be seen [here](https://github.com/joe-p/TEALScript/blob/main/examples/arc12/arc12.ts). There is also a JS SDK reference [here](https://github.com/joe-p/arc12/blob/master/arc12-sdk/src/index.ts). All of the above has been reviewed internally and tested. 

Next step is to talk to dApps/wallets to get some feedback on implementation and then hopefully have a freeze In 23Q1 to send off for auditing.

joe-p:
The ARC has been updated with information about the latest implementation that uses box storage. The current implementation is likely close to the final version unless major feedback needs to be implemented. Next steps are to consult with wallets and dApps in the ecosystem and then get the code audited.

fergalwalsh:
As far as I understand this ARC proposes a specific app will be used as the Master. I think it would be worth clarifying in the spec that a single well known app id will be used. There should be a space in the doc to be updated with the exact ID once it is deployed. In this case also the code provided isn't a reference implementation but the actual only implementation (unlike most other ARCs). This is a great thing in my view but it needs to be made explicit.

BurnusWolf:
Hi @joe-p! 
I think this ARC is a great idea and makes a lot of sense. 
I just have a few concerns for some of the edge-cases that might arise from its implementation. More specifically:

1. AFAIU, there’ll be 3 options here for the users to handle these requests: `Accept` | `Reject` | `Destroy`. 
In the `destroy` option, the ASA creator can still claw back the ASA amount, meaning that they’ve actually only lost the 0.1 Algo that gets sent to the FeeSink/RewardPool. This might still not be expensive enough to stop spamming behaviors directed to well-known accounts (i.e. I try to get established ecosystem projects to opt-in to a created ASA so that everyone will think they support that asset). 
2. What happens if the ASA creator claws back the asset amount from the escrow account, before the user accepts it?

joe-p:
Great point @fergalwalsh I'll update the ARC to reflect the official app ids for respective networks. 

@BurnusWolf 

> AFAIU, there’ll be 3 options here for the users to handle these requests: Accept | Reject | Destroy.
In the destroy option, the ASA creator can still claw back the ASA amount, meaning that they’ve actually only lost the 0.1 Algo that gets sent to the FeeSink/RewardPool. This might still not be expensive enough to stop spamming behaviors directed to well-known accounts (i.e. I try to get established ecosystem projects to opt-in to a created ASA so that everyone will think they support that asset).

Just to clarify, claim and reject are the only two options for the end-user. Reject does what you seem to be describing for destroy. 

I agree, 0.1 ALGO isn't a significant mitigation for targeted spam, but I would argue that people should **not** consider ASAs in a vault to represent the desired ASAs of the end-user since the ASAs can't be transfered or utilized until they are claimed by the end-user. Also, since rejecting is completely free, one could create a off-chain filter that automatically rejects specific assets.

> What happens if the ASA creator claws back the asset amount from the escrow account, before the user accepts it?

This is a good question that I've been thinking about recently after someone recently brought it up in a meeting. In the current implementation, the end-user would still have to claim/reject the asset even if the balance was zero. If the balance is zero, it would make sense for front-ends to only allow rejects which are free for the end-user. I'm curious if anyone can come up with additional concerns regarding clawbacks...

emg110:
Even after the update and use of the new APPLICATION VAULT approach and using boxes (great idea by the way!) is `KNOWING THE USER ACCOUNT ADDRESSES` still a requirement to send arc12 ASAs? I mean do target accounts need to exist and addresses known before ASA distribution?

joe-p:
@emg110 It depends on what you mean by "exist". The account doesn't need to be on the ledger (can have 0 ALGO), but you do need to know the address to reserve the ASA for that particular account.

For example, if an app is onboarding a new user, they must do keypair generation to get the address, but the account doesn't actually need to be funded until they are ready to claim the ASA.

PR #137, opened 2022-10-20
ARC-29 Convention for wallets to prevent spam inside notes

SudoWeezy:
This ARC aims to change the wallets' interface to protect users from reading unintended messages and avoids having clickable URLs inside a transaction's note. 
Algorand's transaction allows sending messages through notes. Unfortunately, as Algorand'transactions do not cost much, this also means that malicious (or just advertising) accounts can send spam to users.

Initial discussion:
https://forum.algorand.org/t/arc-spam-note-transaction-filter/8042/6


deanstef (commented):
Thank you for putting this together.

My first thought: we cannot just standardise how wallets show note fields of a transactions. Notes will be anyway available on the blockchain. Also, I do not really like this approach because it could affect honest activities (e.g. testing, communications, ad-hoc use cases). 

Spam is inevitable. However, wallets should somehow advice users that a certain content could be dangerous. Some ideas are:
- Managing a list of keywords present in spam messages;
- Maintain and update a list of blacklisted addresses;
- other?

KillerGasy:
> Thank you for putting this together.
> 
> My first thought: we cannot just standardise how wallets show note fields of a transactions. Notes will be anyway available on the blockchain. Also, I do not really like this approach because it could affect honest activities (e.g. testing, communications, ad-hoc use cases).
> 
> Spam is inevitable. However, wallets should somehow advice users that a certain content could be dangerous. Some ideas are:
> 
> * Managing a list of keywords present in spam messages;
> * Maintain and update a list of blacklisted addresses;
> * other?

I do like this idea of a Blacklist of wallets, but I guess the issue would be in the Algorand space once you are added to the blacklist, with little to no effort, you are back up and running on a new wallet. 

One way could be only allowing push notifications in relation to NFTs. If a creator of an NFT would like to send you this notification, they would do some from the CREATOR or MANAGER address or something similar to that.

deanstef:
> 

Right. But even if you create a new account, you will need to fund it with ALGOs. If one of the funders is a blacklisted account, that new wallet could be considered malicious as well. This is quite complex though. I'd love to hear feedback on that.

I think push notifications should not be too restrictive. It is quite a nice feature. Also, I would like to be notified if my address is getting spammed all the time. Hiding this kind of information from a user does not solve the problem IMHO.

yigitguler:
Hi,

Due to the low price of transactions, the note field contains a risk of being abused by providers. We recently witnessed a situation where many people received spam transactions from a well-known project. We are working on solutions that will protect users from this kind of notifications. 

However, I am not sure if this topic can be solved with an ARC. Most of the recommendations in this document are UI guidelines. IMO every wallet team should do their best to protect their users.

SudoWeezy:
> Thank you for putting this together.

**These ideas came up after discussing them with 15 members of the community during the Discord ARC lounge on the 19th of October.**

> My first thought: we cannot just standardise how wallets show note fields of a transactions. Notes will be anyway available on the blockchain. Also, I do not really like this approach because it could affect honest activities (e.g. testing, communications, ad-hoc use cases).

**We don't want to change how they display it. The idea is to find rules that prevent "Push Notification" when you receive a transaction (so you are not annoyed if someone one day wants to send you 1000 txn of 0 Algo).
They will still be accessible in your history. Adding a warning to warn users if it looks spammy (Url is also an idea)** 

> Spam is inevitable. However, wallets should somehow advice users that a certain content could be dangerous. Some ideas are:

> * Managing a list of keywords present in spam messages;

**For the moment we only considered clickable URLs because it will be hard to find a good list of keywords IMO**

> * Maintain and update a list of blacklisted addresses;

**We discuss the blacklist, the issue is you will always be one step behind ...**

PR #143, opened 2022-10-31
ARC15-ALGOSMS with Box Storage

SudoWeezy:
This is a version of https://github.com/algorandfoundation/ARCs/pull/49 using box storage


ARCs/arc-0015.md
```diff
@@ -0,0 +1,254 @@
...
+  otPK: Uint8Array;
+
+  /** cipherText of the ALGOSMSV0_MESSAGE structure * */
+  cT: Uint8Array;
+}
+```
+
+```ts
+// UINT8Array ALGOSMSV0_MESSAGE contents as msgpack encoded object
+interface ALGOSMSV0_MESSAGE {
```
fabrice102, line 69: Do we want to enforce the message object format?
I think it may be hard to know in advance what will be used.
msgpack may be hard to decode in other chains or in ZK proofs for example.
So why not allowing any arbitrary binary message?

Also, if "from" is added, there is a need for the wallet to confirm that it is not impersonated, which is not straightforward as there are two .algo domain name providers for example.

ARCs/arc-0015.md
```diff
@@ -0,0 +1,254 @@
...
+For the rest of the document, we will use the term *Receiver* to represent the set of accounts (public account ed25519, multisig account, contract account, application account)
+
+> Comments like this are non-normative.
+
+
+### Account's message Application
+
+To receive a message, the *Receiver* **MUST** create an application that follows the following convention:
+
+- A Local State named `public_key` **MUST** containing a public account `Ed25519-Address`
```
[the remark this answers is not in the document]
SudoWeezy: It was part of the original proposal, I kept it because of wallets:
- Already have a way to import Private keys 
- Creating a new ED25519 is part of the algo SDK

But yes, in theory, with this approach, we can ask users to use Curve25519 directly

ARCs/arc-0015.md
```diff
@@ -0,0 +1,254 @@
...
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in <a href="https://www.ietf.org/rfc/rfc2119.txt">RFC-2119</a>.
+
+For the rest of the document, we will use the term *Receiver* to represent the set of accounts (public account ed25519, multisig account, contract account, application account)
+
+> Comments like this are non-normative.
+
+
+### Account's message Application
+
+To receive a message, the *Receiver* **MUST** create an application that follows the following convention:
```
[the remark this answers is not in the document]
SudoWeezy: The idea behind this was:
- Allow the account to say if he wants to receive messages (by creating this application)
- Using Local storage, you are opt-in to the app so wallets can parse for opt-app

You are right; we miss the part about identifying the SMS Application. Maybe add a global state, "ARC-15."

Global Fixed application was thought too. However, it did not seem right to store the public key  + encoded messages of every account in the same app.

ARCs/arc-0015.md
```diff
@@ -0,0 +1,241 @@
...
+```json
+{
+  "name": "ARC_0015",
+  "desc": "Interface for an encrypted messages application",
+  "methods": [
+    {
+      "name": "write",
+      "desc": "Write encrypted text to the box inbox",
+      "args": [
+        { "type": "byte[]", "name": "text", "desc": "Encryted text provided by the sender." }
```
fabrice102, line 50:
```suggestion
        { "type": "byte[]", "name": "text", "desc": "Encrypted text provided by the sender." }
```

ARCs/arc-0015.md
```diff
@@ -0,0 +1,241 @@
...
+      ],
+      "returns": { "type": "void" }
+    },
+    {
+      "name": "remove",
```
fabrice102, line 55: I'm not sure about this function:
nothing can be removed from the blockchain.

If we think it's important, I would recommend adding a warning that people can still retrieve "Removed" messages. There is no way to prevent that.

ARCs/arc-0015.md
```diff
@@ -0,0 +1,241 @@
...
+
+The associated smart contract **MUST** implement the following ABI interface:
+```json
+{
+  "name": "ARC_0015",
+  "desc": "Interface for an encrypted messages application",
+  "methods": [
+    {
+      "name": "write",
+      "desc": "Write encrypted text to the box inbox",
```
fabrice102, line 48: Do you require payment for the min balance accrual if using boxes?
Also, some implementations may just decide to do nothing and not store messages in state/boxes:
an indexer can always retrieve all messages even if not stored.

ARCs/arc-0015.md
```diff
@@ -0,0 +1,241 @@
...
+## Specification
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in <a href="https://www.ietf.org/rfc/rfc2119.txt">RFC-2119</a>.
+
+> Comments like this are non-normative.
+
+### Account's message Application
+
+To receive a message, an Account **MUST** create an application that follows this convention:
+
+- A Local State named `public_key` **MUST** contain an `NACL Public Key (Curve 25519)`
```
fabrice102, line 29: What format are you considering?
It looks like you are using base64, which is not optimal storage-wise as raw bytes would be shorter.

Is it local state or global state?

Also, I would recommend adding some versioning to allow new cryptosystems.
Maybe: a key `arc` which contains `arc15-nacl-curve25519`?

ARCs/arc-0015.md
SudoWeezy: I think it is still important to have a way for "unlocking" Algo used for the Box.
Nevertheless, I will add the Warning!

ARCs/arc-0015.md
SudoWeezy: I modified the implementation to use RawEncoding instead of base64.
Added also the `arc` key

assets/arc-0015/main.py
```diff
@@ -0,0 +1,119 @@
+import algosdk
+import base64
+from pyteal import *
```
tzaffi, line 3: For the purposes of clarity, I recommend not using `*` imports here. It's easier for the reader to discern where a particular python object is sourced in that case.

PR #147, opened 2022-11-14
ARC-72. Application Non-Fungible Token

SudoWeezy:
It is a first draft inspired by ERC-721 for NFT inside an Application using box storage.

It is needed for:
Creator, to create NFT without using ASA
Wallets, when you Buy/Trade, NFTs need to be displayed properly inside the transaction window
dApps, to display and use NFTs


SudoWeezy:
Close in favor of https://github.com/algorandfoundation/ARCs/pull/162

PR #150, opened 2022-12-08
ARC-0032 - Application Specification

barnjamin:
WIP

Things to consider and provide feedback on:

```
1) Should we propose the `DefaultArgument` as a separate ARC similar to the `read-only` attribute? There is probably enough discussion that could happen around how to describe and resolve the data for it that it makes sense.

2) Should we allow type _aliases_ (in addition to structs) like: `type digest = bytes[32]`? probably yes.

3) Is there anything else that might be needed in an ErrorSpec besides message? related, should we key the error messages by the string value of the PC since some things (namely json) don't allow numeric keys?
```

### DIFF
--- a/ARCs/arc-0032.md
+++ b/ARCs/arc-0032.md
@@ -2,7 +2,7 @@
 arc: 32
 title: Application Specification
 description: A specification for fully describing an Application, useful for Application clients.
-author: Benjamin Guidarelli (@barnjamin), ...
+author: Benjamin Guidarelli (@barnjamin)
 discussions-to: https://github.com/algorandfoundation/ARCs/issues/150
 status: Draft
 type: Standards Track
@@ -68,7 +68,7 @@ type SourceSpec = {
 TODO: describe diff between Declared/Reserved and motivation for providing both
 TODO: describe how to decode the type and the value given the decoded type
 
-The schema of an application is critical to know prior to creation since it is immutable after create. It also helps clients of the application understand the data that is available to be queried from off chain. Individual fields can be referenced from the [default arguments](#default-arguments) to provide input data to a given ABI method.
+The schema of an application is critical to know prior to creation since it is immutable after create. It also helps clients of the application understand the data that is available to be queried from off chain. Individual fields can be referenced from the [default argument](#default-argument) to provide input data to a given ABI method.
 
 While some fields are possible to know ahead of time, others may be keyed dynamically. In both cases the data type being stored MUST be known and declared ahead of time. 
 
--- a/ARCs/arc-0032.md
+++ b/ARCs/arc-0032.md
@@ -1,14 +1,14 @@
 ---
-arc: <to be assigned>
+arc: 32
 title: Application Specification
 description: A specification for fully describing an Application, useful for Application clients.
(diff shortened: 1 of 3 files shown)


ARCs/arc-0032.md
```diff
@@ -0,0 +1,233 @@
...
+To build an error spec, compile the source teal program with its source map, and map the `PC` of any `assert` or `err` opcodes with a comment on the immediately preceding line into the `ErrorSpec` map. 
+
+ex: the following will produce an entry in the error spec like `{..., 10: "// Sorry but 2 is not less than 1", ...}` assuming the `PC` of the `assert` op is 10.
+```
+int 1    
+int 2
+<
+// Sorry but 2 is not less than 1
+assert
+```
```
cusma, line 199: Just chiming in to say that I really like the idea of having a structured way of encoding/enumerating App’s errors.

Something like: if I fail on `PC=10` we can use `sourcemap` to easily return something like:

`ERR-010: This is a bad error` 

to the client so that an eventual App front-end could provide better error handling experience.

Also this would enable more "granularity" at testing level, to prove that a given transaction is failing due to `ERR-010`, as expected, and not due to `ERR-042` or anything else.

barnjamin: Good point, ill just caution against adding the pc to the error message if that is what the 10 in  `ERR-010`  is meant to signify since the pc could change with every edit to the program

cusma: Right, also maybe want to have the same error code at different PCs, like `ERR-X: Insufficient funds` that can be used in the execution path of different methods.

cusma: The important thing is having a way to enumerate the error causes. At PyTeal (or Beaker) / SDK level some helper classes could facilitate errors mapping, something like:

```python
Assert(Int(2) < Int(1), comment=Errors.MATH)
```

The `Errors` class (or something similar) could be used to generate the related Errors AppSpec.

barnjamin: This is indeed what I did for the AMM example:
https://github.com/algorand-devrel/beaker/blob/master/examples/amm/amm.py#L47-L57

robdmoore:
Should this also have Template variables too?

ARCs/arc-0032.md
```diff
@@ -0,0 +1,233 @@
...
+type ABIType = string;
+
+// Fields that have an explicit key
+type DeclaredSchemaValueSpec = {
+  type: AVMType | ABIType;
+  key: string;
+  desc: string;
+}
+
+// Fields that have an undetermined key
```
achidlow, line 100: What about the case of known keys, but they're autogenerated from some prefix. E.g. if you have `blob_0`, `blob_1`, and so on. These keys are determined already, but would be unwieldy to list individually?

ryanRfox:
I'd like to see a section for including the high level language source code. Non-trivial apps are most often written in beaker-pyteal, reach, tealscript, tealish, etc. and much easier to reason about what they do compared to their generated TEAL program. I realize the HLL compiler may change over time, which may then generate different TEAL output. It's my understanding from speaking with auditors that they evaluate the HLL not the TEAL. I feel this ARC should include the implementation source code.

PR #151, opened 2022-12-13
ARC-34 Governance proposal process

SudoWeezy:
This ARC will be used to standardize the creation of new proposals for governance.

Questions that need to be answered have been listed in the `Rationale` part.

Please refer to this part and give your thought.

PR #152, opened 2022-12-13
ARC-33 Becoming an xGov

SudoWeezy:
This ARC will be used to standardize the way to become an xGov.

Questions that need to be answered have been listed in the `Rationale` part.

Please refer to this part and give your thought.


ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+ 
+### What is an xGov
+Expert Governors (xGov) are decentralized decision makers who demonstrate enduring commitment, are knowledgeable of the realities of the Algorand community, and whose interests are aligned with the good of blockchain.
+ 
+xGov **SHOULD** understand the community's needs and align their interests with those of the Algorand ecosystem.
+xGov **MAY** propose measures for a vote.
+xGov **MAY** participate in the designation of proposals.
+ 
+### Requirement to become an xGov
+The Foundation suggests that to become an xGov an Algo holder **MUST**:
```
solkimicreb, line 28: This is not super clear to me. Do we need to participate in another reward-bearing program on top of the regular governance or is regular governance enough to qualify? Can we become xGov without being regular governors but participating in another qualified reward-bearing AF program?

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+- Participate in another reward-bearing AF program
+- Commit the rewards to an xGov DAO for a period
```
solkimicreb, line 30: How can this be enforced? What does "commit" mean here? Is it enough if we maintain a balance higher than the received rewards on our account? Will a bot scrape the chain and look for reward transactions for all xGov accounts?

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+ 
+The foundation will maintain lists of eligible reward-bearing programs, and a list of qualified DAOs.
```
solkimicreb, line 32: Who will maintain this list? How often can the list change and on whose authority? Can a single person add and delete items to/from the list?

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+### What is an xGov
+Expert Governors (xGov) are decentralized decision makers who demonstrate enduring commitment, are knowledgeable of the realities of the Algorand community, and whose interests are aligned with the good of blockchain.
+ 
+xGov **SHOULD** understand the community's needs and align their interests with those of the Algorand ecosystem.
+xGov **MAY** propose measures for a vote.
+xGov **MAY** participate in the designation of proposals.
+ 
+### Requirement to become an xGov
+The Foundation suggests that to become an xGov an Algo holder **MUST**:
+- Participate in another reward-bearing AF program
```
solkimicreb, line 29: Is there any chance we can count having "online stake" as a "reward bearing program"? I understand it goes against TVL expansion but it seems like online stake is gradually dropping. (Number of nodes may be growing though).

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+- Commit the rewards to an xGov DAO for a period
+ 
+The foundation will maintain lists of eligible reward-bearing programs, and a list of qualified DAOs.
+ 
+#### Example
+Alice participates in governance during Q1 2023
+the potential reward is 5 Algo, Alice:
+1) Commit 100 Algo in an AF program
+2) Express her will to participate in xGov
+3) Will not receive 5 Algo at the end of Q1 2023. They are instead locked in one big xGov pool for 1 year
```
solkimicreb, line 39: The technical implementation is not clear to me as I mentioned earlier. Does this mean all reward-bearing AF qualified services have to explicitly integrate with xGov at least for reward withdrawal?

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+4) Will receive 5 Algo at Q1 2024
+ 
+## Rationale
+In this ARC we seek community input on issues such as, but not limited to:
+- The types of reward-bearing activities that **SHOULD** be included in the qualification list (ie. general governance)
+- Whether xGovs **MUST** vote on all measures to keep their status
```
solkimicreb, line 45: It's a definite no from me. Adding a minimum participation threshold to votes makes a lot more sense to me. Different xGovs have different interests but they will likely have a community behind them that they can convince to vote if it's important for them. Forcing NFT people to vote on DEX votes will just create noise as it feel like there is little overlap between the two communities atm.

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+- Whether xGovs tokens **SHOULD** be sold/transferred or **SHOULD** be burned when exiting the xGov DAO
```
solkimicreb, line 46: How do we get xGov tokens? What are they used for? It's the first time these tokens are mentioned in the document.

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+- Whether xGovs tokens expire after a year or gain more voting power if the stake remains in the DAO
+- Whether there **SHOULD** be a minimum and a maximum amount of Algos required to become an xGov
```
solkimicreb, line 48: I would add a minimum threshold but keep voting rights equal once the threshold is reached. Maybe even require the held algos to be online stake?

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+- Whether there be a maximum amount of Algos that can be locked by xGovs for the each period, limiting the total number of xGovs
```
solkimicreb, line 49: Having a nice minimum threshold is enough to avoid a huge number of xGovs imo.

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+- Whether xGovs **SHOULD** be disqualified and lose their status if they behave in a suboptimal fashion
+- What types of suboptimal behavior would trigger disqualification or loss of status?
```
solkimicreb, line 51: Slashing is only acceptable if it's fully decentralized with a large majority in my opinion. If we do this I would not create rules in advance but allow xGov proposals to disqualify other xGov members with a big majority of votes.

Even if we allow this what is stopping the xGov member to join again (maybe with another wallet)? Do we plan to have any kind of doxxing?

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+- Whether new xGovs be allowed into the pool at every quarter or only once per year
```
solkimicreb, line 52: Questions like this should be polls somewhere probably.

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+Alice participates in governance during Q1 2023
+the potential reward is 5 Algo, Alice:
+1) Commit 100 Algo in an AF program
+2) Express her will to participate in xGov
+3) Will not receive 5 Algo at the end of Q1 2023. They are instead locked in one big xGov pool for 1 year
+4) Will receive 5 Algo at Q1 2024
+ 
+## Rationale
+In this ARC we seek community input on issues such as, but not limited to:
+- The types of reward-bearing activities that **SHOULD** be included in the qualification list (ie. general governance)
```
solkimicreb, line 44: Not sure yet (apart from the obvious ones) but I think it is a nice opportunity to enforce certain security standards, like immutable SCs and a mandatory none-eligible period after a new SC suite goes to production.

ARCs/arc-0033.md
paulhinrichsen: You can always make this a must, but also leave an opt-on option on every vote if you want to track activity but not force someone to answer.

ARCs/arc-0033.md
paulhinrichsen: Yea, what is the purpose of the tokens? Would rather not have tokens involved unless they serve a really specific purpose.

ARCs/arc-0033.md
paulhinrichsen: agreed with @solkimicreb

ARCs/arc-0033.md
paulhinrichsen: yea I'm just not sure how you're going to police this by creating issues of centralization in itself if you give the AF foundation the ability to pick and choose suspensions in some sort of process that isn't completely written in code.

ARCs/arc-0033.md
paulhinrichsen: i think this depends on how much work it creates for the foundation to manage. If it's a huge task to run this each quarter and the resources could be spent elsewhere, then I understand limiting it to one year. but if it's an automated process, I would keep open each quarter.

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+ 
+The foundation will maintain lists of eligible reward-bearing programs, and a list of qualified DAOs.
+ 
+#### Example
+Alice participates in governance during Q1 2023
+the potential reward is 5 Algo, Alice:
+1) Commit 100 Algo in an AF program
+2) Express her will to participate in xGov
+3) Will not receive 5 Algo at the end of Q1 2023. They are instead locked in one big xGov pool for 1 year
+4) Will receive 5 Algo at Q1 2024
```
KillerGasy, line 40: Are there no extra incentives or slashing being an xGov? What is the purpose of the pool?

ARCs/arc-0033.md
KillerGasy: Agreed but at the same time we don't want to price out a lot the community and make it only whales who are an xGov?

ARCs/arc-0033.md
KillerGasy, line 52: How hard is it to change the members of xGov? Who is building the xGov DAO Platform?

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+ 
+## Rationale
+In this ARC we seek community input on issues such as, but not limited to:
+- The types of reward-bearing activities that **SHOULD** be included in the qualification list (ie. general governance)
+- Whether xGovs **MUST** vote on all measures to keep their status
+- Whether xGovs tokens **SHOULD** be sold/transferred or **SHOULD** be burned when exiting the xGov DAO
+- Whether xGovs tokens expire after a year or gain more voting power if the stake remains in the DAO
+- Whether there **SHOULD** be a minimum and a maximum amount of Algos required to become an xGov
+- Whether there be a maximum amount of Algos that can be locked by xGovs for the each period, limiting the total number of xGovs
+- Whether xGovs **SHOULD** be disqualified and lose their status if they behave in a suboptimal fashion
```
KillerGasy, line 50: Yeah with slashing to motivate people to not act in this way!

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+The purpose of this ARC is to define the standard by which xGov status is achieved.
+ 
+## Specification
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in <a href="https://www.ietf.org/rfc/rfc2119.txt">RFC-2119</a>.
+ 
+### What is an xGov
+Expert Governors (xGov) are decentralized decision makers who demonstrate enduring commitment, are knowledgeable of the realities of the Algorand community, and whose interests are aligned with the good of blockchain.
+ 
+xGov **SHOULD** understand the community's needs and align their interests with those of the Algorand ecosystem.
+xGov **MAY** propose measures for a vote.
```
KillerGasy, line 24: Will there need to be a Quorum amongst XGov's before it goes out to the public for a vote or can an xGov propose an unlimited amount of things?

KillerGasy: Or it could be that they are limited to one per quarter

ARCs/arc-0033.md
KillerGasy, line 30: Soft commit or hard commit?

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+- Participate in another reward-bearing AF program
+- Commit the rewards to an xGov DAO for a period
+ 
+The foundation will maintain lists of eligible reward-bearing programs, and a list of qualified DAOs.
+ 
+#### Example
+Alice participates in governance during Q1 2023
+the potential reward is 5 Algo, Alice:
+1) Commit 100 Algo in an AF program
+2) Express her will to participate in xGov
```
KillerGasy, line 38: Everyone who expresses becomes an xGov?

SudoWeezy: In this example, yes, but keep in mind that this is an explanation of how it could work, not the future process.

ARCs/arc-0033.md
SudoWeezy: This explains how it could work, not the future process.
Please comment on the # Rationale part

ARCs/arc-0033.md
SudoWeezy: This is just an example that explains how it could work, not the future process.
Please comment on the # Rationale part

ARCs/arc-0033.md
paulhinrichsen: from their previous posts , I think they're referring to normal Algorand governance, but the fact they use the word "another" doesn't make sense to me

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+description: Explanation on how to become Expert Governors.
+author: Stéphane Barroso (@SudoWeezy), Adriana Belotti, Massimo Morini, Michel Treccani, John Woods, Shai Halevi
+discussions-to: https://github.com/algorandfoundation/ARCs/issues/
+status: Draft
+type: Meta
+created: 2022-11-22
+---
+ 
+## Abstract
+xGov status gives a governor the right to vote on proposals raised by the community, to ratify the proposals as valid for final community vote, where said proposals pertain to spending a previously specified amount of Algo in a given epoch on particular initiatives.
```
averagezen, line 13: In general, I'm curious what part the Foundation and/or Algorand Ventures, will have in the day-to-day operations, support, and implementation of xGov ratified proposals and the ratification process in general as well as any potential internal or utility measures for the xGov program itself.

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+ 
+The purpose of this ARC is to define the standard by which xGov status is achieved.
+ 
+## Specification
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in <a href="https://www.ietf.org/rfc/rfc2119.txt">RFC-2119</a>.
+ 
+### What is an xGov
```
averagezen, line 20: What are the particular metrics that could be used to determine a potential xGov's understanding of the community and its needs? How can we determine whether a potential xGov is aligned with the growth of the Algorand ecosystem?

ARCs/arc-0033.md
averagezen: I would also like clarification on whether xGovs are generation proposals are essentially sorting through community made proposals. Can an xGov make a proposal at all or is that a conflict of interest?

ARCs/arc-0033.md
averagezen, line 50: I believe, just as general governance stake can become ineligible, so too should an xGov. The details fleshed out in 51 is crucial to justifying this portion however.

ARCs/arc-0033.md
averagezen, line 51: We really can't know the full extent of answers to this question without a fully fleshed out structure with specialized bug bounty hunters willing to stress test any number of harmful scenarios that might reveal potential bad actors. Without more information on what sort of "suboptimal behavior" xGovs might be faced with, it is hard to say what might trigger such a disqualification.

ARCs/arc-0033.md
averagezen, line 52: I would think that twice a year would be optimal. The halfway point between general governance would give any new xGovs time to catch up before the next General vote.

averagezen: > Questions like this should be polls somewhere probably.
Or a General Governance vote in itself.

ARCs/arc-0033.md
averagezen: Staking in any Algo LP (could be tracked in much the same way as current Governance LPs are tracked) with a minimum volume and TVL threshold TBD would come to mind as a possible alternate qualifier to traditional Governance. Possibly NFT staking or staking of certain other ASA's, though choosing candidates for these might be a vote in itself.

ARCs/arc-0033.md
GhostOfMcAfee, line 49: There should not be a total maximum Algo commitment in any circumstance. This is particularly true if there is no maximum amount for being an xGov. Otherwise, a few large entities could come in and reach the maximum total commit threshold early on, thereby locking out a vast majority of Algo holders. 

Even if there is a maximum amount for being an xGov, there should not be a total maximum that can be locked for xGov generally. Whales can split wallets and still achieve the same thing as before by committing multiple wallets just below the per wallet limit. Instead, I will once again say that participation nodes should be the weeding mechanism. In such a scenario, whales can still split those wallets to try to evade the per wallet cap, but that also means they need to add more nodes, which makes collusive behavior (even among a single entity) more difficult. There is no way to stop whales from splitting wallets to evade either a per wallet cap or a from taking advantage of a global cap freeze out retail an control the process. But, at least a node requirement plus a per wallet cap (without a global cap) means they make it harder on themselves and nominally increase decentralization if they want to play those games.

ARCs/arc-0033.md
GhostOfMcAfee: "suboptimal" needs to be defined.

ARCs/arc-0033.md
GhostOfMcAfee: Less than 90% participation node uptime during any quarter or less than 95% uptime during a 1 year period. 

Failing to vote at any time in quarter. 

Failing to comment on X number of proposals in a 1 year period (we want proactive participants).

ARCs/arc-0033.md
GhostOfMcAfee: Bring them in every quarter. Flexibility is important.

ARCs/arc-0033.md
GhostOfMcAfee: > Agreed but at the same time we don't want to price out a lot the community and make it only whales who are an xGov?

These are two different issues. The prior line dealt with a maximum single commit. This line deals with a maximum total commit. A maximum total commitment of Algos can lead to centralization. See my comment below.

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+Expert Governors (xGov) are decentralized decision makers who demonstrate enduring commitment, are knowledgeable of the realities of the Algorand community, and whose interests are aligned with the good of blockchain.
+ 
+xGov **SHOULD** understand the community's needs and align their interests with those of the Algorand ecosystem.
+xGov **MAY** propose measures for a vote.
+xGov **MAY** participate in the designation of proposals.
+ 
+### Requirement to become an xGov
```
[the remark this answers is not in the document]
KillerGasy: This is an interesting point

ARCs/arc-0033.md
KillerGasy: I don't think every person should be able to become an xGov

ARCs/arc-0033.md
```diff
@@ -0,0 +1,60 @@
...
+2) Express her will to participate in xGov
+3) Will not receive 5 Algo at the end of Q1 2023. They are instead locked in one big xGov pool for 1 year
+4) Will receive 5 Algo at Q1 2024
+ 
+## Rationale
+In this ARC we seek community input on issues such as, but not limited to:
+- The types of reward-bearing activities that **SHOULD** be included in the qualification list (ie. general governance)
+- Whether xGovs **MUST** vote on all measures to keep their status
+- Whether xGovs tokens **SHOULD** be sold/transferred or **SHOULD** be burned when exiting the xGov DAO
+- Whether xGovs tokens expire after a year or gain more voting power if the stake remains in the DAO
```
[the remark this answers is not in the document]
KillerGasy: There needs to be an incentive to be an XGov!

ARCs/arc-0033.md
KillerGasy: There also needs to be a charter for the xGov to stop things like front running or people submitting proposals for personal gains and repercussions for doing so

ARCs/arc-0033.md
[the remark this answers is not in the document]
oysterpack: Totally agree. Security of the network is fundamental. Consensus participation shows commitment.

All ALGO committed to governance by xGov should have a mandatory requirement for consensus participation.

ARCs/arc-0033.md
oysterpack: Consensus  participation should be mandatory for xGovs.

ARCs/arc-0033.md
oysterpack: These are good questions, but we should focus on defining the high level requirements. These are great questions for the design phase.

As a requirement, I support the requirement.

ARCs/arc-0033.md
oysterpack, line 32: The initial list should be proposed by the AF along with the community. However, final approval and management of the list should go through Governance, i.e., voted on.

ARCs/arc-0033.md
oysterpack, line 45: Yes. Voting on all measures should be required, but there should be an option to abstain from voting. xGovs should be aware of all measures, but if the xGov is not qualified to vote on the measure, then the xGov should have the option to abstain from the vote.

ARCs/arc-0033.md
oysterpack: There should be no xGov tokens and there is no need. xGov status should simply be tracked by smart contracts.

ARCs/arc-0033.md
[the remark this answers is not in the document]
oysterpack: Consensus participation should be required for xGov, thus the min should be 1 online staked ALGO. There should also be a max to defend against whales controlling xGovs. The max is TBD. However, imposing a max requires some form of on-chain digital ID to defend against whales creating multiple accounts to work around the max limit. 

Vendible's digital ID solution is a good candidate to integrate into xGov.

ARCs/arc-0033.md
cusma: > Instead, I will once again say that participation nodes should be the weeding mechanism. In such a scenario, whales can still split those wallets to try to evade the per wallet cap, but that also means they need to add more nodes, which makes collusive behavior (even among a single entity) more difficult.

This is not accurate, who participates in consensus are Algorand Accounts not Algorand Nodes. This means you can still have different Participating Accounts on the same Node.

GhostOfMcAfee: > This is not accurate, who participates in consensus are Algorand Accounts not Algorand Nodes. This means you can still have different Participating Accounts on the same Node.

Fair enough and that is good to know. However, I still don't think there should be a global maximum amount because if whales are the first in, they would lock out middle/small holders by causing the cap to get reached quickly.

ARCs/arc-0033.md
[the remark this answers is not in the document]
Fritos2: I also agree, xGov making major decisions should be participating in consensus. This also provides incentive for participating in consensus itself as Ghost has also mentioned (elsewhere)

ARCs/arc-0033.md
Guligus: I would add participating in consensus as a must have for xGov consideration.

ARCs/arc-0033.md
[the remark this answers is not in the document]
joe-p: I am against node participation being a requirement. The primary reason is that I think there are a lot of people that deserve a fair chance at being an xGov and don't have the resources or technical knowledge to run a node. Before you say "Oracle Free Tier", I think putting any significant portion of live stake behind Oracle's free tier is a bad idea for a couple of reasons. 

1. They might just stop the free tier at some point. 
2. If node requirements get raised even a bit all of sudden you have a bunch of stake that can no longer participate. 
3. Oracle could ban nodes at any time (see Hetzner). 
4. If the goal is decentralization, putting a portion of live stake online with a company that has contiously been a bad actor in the FOSS space is a bad idea. 

Also, if only node runners are xGovs they will be incentivized to push through proposals that only benefit node runners, but not necessarily the network as a whole.

ARCs/arc-0033.md
averagezen: > I second this. Consensus particupation is also getting less and less governance rewards since those are only going to DeFi. Giving node runners xGov status might at least incentivize them a little.

I don't think it is necessary to run a node to be an xGov, though I'm not against the idea if there are a number of options available to easily onboard this process. That being said, we don't need this to be a requirement to incentivize node runners. The xGov process can be used to propose funding, and I imagine it very much will be almost immediately if something to that effect hasn't been passed before this is implemented.

ARCs/arc-0033.md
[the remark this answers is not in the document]
joe-p: I am against max commit because that just opens us up to Sybil "attacks" without any sort of KYC. I have no problem with Vendible but I think reliance on a centralized third party for governance is not ideal. We should be trying to think as long term as possible here and who knows where Vendible (or more specifically, the validators) will be in five, ten, fifteen years. 

If anything, perhaps a decent mitigation would have "no xGov partitipation" be a clause in any sort of contractual trade the Foundation makes. Of course, that does account for all of the exhisting whales but I think it's hard to fully get away from the influence of whales in a token-based governance system.

ARCs/arc-0033.md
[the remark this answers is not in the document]
tommythecatdotalgo: Agree with @GhostOfMcAfee. Participation in consensus is critical for the ecosystem and this would be a reasonable expectation of someone who is trying to guide decision making more.

To @joe-p's points on hardware requirements that is a reasonable concern given that the spirit of Algorand is to make participation as low barrier as possible. However, I think the concerns listed for relying on Oracle/IaaS providers is there for consensus in general and not just in the scope of making it an xGov requirement.

I know the Inc has been very conscientious of making changes that would require higher node requirements, and would make requirements like that known ahead of time for consensus participants to react. 

In my mind xGovs are a bit of representative democracy and overall having them need to demonstrate commitment to the network and lead by example would be beneficial to the ecosystem.

ARCs/arc-0033.md
jumperavocado: We need more than a binary Yes/No vote. A few more options like Abstain and Veto could be useful.

ARCs/arc-0033.md
jumperavocado: One thing to keep in mind is that selling/transferring hipothetical xGov tokens might have tax implications for xGovernors.

ARCs/arc-0033.md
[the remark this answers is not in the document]
joe-p: To be clear, I'm not against incentivizing nodes. I think it's a great way to bootstrap networks. I just don't think it should be tied to governance. 

Even with hardware requirements being low, there are still plenty of places in the world where they are hard to meet. In the US there are a lot of places that don't have access to quality internet service. Then look at a country like Nigeria, which is strongly adopting Algorand. The new limitations put in place by the central bank alone likely makes it harder to acquire and maintain the resources to run a node, let alone any other factors in the Nigerian economy. The same is true for many developing markets that *need* blockchain technology and I think it's important to give them equitable access to participate in governance.

ARCs/arc-0033.md
fifthrace: Or instead of uptime you have to vote on consensus X amounts per month. I don't know the odds, but even with 1 ALGO it's likely to happen with some percent chance. This would be much easier to measure I think.

ARCs/arc-0033.md
SilentRhetoric: Unless I am mistaken, the intent here is:
1. For the community to be able to write proposals on which xGovs can vote, and
2. Proposals, if promoted by xGovs to a full governance measure and approved by a majority of all governors, would always be implemented by the Foundation.  

In neither case--proposal writing or approved measure implementation--would this involve Algorand Ventures.  Someone please correct me if I am under the wrong impression.

ARCs/arc-0033.md
SilentRhetoric: > I would also like clarification on whether xGovs are generating these proposals or "designating" them from the community. Can xGov make a proposal at all or is that a conflict of interest?

Having a voting body be able to set its own agenda is a standard governance model used in many contexts.  I cannot fathom how that would be a conflict of interest in any way.

ARCs/arc-0033.md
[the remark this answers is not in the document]
SilentRhetoric: Providing increased voting power in any way would violate the "1 Algo, 1 Vote" principle which seems to be the prevailing model that resonates with people in the absence of a KYC/Proof-of-Humanity "1 Human, 1 Vote" model.  

Additionally, appreciating voting power based on length of service could also run the risk of concentrating power in the hands of a group who may accumulate enough voting power to control voting outcomes without actually owning a majority of xGov stake.  This seems highly problematic to me.

ARCs/arc-0033.md
SilentRhetoric: A maximum would incentivize Sybil-style behavior/attacks, which is opening a can of worms.

ARCs/arc-0033.md
averagezen: > > I would also like clarification on whether xGovs are generating these proposals or "designating" them from the community. Can xGov make a proposal at all or is that a conflict of interest?
> 
> Having a voting body be able to set its own agenda is a standard governance model used in many contexts. I cannot fathom how that would be a conflict of interest in any way.

What I mean is that is not clearly written out that xGovs are allowed to participate in the act of proposal generation. I would assume the case, but I'm not sure from the wording. As to how there could be a conflict of interest, I could think of a few reasons why a particular xGov candidate that may stand to profit directly from a proposal in xGov hands might need to abstain from a vote. It is also possible, depending on how xGovs might be selected for a particular vote, that might unknowingly place themself in an unethical position. This would all depend on exactly how this was structured, in my mind.

SilentRhetoric: 1. The xGov group may overlap significantly with the engaged core of the community, and excluding them from writing proposals would exclude many of the very people whom you would want to be crafting policies and proposals.  That seems antithetical to what we're trying to achieve here.
2. Any proposal written by an xGov, even if entirely self-serving, would need to be approved by a majority of xGovs to be made into a full governance measure and then passed by a majority in a full governance vote.  This is a significant hurdle to clear for someone to benefit from a self-serving proposal.  For this reason, I don't think it would be necessary to take away xGov's ability to write proposals.

ARCs/arc-0033.md
solkimicreb: I think Silvio said something like "the hardest part of blockchain is people". Please don't take the quote seriously, it's from a distant memory.

Anyways, I see why incentives can be a double-edged sword. Currently, the only node running incentive is to secure the system that we are using. The moment we change this creative people will come up with creative ways to profit.

The most common example is the formation of staking pools on other chains with monetary incentives. These pools help bring the online stake up but introduce a new form of centralization.

I am not against node runner incentives but I don't think the consequences can be predicted.

ARCs/arc-0033.md
solkimicreb: @oysterpack This makes sense. Since these are pretty meta "votes on votes" I assume they will have a simple "support" / "not support" binary decision so it will be simple to abstain.

SudoWeezy: After some discussion, we tend to have a threshold/number of votes ( still to be determined and depends on [ARC-34](https://github.com/algorandfoundation/ARCs/pull/151))

We think having 3 choices (Yes/No/Neutral) might be enough.

ARCs/arc-0033.md
SudoWeezy: We can avoid the use of tokens.
It feels like we can store accounts participating as xGov into an Application. (Using Box Storage?)

ARCs/arc-0033.md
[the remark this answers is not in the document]
SudoWeezy: We will not increase the voting power based on the time.

ARCs/arc-0033.md
SudoWeezy: The sentiment tends to be 1Algo = 1 Vote & minimum Algo needed to participate = 1.

ARCs/arc-0033.md
SudoWeezy: We will not have a maximum of algo (This can be abused by splitting Algos into multiple accounts)

ARCs/arc-0033.md
SudoWeezy: Sub-Optimal fashion means not voting enough on proposals during the xGov process (threshold needs to be determined)

ARCs/arc-0033.md
SudoWeezy: Every quarter seems to be the way to go

ARCs/arc-0033.md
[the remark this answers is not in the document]
RonanHerry: I agree that participating in consensus with a node and a high uptime should be one requirement to become xGov. It shows in a still limited way : technical understanding, technical means, and commitment to the protocol.

benbio:
Hey governors, 
Benedetto Biondi here, CEO of Folks Finance.

Thanks to the AF for opening this forum for discussions. At Folks we have been reasoning about a potential pilot program to start and run xGov quickly.

The idea is very straightforward.

Algofi and Folks Finance run a decentralized voting system: 
- https://app.folks.finance/algo-liquid-governance
- https://gov.algofi.org/

FF and Algofi provide DeFi governance participation solutions, so the protocols communities are already directly interested in AF governance measures. 

I recently discussed with John Clarke, Algofi Co-Founder, and we both agree that we could run a pilot program that would work as follows: 

1. Select an xGov manager for each protocol that collects proposals and feedback.
2. Having our internal forum discussion with relative communities.
3. Run a vote on the most relevant proposals on the respective voting system.
4. Approved proposals will be sent to AF for a global communitarian vote. 

This simple solution could be quickly implemented right after the beginning of G6. 
Direct Advantages: 
- decentralizing the AF governance 
- the community would have the chance to be more active
- fast execution

Multiple community proposals could be collected and provided to AF for the global vote, increasing the number of decisions and actions to be taken within the AF governance.

Every proposal given directly from AF will have a certain amount of complaints because it is difficult to manage each single vertical of such a large ecosystem. Indeed, having different communities' votes managed by a single protocol under xGov would remove the AF's responsibility of proposing new measures. 

In our opinion, the AF has the right to evaluate each xGov final proposal before being posted to the global. In case of a clear dangerous xGov-approved proposal, the AF should be denying the global vote. 

We believe that the one-year commitment requirement can be independent then the running of xGov to existing liquid governance solution provider as Folks and Algofi. 

We are looking forward to discussing this further.

ARCs/arc-0033.md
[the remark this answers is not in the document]
robert-zaremba: Agree with @joe-p . Making _running a node_ as a strong requirement would:
* favor technical members
* require a non-technical person to hire some devop guy / or use external service, to run his node

Since there is no minimum amount of Algo required to be a consensus member (or maybe 1 Algo?), I don't think this requirement makes much sense overall. 
On the other hand, it could be a good recommendation to run a consensus node.

ARCs/arc-0033.md
robert-zaremba: The requirement is too vague. It needs more clarity:
* what is a reward?
* what portion of the reward? For how long? 
* Is it going to be returned? When (right after, or with delay -> on chain param)?

ARCs/arc-0033.md
robert-zaremba: > A maximum would incentivize Sybil-style behavior/attacks, which is opening a can of worms.

Only when there is a KYC or a Proof of Humanity. Otherwise , a whale can forge multiple accounts.

ARCs/arc-0033.md
```diff
@@ -0,0 +1,71 @@
...
+#### Example
+Alice participates in governance during Q1 2023
+the potential reward is 5 Algo, Alice:
+1) Commit 100 Algo in an AF program
+2) Express her will to participate in xGov
+3) Will not receive 5 Algo at the end of Q1 2023. They are instead locked in one big xGov pool for 1 year
+4) Will receive 5 Algo at Q1 2024
+ 
+## Rationale
+In this ARC we seek community input on issues such as, but not limited to:
```
robert-zaremba, line 54: Humanity gate should be added. People should be accountable, not bots. If someone wants to stay private, there should be an option for a private humanity check.

ARCs/arc-0033.md
```diff
@@ -0,0 +1,71 @@
...
+- Whether xGovs **MUST** vote on all measures to keep their status
+- Whether xGovs tokens **SHOULD** be sold/transferred or **SHOULD** be burned when exiting the xGov DAO
+- Whether xGovs tokens expire after a year or gain more voting power if the stake remains in the DAO
+- Whether there **SHOULD** be a minimum and a maximum amount of Algos required to become an xGov
+- Whether there be a maximum amount of Algos that can be locked by xGovs for the each period, limiting the total number of xGovs
+- Whether xGovs **SHOULD** be disqualified and lose their status if they behave in a suboptimal fashion
+- What types of suboptimal behavior would trigger disqualification or loss of status?
+- Whether new xGovs be allowed into the pool at every quarter or only once per year
+
+Community members can add their own questions and solutions to the above list
```
uhudo, line 65: Can anyone define what type of governance is Algorand Governance striving to achieve and the role of xGov in it?

Based on past discussions and G3 vote, it seems the sentiment of *1 ALGO = 1 vote* is prevalent, although it was never explicitly ratified. For future use, the term "algocracy" might be appropriate when referring to this form of governance.
But is Algorand Governance striving for a direct/pure or a representative form of "algocracy"?

Current suggestions about xGov seem to be leaning towards a representative form - i.e. with participation through different DAOs as well as due to the different requirements for Governors and xGov (which creates two classes of ALGO with different amounts of power).
With one of the main goals of blockchain being decentralization and equality, a pure form would be more suitable.
A pure form as the basis would still enable formation of representatives of different groups, which could be facilitated by the Foundation (e.g. through the recently established [Governance Advisory Council](https://www.algorand.foundation/news/improving-governance-flows-and-key-dates-for-governance-period-6)) or anyone else (e.g. formation of DAOs per vertical as currently suggested). But the Governance core would be a "pure algocracy", holding onto the same principles as in PPoS consensus.

A possible form of direct participation are initiative referendums, where groups form to bring attention to an initiative and gather initial support for it before being put to a general vote. xGovs could be the entities that start, review and endorse these initiatives, as well as be held accountable - both for the case of malicious/spam initiatives being put to the general vote and the outcomes of the passed initiatives.
A POC implementation of this approach on Algorand can be found at: https://github.com/uhudo/xGov-initiative-referendums

MHO12345:
There should be a few simple rules that can create a consistent and long term pool of committed xGovs:

- min stake: 100k or potentially 50k Algo in order to bring more decentralization;
- min locked up period of stake: 6 months or more, with longer periods bringing higher APRs;
- participate in consensus: altough I see this as a must in the long term, I would not roll it out from the first xGOV session, but include it as being mandatory fron 2nd period onwards and give xGovs time to get up to speed with requirements; 
- participate in all mandatory measures, else not eligible for rewards - but would not include as mandatory to participate in projects/DAOs hand picked by the AF, only whoever wants that for any additional incentive granted by that particular project; 
- last but perhaps the most important aspect, mandatory past governance track record: imo xGovs address should prove that it has successfully participated (voted&rewards claimed) in past governance rounds (at least 2 or 3 past periods);

All these measures would ensure that participants are committed and have a long-term vision for their governor status.

ARCs/arc-0033.md
SudoWeezy: It is a hard commitment of rewards (not the stake used to gain those rewards.)

robdmoore:
This document has a bunch of gendered terms (he, etc.). This needs to be fixed to be gender neutral.

dominagiu (approved):

ARCs/arc-0033.md
```diff
@@ -0,0 +1,71 @@
...
+description: Explanation on how to become Expert Governors.
+author: Stéphane Barroso (@SudoWeezy), Adriana Belotti, Massimo Morini, Michel Treccani, John Woods, Shai Halevi
+discussions-to: https://github.com/algorandfoundation/ARCs/issues/
+status: Draft
+type: Meta
+created: 2022-11-22
+---
+ 
+## Abstract
+xGov status gives a governor the right to vote on proposals raised by the community, to ratify the proposals as valid for final community vote, where said proposals pertain to spending a previously specified amount of Algo in a given epoch on particular initiatives.
```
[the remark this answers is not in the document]
urtho: Opening up the vote to protocol changes might be too early. Optimal protocol (eg vote thresholds) and network (eg number of relays) parameters are a compromise between speed, security and performance. Not sure xGovs will be qualified at this stage to have a say or have resources to support their hypothesis with tests. 

On the other hand there should be a body, in the future, that is able to do that. xGovs could vote to allocate resources for such a think tank(s).

ARCs/arc-0033.md
urtho: Anybody thought of making this not 1:1 or 1 per "person" but a quadratic voting + KYC/SBT to prevent Sybil attacks ?
This would increase the chances of out voting the whales but still give higher weights to larger stake ?

ARCs/arc-0033.md
```diff
@@ -0,0 +1,93 @@
...
+  <tr>
+    <td>Voting Power</td>
+    <td colspan="2">1 Algo commited = 1 Vote</td>
+  </tr>
+  <tr>
+    <td>Duty</td>
+    <td colspan="2">Voting on 90% of proposals&nbsp;&nbsp;(Yes No Neutral)</td>
+  </tr>
+  <tr>
+    <td rowspan="1">Disquallification</td>
```
michaeltchuang, line 50:
```suggestion
    <td rowspan="1">Disqualification</td>
```

typo

ARCs/arc-0033.md
```diff
@@ -0,0 +1,93 @@
...
+
+> The rewards will remain in the xGov reward pools & will be distributed among remaining xGovs
+
+## Rationale
+This proposal provides a clear and simple method for participation in xGov process, while also providing incentives for long-term commitment to the network. Separate pools for xGov and Gov allow for a more diverse range of participation, with the xGov pool providing an additional incentive for longer-term commitment. The requirement to vote on at least 90% of available proposals will ensure that participants are actively engaged in the decision-making process. 
+
+After weeks of engagement with the community, it has been decided:
+- That the xGov process will not utilize token or NFT.
+- There will be no minimum or maximum amount of Algo required to participate in the xGov process
+- In the futur, the possibility of node operation being considered as a form of participation eligibility is being explored
```
michaeltchuang, line 83:
```suggestion
- In the future, the possibility of node operation being considered as a form of participation eligibility is being explored
```

typo

ARCs/arc-0033.md
Niftgen: I agree with this idea

We could use decentralized identity so no one can use multiple wallets. That way people cannot cheat. 

For quadratic voting, perhaps it can be applied if the balance you have in your wallet is above a certain amount ? If we do decentralized identity, then no one would be able to cheat by disbursing their amounts to different wallets.

ARCs/arc-0033.md
daisyifyoud0: We really need to look into Conviction voting or Continuous approval voting.

PR #154, opened 2022-12-19
An alternative xGovernance community proposal

solkimicreb:
The [current xGovernor proposal](https://github.com/algorandfoundation/ARCs/pull/152/files) ties the xGovernor role to participating in certain reward-bearing projects. These projects would be picked by the Foundation, which goes against democratization. Moreover involving any token or project other than Algo itself in the governance of the ecosystem makes us vulnerable to single-project failures.

The goal of this proposal is to rethink these issues and spark a different and parallel discussion to eventually reach a common conclusion.

### DIFF
--- a/ARCs/arc-0033.md
+++ b/ARCs/arc-0033.md
@@ -1,0 +1,120 @@
+---
+arc: 33
+title: xGovernance
+description: Explanation of the xGovernance system.
+author: Miklos Bertalan (@solkimicreb)
+discussions-to: https://github.com/algorandfoundation/ARCs/issues/
+status: Draft
+type: Meta
+created: 2022-12-19
+---
+
+# The xGovernance System
+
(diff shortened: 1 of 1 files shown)


zeus-algo:
I think requiring xGovs to hold a stake is nonsense.

There's no requirement of Foundation or Inc employees to hold a stake.

solkimicreb:
> I think requiring xGovs to hold a stake is nonsense.
> There's no requirement of Foundation or Inc employees to hold a stake.

Hey Zeus! I am glad you are here 🙂 

The proposal mentions these:

- minimum of 100k algo committed to be an xGov
- xGov voting power is weighted by algos committed (like normal governance)
- xGovs can propose one vote per 100k algo committed per governance period

Which of these are you against? All of them?

PS: why the hack can't I just normally reply to this, Github??

PR #156, opened 2023-01-03
ARC-35: Secure Offline Wallet Backup Protocol

yigitguler:


danielalcivarm:
Hi @yigitguler ,I think we should also consider including exporting/importing contacts from the Addressbook and the watched/tracked/followed addresses as well.

yigitguler:
Hi @danielalcivarm I can now confirm that this PR also covers all the contract scenarios using the "name" and "account_type" fields.

ARCs/arc-0035.md
```diff
@@ -0,0 +1,119 @@
...
+    }
+  ],
+}
+```
+
+### Encryption
+
+Once the input JSON is ready as specified above, it needs to be encrypted. Even if it is assumed that the user is going to store this information in a secure location, copy-pasting it without encryption is not secure since multiple applications can access the clipboard. 
+
+The information needs to be encrypted using a very long passphrase. 12 word mnemonic will be used as the key. 12-word mnemonic is secure and it will not create confusion with the 24-word mnemonics that are used to encrypt accounts.
```
taylanpince, line 74: Minor correction: Should be 25-word mnemonics

ARCs/arc-0035.md
```diff
@@ -0,0 +1,119 @@
...
+
+The wallet applications should not allow users to copy the 12-word mnemonic nor allow taking screenshots. The users should note it visually.
```
taylanpince, line 76: After extensive discussion, Pera product team would like this requirement for preventing copy actions removed from ARC-35. What we've seen in the wild is that users will copy text on screen one way or another (users are very resilient against blocks like this). The whole point of ASB is that it makes it safe to copy the 12-word passphrase AND the encrypted backup. That way neither copied item provides direct access to user accounts.

What should be specified here is that the wallets should clearly direct the user to store the 12-word passphrase and the encrypted backup **separately**.

ARCs/arc-0035.md
```diff
@@ -0,0 +1,119 @@
...
+```
+{
+  "device_id": "UNIQUE IDENTIFIER FOR DEVICE (OPTIONAL)",
+  "provider_name": "PROVIDER NAME (OPTIONAL, i.e. Pera Wallet)",
+  "accounts": [
+    {
+      "address": "ACCOUNT PUBLIC ADDRESS (REQUIRED)"
+      "name": "USER DEFINED ACCOUNT NAME (OPTIONAL)",
+      "account_type": "TYPE OF ACCOUNT: single, multisig, watch, contact, ledger (REQUIRED)",
+    "private_key": "ACCOUNT PRIVATE KEY AS BASE64 ENCODING OF 64 BYTE PRIVATE KEY as encoded by NACL/libsodium (NOT PASSPHRASE, REQUIRED for user-owned accounts, can be omitted in case of watch, contact, ledger accounts)"
```
yuli-ferna, line 43: I wanted to confirm one thing regarding multisig accounts, in the JSON format for the private key, it seems to me that it would be a base 64 encoding of the object that has its data, for example:
```
{
        version: 1,
        threshold: 2,
        addrs: [
            account1.addr,
            account2.addr,
            account3.addr,
        ],
};
```
Is this correct?

ARCs/arc-0035.md
```diff
@@ -0,0 +1,119 @@
...
+      "metadata": "ANY ADDITIONAL CONTENT (OPTIONAL)",
```
yuli-ferna, line 44: In the case of the ledger account, where should I include the path or the accountId? As an object in metadata or would a new field be included?

ARCs/arc-0035.md
```diff
@@ -0,0 +1,119 @@
...
+Once the input JSON is ready as specified above, it needs to be encrypted. Even if it is assumed that the user is going to store this information in a secure location, copy-pasting it without encryption is not secure since multiple applications can access the clipboard. 
+
+The information needs to be encrypted using a very long passphrase. 12 word mnemonic will be used as the key. 12-word mnemonic is secure and it will not create confusion with the 24-word mnemonics that are used to encrypt accounts.
+
+The wallet applications should not allow users to copy the 12-word mnemonic nor allow taking screenshots. The users should note it visually. 
+
+The encryption should be made as follows:
+
+1. The wallet generates a random 16-byte string S (using a cryptographically secure random number generator)
+2. The wallet derives a 32-byte key: `key = HMAC-SHA256(key="Algorand export 1.0", input=S)`
```
elranu, line 81: Unfortunately [libsodiumjs](https://github.com/jedisct1/libsodium.js/) does not have exposed the functions: crypto_auth_hmacsha256_init / crypto_auth_hmacsha256_update / crypto_auth_hmacsha256_final. Only has wrapped crypto_auth_hmacsha256 which only supports a fixed key of 32 bytes [here](https://github.com/jedisct1/libsodium/blob/6d566070b48efd2fa099bbe9822914455150aba9/src/libsodium/crypto_auth/hmacsha256/auth_hmacsha256.c#L101).

What can we do instead? A fixed text of 32bytes, in stead of  "`Algorand export 1.0`"? Or add padding to complete the 32bytes? Any other suggestion?

Regards

ARCs/arc-0035.md
yigitguler: @yuli-ferna, this is a great point. Apparently, we will need a new optional field: multi_sig. And it can contain this format. How does it sound?

yigitguler: Formatting in base64 will complicate the debugger's mind for no reason. This way, it can be easier to read & debug.

yigitguler: I updated it. Please review if it looks OK.

ARCs/arc-0035.md
yigitguler: @OEA, what are you using here?  As @elranu states, it may be good to change this text.
cc @fergalwalsh

OEA: I used Go's  `crypto/hmac` and `crypto/sha256` packages here.

This is the example code of generating 32-byte key in Go:

```
func GenerateCipherKey(input []byte) ([]byte) {
	cipher := hmac.New(sha256.New, []byte("Algorand export 1.0"))
	cipher.Write(input)
	return cipher.Sum(nil) 
}
```

ARCs/arc-0035.md
yuli-ferna: Perfect! Thank you :)

ARCs/arc-0035.md
elranu: Ok, so libsodium is not mandatory

yuli-ferna: Ok, and in the case of step 3, using `crypto_secretbox_easy` how do we assign and store the nonce value?
Because nonce can't be null, I see this example in [libsodiumjs](https://github.com/jedisct1/libsodium.js/#api) doc, they concat the nonce value with the ciphertext:

```
let key = sodium.from_hex('724b092810ec86d7e35c9d067702b31ef90bc43a7b598626749914d6a3e033ed');

function encrypt_and_prepend_nonce(message) {
    let nonce = sodium.randombytes_buf(sodium.crypto_secretbox_NONCEBYTES);
    return nonce.concat(sodium.crypto_secretbox_easy(message, nonce, key));
}

function decrypt_after_extracting_nonce(nonce_and_ciphertext) {
    if (nonce_and_ciphertext.length < sodium.crypto_secretbox_NONCEBYTES + sodium.crypto_secretbox_MACBYTES) {
        throw "Short message";
    }
    let nonce = nonce_and_ciphertext.slice(0, sodium.crypto_secretbox_NONCEBYTES),
        ciphertext = nonce_and_ciphertext.slice(sodium.crypto_secretbox_NONCEBYTES);
    return sodium.crypto_secretbox_open_easy(ciphertext, nonce, key);
}
```

Is this the right thing to do, or is another strategy being considered?

ARCs/arc-0035.md
yigitguler: Hi @yuli-ferna, sorry for the late reply.

Wouldn't address field work for this? I may be missing something, please let me know if I misunderstood the accountID.

ARCs/arc-0035.md
OEA: @yuli-ferna As far as I see in this example, it looks like right. Because the nonce value always randomly generated. When you decrypt it, you can slice the nonce and pass it to necessary function.

yuli-ferna: Great! Thank you! In this case, @yigitguler don't you think we should specify how to save the nonce in the doc?

ARCs/arc-0035.md
yuli-ferna: In order to sign a txn via Ledger Nano I need the account index (accountID) to indicate the device which account use to sign. For example:
Account Index #1 reference Account #1
Account Index #2 reference Account #2
Account Index #n reference Account #n

This would be what I mean when I say accountID value, it's a number that would be a separate value from the address.

ARCs/arc-0035.md
yigitguler: Yes, I think it would be great. Could you please add a suggestion so that we can add?

Thank you.

ARCs/arc-0035.md
yigitguler: Yes, now I see. Thank you. @k13n how you are planing to approach this issue?

k13n: hey @yigitguler, @yuli-ferna. We (Defly) store what you call the accountID in the `metadata` object, but refer to that as the `ledger_index`.

Perhaps it'd be a good idea to put together a list of common attributes that all wallets need to know and agree on certain names for them such that wallets can import and handle them appropriately.

Here's a list of attributes that we use in `metadata` and that might be useful for other wallets. Notice that for us these values are nullable but not optional (if an attribute doesn't apply, we keep the key in the object, but set its value to `null`).
- `ledger_connection_type` [int]: the type of the ledger (1=bluetooth/ble, 2=usb)
- `ledger_index` [int]: this is the index that hardware ledgers use to derive an account
- `ledger_device_id` [string]: this is the ID of the ledger
- `msig_signers` [string array]: the list of signers that constitute a multisig
- `msig_threshold`: [int]: the threshold of the multisig
- `msig_version` [int]: the version of multisig

yigitguler: I think it would be good to document these in the ARC specification, otherwise, it can be impossible to parse some types of accounts or wallets may need to write.

@k13n is there a reason why ledger_connection_type is an int? It can be more error-proof if it is a string.

If you receive a backup file without these keys, does the app assumes that these are null or treat this as an error?

k13n: @yigitguler, we just noticed that in the latest version of the ARC there's a `multisig` key in the account object that already contains what we called above the `msig_signers`, `msig_threshold`, and `msig_version`. That means we don't need this data in the `metadata` field. Should we add a similar object for ledger devices, something like this:

```json
{
  "derivation_index": "A LEDGER DEVICE CAN DERIVE MULTIPLE ALGORAND ACCOUNTS, THIS INDEX DENOTES THE N-TH SUCH ACCOUNT (REQUIRED, NUMBER FIELD)",
  "connection_type": "HOW THE LEDGER IS CONNECTED, EITHER bluetooth OR usb (OPTIONAL, STRING)",
  "device_id": "THE ID OF THE LEDGER DEVICE (OPTIONAL, STRING)"
}
```

Here's an example:

```json
{
  "device_id": "2498232091970170817",
  "provider_name": "Defly Wallet",
  "accounts": [
    {
      "address": "ELWRE6HZ7KIUT46EQ6PBISGD3ND6QSCBVWICYR2QR2Y7LOBRZRCAIKLWDE",
      "name": "My NFT Account",
      "account_type": "single",
      "ledger": {
        "device_id": "xyz",
        "index": 1,
        "connection_type": "bluetooth"
      }
    }
  ]
}
```

What do you think?

yigitguler: Sounds great. I am updating the spec if you approve @OEA @miracozkan @yuli-ferna

yigitguler: Waiting for @yuli-ferna's approval to close this issue.

ARCs/arc-0035.md
kewde: Why not use an actual KDF? (e.g HKDF)

scholtz:
hi, i have few points to change

1) can we change the encryption in the way that person can encrypt data using the algorand account, not 12 word secret?

I suggest using Curve25519XSalsa20Poly1305 .. it is used in the vote coin standard, and here for example : https://github.com/scholtz/beaker-arc34-boxes/blob/master/tests/encryption.test.ts

this would be much more easier for people to use as they can create account in destination wallet, encrypt data just with the public key of the account (address) , and export the accounts to the destination wallet 

2) How does it work with different account types than X25529 accounts and multisig? In [AWallet](https://github.com/scholtz/wallet) we have few more types, for example 2FA accounts (2fa api account signs tx at specific api), wallet connect accounts (the wc storage is in encrypted storage and person in one wallet can request signature of other person from different wallet for example for multisig) ...

PR #160, opened 2023-01-13
ARC-0031 : Authentication with Algorand accounts

deanstef:
This PR introduces a refined version of **ARC-0031 Authentication with Algorand accounts** and replaces #84. The PR proposes the following changes/improvements to the old one:
- Allowing signing of **only** random bytes - no transactions;
- Allowing authentication with rekeyed accounts;
- Allowig authentication with multisig accounts.

The ARC also includes a reference implementation (providing a simple authentication process - no rekey/multisig - with MyAlgoWallet) credits to [mrcointreau](https://github.com/mrcointreau)


ARCs/arc-0031.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Abstract
+
+This ARC introduces a standard for authenticating users with their Algorand accounts. It leverages asymmetric encryption <*PK, SK*> to verify the identity of a user, owner of an Algorand account. This approach fosters the adoption of novel identity management systems for both Web3 and Web2 applications.
+
+### Definitions
+
+- **System**: any frontend/backend application, service provider, or in general an entity not based on blockchain;
+- **Credentials**: any type of authentication used by users to access their online accounts, e.g. username/password, PIN, public/secret key pair;
+- **Blockchain identity**: a public/secret key pair <*PK, SK*> representing a blockchain account;
```
fabrice102, line 24: Maybe specify here that multisig are supported too.

ARCs/arc-0031.md
```diff
@@ -0,0 +1,269 @@
...
+- **Algorand account**: a blockchain identity on Algorand identified with the key pair <*PKa, SKa*>;
+- **Algorand address**: the public key *PKa* of an Algorand account;
```
fabrice102, line 26: Technically the address is not exactly the public key as there is additional base32 encoding + checksum. I see the distinction is made clear later.

ARCs/arc-0031.md
```diff
@@ -0,0 +1,269 @@
...
+ /** Unique random nonce generated by the Verifier */
+ nonce: string;
+ /** Optional, description of the Verifier */
+ desc?: string;
+ /** Optional, metadata */
+ meta?: string;
+}
+```
+
+The `auth-msg` **SHOULD** be compliant with [ARC-2](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0002.md), having the parameter `\<dapp-name\>`=`arc31`, for example:
```
fabrice102, line 146: I don't think this is needed. ARC-2 is for notes in transactions.
Since there is no transaction, I think we should use directly the JSON

fabrice102 (commented):
I've made some comments inside the proposal.

But one main comment is that I think that the wallet ARCs (in particular https://arc.algorand.foundation/ARCs/arc-0011) should be extended by this ARC-31 to add a function specific for wallet signature.
The issue with using MyAlgoWallet arbitrary message features is that it train users to sign arbitrary messages.
Since this function does not do any verification, this allows to easily trick users accustomed to use this function into man-in-the-middle attacks.

Instead, a specific wallet function would do all the required checks to make man-in-the-middle attacks less likely.

I think this ARC-31 needs to specify the exact expected interface (similar to how `signTxns` is defined in ARC-1, which includes what the wallet needs to verify precisely and show to the user.
In particular: How is the domain name verify? Is it an exact match? (I think "yes") Do we need to enforce SSL/TLS (I think "yes")? How to handle multiple network, in particular for rekeying? What is the exact returned type to indicate multisig and so on (only the wallet knows about multisig).

What do you think?

ARCs/arc-0031.md
```diff
@@ -0,0 +1,269 @@
...
+The *Authentication Message* is represented with the following interface:
+
+```typescript
+interface AuthMessage {
+ /** The domain name of the Verifier */
+ domain: string;
+ /** Algorand account to authenticate with*/
+ authAcc: string;
+ /** Unique random nonce generated by the Verifier */
+ nonce: string;
```
ehanoc, line 138: Just throwing a consideration here. Perhaps we should use **challenge** instead of nonce. 

There are similar auth standards being developed and proposed ([DIDAuth](https://w3c-ccg.github.io/vp-request-spec/#did-authentication), FIDO [U2F](https://web.archive.org/web/20230416040757/https://fidoalliance.org/specs/fido-u2f-v1.0-ps-20141009/fido-u2f-overview-ps-20141009.html) / [HID](https://web.archive.org/web/20220308142005/https://fidoalliance.org/specs/fido-u2f-v1.0-ps-20141009/fido-u2f-hid-protocol-ps-20141009.html))

Might be a small thing, but operating close to auth standards might be useful with integrations, complaince, etc.

ARCs/arc-0031.md
```diff
@@ -0,0 +1,269 @@
...
+arc31:j{
+ "domain": "www.verifierdomain.com",
+ "authAcc": "KTGP47G64KCXWJS64W7SGJNKTHE37TYDCI64USXI3XOYE6ZSH4LCI7NIDA",
+ "nonce": "1234abcde!%",
+ "desc": "The Verifier",
+ "meta": "arbitrary attached data",
+}
+```
+
+The `nonce` field **MUST** be unique for each authentication and **MUST NOT** be used more than once to avoid replay attacks.
```
aldur, line 158: Minimum recommended length?

0x9090: 15 characters minimum, which must include upper case and lower case letters and numbers. Nonce checking must be case sensitive. Nonce must be derived from a cryptographically strong PRNG. This would give a search space of 7.82 x 10^26 making it impossible to guess

ARCs/arc-0031.md
```diff
@@ -0,0 +1,269 @@
...
+### Assumptions
+
+The standard proposed in this document works under the following assumptions:
+
+- User and Verifier communicate over secure SSL/TLS encrypted channels;
+- The Verifier knows the Users’ *PKa*;
+- For each *PKa* the Verifier generates a unique message to be signed;
+- The message MUST change arbitrarily for each authentication request to avoid [replay attacks](https://en.wikipedia.org/wiki/Replay_attack);
+- User's secret key is safely kept into a Wallet;
+- Users do not change their public address *PKa* for authentication;
```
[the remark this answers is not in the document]
deanstef: Leftover from the previous ARC-13. This ARC now allows rekey

ultrade-dev:
Hope this will get done soon. We definitely could use it.

scholtz:
> Hope this will get done soon. We definitely could use it.

How do you see usage of [ARC14](https://github.com/algorandfoundation/ARCs/pull/218)? Arc14 does not require usage of signdata, but is using the txn.sign to authenticate person, so that it can be used now with any wallet connect capable wallets.

emg110:
The living web standard, WebAuthn (Web Authentication API) supports the ed25519 elliptic curve and is also verifiable using other web standard API (Web Crypto through Crypto.subtle methods) or TweetNACL if that's your thing!
The process is clear and standard and also comprehensive and extendible (WebAuthn Extension points)! It is some how a superset of many standards now: FIDO Alliance, U2F, PassKey, Passwordless and many other unified under or derived from WebAuthn somehow (although backward compatibility with some much older versions is Iffy)!

The living standard, WebAuthn (Web Authentication API) supports the ed25519 elliptic curve and is also verifiable using other web standard API (Web Crypto through Crypto.subtle methods) or TweetNACL if that's your thing!
The process is clear and standard is comprehensive and extendible (WebAuthn Extension points)!

Very nice extensions like appId extension, User Verification Method extension and Credentials properties extension make it possible to add what ever data, metadata or procedure elements needed under Algorand scope and still comply to standard!

It has separated the two major concerns: Attestation and Assertation so that devs are free to implement limitless scenarios of identity and AAA using WebAuthn.

https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API

https://www.w3.org/TR/webauthn-2/#sctn-intro

Just a humble suggestion, that's all!

emg110:
I humbly believe that using the composability of WebAuthn and Algorand (ASCs, ASAs and ATs plus Accounts) with a little team effort, a very well-structured identity and AAA framework can emerge that using federated identities can transparently and natively bridge Algorand blockchain identities to Web 2.0 identities via OAuth, Auth0, SAML,...

ARCs/arc-0031.md
```diff
@@ -0,0 +1,269 @@
...
+- `nonce`: a unique/random value generated by the Verifier;
+- `description`: Verifier details or general description;
+- `metadata`: arbitrary message data.
+
+The *Authentication Message* is represented with the following interface:
+
+```typescript
+interface AuthMessage {
+ /** The domain name of the Verifier */
+ domain: string;
```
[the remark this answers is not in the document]
0x9090: We should do exact protocol + domain + port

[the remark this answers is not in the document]
deanstef: @0x9090 as specified in the assumptions, the protocol MUST be `https`; port cannot be different from 443.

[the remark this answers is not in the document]
pbennett: Careful - https is just the scheme.  You can still connect via https on other ports. Requiring TLS is one thing via HTTPS:[...] but the port is independent and shouldn't be prevented from being part of the URI.  8443 is quite common as well as there being arbitrary ports that might be used in development, testing, etc.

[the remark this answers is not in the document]
deanstef: Thank you for clarifying, Patrick. 

What I meant is that in the most cases port will be 443 (it is the most common port for web browsing data transmission). Certainly, other ports can be used over HTTPS, and I agree that we should not prevent them from being part of the URI. As long as we require secure communication channels, we should be good.

emg110:
Given the status of ARC60 and also the fact that essence of this ARC (the arbitrary data signing and aside from process and flow) are there and in order to avoid redundant standardization, and also given the ARC80 which includes exactly the neccessary process and flow which complements ARC60, I propose to retire this ARC in favor of ARC60+ARC80 and given the fact that it is a bit too opinionated to be used with known third party Authentication frameworks like Oauth and OIDC (on top of Oauth) out of the box.
I also suggest all discussed best practices (rekeyed accounts, multisig support,...) to be moved from this to ARC60 to make it most efficient and comprehensive.
Briefly : ARC60 + ARC80 make a complete solution for authentication on Algorand which is 100% compatible with all known web standard frameworks and IETF RFCs.

cusma:
@deanstef, @mrcointreau, @ehanoc: I’m closing this PR because it has become stale, having been inactive for over two years.

PR #162, opened 2023-01-19
ARCs for smart contract NFTs and interface detection

willghatch:
This PR includes two ARCs.  I decided to put them together because one refers to the other (specifically, the NFT ARC depends on the interface detection ARC).

The interface detection ARC is nearly identical to ERC-165 with minor adjustments for Algorand.

The NFT ARC is similar to ERC-721.  Besides minor changes to fit Algorand, the most notable change is moving the transfer management features out of the core spec and into an extension.  This makes the core interface significantly simpler.

The biggest drawback of this NFT ARC is that it relies on ARC-22 `read-only` methods, which are not yet well supported by `algod` and indexers to view data without publishing an actual transaction.  The most important views of an NFT are `ownerOf` and `metadataURI`.  The `ownerOf` method provides data that is redundant with the `Transfer` event, so without support for free calling of ARC-22 `read-only` methods, an indexer can still get this info.  I've considered adding an extra a`metadataURIEvent` extension for indexers to have an easy way to index metadata as well for NFTs created while ARC-22 support is still poor.

### DIFF
--- a/ARCs/arc-draft_interface-detection.md
+++ b/ARCs/arc-draft_interface-detection.md
@@ -1,0 +1,86 @@
+---
+arc: <to be assigned>
+title: Algorand Interface Detection Spec
+description: A specification for smart contracts and indexers to detect interfaces of smart contracts.
+author: William G Hatch <william@hatch.uno>
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category : Interface
+created: 2023-01-10
+requires: 4, 22, 28
+---
+
+# Algorand Interface Detection Spec
+
+## Abstract
+
+This ARC specifies an interface detection interface based on <a href="https://eips.ethereum.org/EIPS/eip-165">ERC-165</a>.
+This interface allows smart contracts and indexers to detect whether a smart contract implements a particular interface based on an interface selector.
+
+
+## Motivation
+
+ARC-4 applications have associated Contract or Interface description JSON objects that allow users to call their methods.
+However, these JSON objects are communicated outside of the consensus network.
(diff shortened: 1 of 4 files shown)


willghatch:
This NFT ARC could possibly subsume the current ARC-72 PR.  ARC-72 is similarly an NFT spec based on ERC-721, but is currently not as fleshed out as this one.

ARCs/arc-draft_interface-detection.md
```diff
@@ -0,0 +1,86 @@
+---
+arc: <to be assigned>
```
SudoWeezy, line 2: You can use the ARC-72, I will close the other PR

KillerGasy:
> We will also be submitting a spec for an NFT Index interface to pair with smart contract NFTs. People should not be hitting the chain for anything other than verification and/or state change.

Will this work with conduit?

cswenor:
> > We will also be submitting a spec for an NFT Index interface to pair with smart contract NFTs. People should not be hitting the chain for anything other than verification and/or state change.
> 
> Will this work with conduit?

We are only submitting the interface, people can do whatever they want to deliver the interface.

ARCs/arc-draft_interface-detection.md
willghatch: Well, this PR has two ARCs.  Maybe I should have separated them, but I wanted to submit them at the same time because they are to some degree related.  Should I use ARC-72 for interface detection and something else for the NFT spec?

SudoWeezy: In that case, you can use 72 for NFT-ARC & 73 for interface detection

willghatch:
Superseded by #164 and #165.

PR #164, opened 2023-01-26
add NFT standard draft arc-72

willghatch:
This is the NFT standard from #162 separated from interface detection.

### DIFF
--- a/ARCs/arc-0072.md
+++ b/ARCs/arc-0072.md
@@ -88,7 +88,8 @@ Additionally, the smart contract MUST implement the following interface:
 
 Ownership of a token ID by the zero address indicates that ID is invalid.
 The `ownerOf` method MUST return the zero address for invalid token IDs.
-The `transferFrom` method MUST error when `from` is not the owner of `tokenId`, unless called by an approved operator as defined by an extension such as the transfer management extension defined in this ARC.
+The `transferFrom` method MUST error when `from` is not the owner of `tokenId`.
+The `transferFrom` method MUST error unless called by the owner of `tokenId` or an approved operator as defined by an extension such as the transfer management extension defined in this ARC.
 The `transferFrom` method MUST emit a `Transfer` event a transfer is successful.
 A `Transfer` event SHOULD be emitted, with `from` being the zero address, when a token is first minted.
 A `Transfer` event SHOULD be emitted, with `to` being the zero address, when a token is destroyed.
--- a/ARCs/arc-0072.md
+++ b/ARCs/arc-0072.md
@@ -1,0 +1,341 @@
+---
+arc: 72
+title: Algorand Smart Contract NFT Specification
+description: Base specification for non-fungible tokens implemented as smart contracts.
+author: William G Hatch <william@hatch.uno>
+discussions-to: https://github.com/algorandfoundation/ARCs/pull/162
+status: Draft
+type: Standards Track
+category: Interface
+created: 2023-01-10
+requires: 3, 4, 22, 28, 73
+---
+
+# Algorand Smart Contract NFT Specification
+
+## Abstract
+This specifies an interface for non-fungible tokens (NFTs) to be implemented on Algorand as smart contracts.
+This interface defines a minimal interface for NFTs to be owned and traded, to be augmented by other standard interfaces and custom methods.
+
+
+## Motivation
+Currently most NFTs in the Algorand ecosystem are implemented as ASAs.
+However, to provide rich extra functionality, it can be desireable to implement NFTs as a smart contract instead.
+To foster an interoperable NFT ecosystem, it is necessary that the core interfaces for NFTs be standardized.
+
+
+## Specification
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in <a href="https://www.ietf.org/rfc/rfc2119.txt">RFC-2119</a>.
+
+
+### Core NFT specification
+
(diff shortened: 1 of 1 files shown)


ARCs/arc-0072.md
```diff
@@ -0,0 +1,341 @@
...
+        }
+      ]
+    }
+  ]
+}
+```
+
+Ownership of a token ID by the zero address indicates that ID is invalid.
+The `ownerOf` method MUST return the zero address for invalid token IDs.
+The `transferFrom` method MUST error when `from` is not the owner of `tokenId`, unless called by an approved operator as defined by an extension such as the transfer management extension defined in this ARC.
```
AustP, line 91: This doesn't prevent others from transferring my NFTs. As long as they call the method and specify my address as the first argument, then they could transfer it. What we care about is that the entity *calling* the `transferFrom` method is either the owner or an approved operator.

willghatch:
On Mon, Feb 06, 2023 at 08:32:52PM -0800, Aust wrote:
>@AustP commented on this pull request.
>
>
>
>> +          "desc": "The new owner of the NFT"
>+        },
>+        {
>+          "type": "uint256",
>+          "name": "tokenId",
>+          "desc": "The ID of the transferred NFT"
>+        }
>+      ]
>+    }
>+  ]
>+}
>+```
>+
>+Ownership of a token ID by the zero address indicates that ID is invalid.
>+The `ownerOf` method MUST return the zero address for invalid token IDs.
>+The `transferFrom` method MUST error when `from` is not the owner of `tokenId`, unless called by an approved operator as defined by an extension such as the transfer management extension defined in this ARC.
>
>This doesn't prevent others from transferring my NFTs. As long as they call the method and specify my address as the first argument, then they could transfer it. What we care about is that the entity *calling* the `transferFrom` method is either the owner or an approved operator.

Oops, you're right.  It's fixed now.  I was getting ahead of myself
when writing that thinking about two different error conditions but
putting them together.

ARCs/arc-0072.md
```diff
@@ -0,0 +1,342 @@
...
+    }
+  ],
+}
+```
+
+URIs shorter than the return length MUST be padded with zero bytes at the end of the URI.
+The token URI returned SHOULD be an `ipfs://...` URI so the metadata can't expire or be changed by a lapse or takeover of a DNS registration.
+The token URI SHOULD NOT be an `http://` URI due to security concerns.
+The URI SHOULD resolve to a JSON file following the JSON Metadata File Schema defined in [ARC-3](./arc-0003.md).
+Future standards could define new recommended URI or file formats for metadata.
```
AustP, line 128: This doesn't say what should happen if an invalid tokenId is given. Should we add that into the standard? I imagine people will either cause it to error or return all 0s. It might be nice to enforce one way.

ARCs/arc-0072.md
```diff
@@ -0,0 +1,342 @@
...
+}
+```
+
+The `Approval` event MUST be emitted when the `approve` method is called successfully.
+The zero address for the `approve` method and the `Approval` event indicate no approval, including revocation of previous single NFT controller.
+When a `Transfer` event emits, this also indicates that the approved address for that NFT (if any) is reset to none.
+The `ApprovalForAll` event MUST be emitted when the `setApprovalForAll` method is called successfully.
+The contract MUST allow multiple operators per owner.
+The `transferFrom` method, when its `nftId` argument is owned by its `from` argument, MUST succeed for when called by an address that is approved for the given NFT or approved as operator for the owner.
+
```
AustP, line 234: This section needs to specify under what conditions an operator can be set. For example, the owner has to be the one calling. But also, can an operator for a single NFT designate a new operator? Can an operator for a user add more operators?

AustP:
Another thing that seems a little odd is that the spec uses `uint256` as the type for integers. `tokenId`, `balanceOf`, etc. I know that probably comes from Ethereum, but Algorand's integers are `uint64`. It would be a lot more convenient to build/interface with if `uint64` was used instead. That would enable us to use opcodes/methods that worked with integers.

ARCs/arc-0072.md
SudoWeezy: If you provide an Invalid TokenId, it will not work, and you will have an error at the app call level. (Key does not exist inside the Box)

SudoWeezy:
> Another thing that seems a little odd is that the spec uses `uint256` as the type for integers. `tokenId`, `balanceOf`, etc. I know that probably comes from Ethereum, but Algorand's integers are `uint64`. It would be a lot more convenient to build/interface with if `uint64` was used instead. That would enable us to use opcodes/methods that worked with integers.

Based on ERC-721
`The choice of uint256 allows a wide variety of applications because UUIDs and sha3 hashes are directly convertible to uint256.`

I agree it is not convenient yet to use uint256, but we are discussing how it can be implemented in an upcoming release internally. (same for uint128 - 512)

ARCs/arc-0072.md
SudoWeezy: We are only designing the ABI and not the logic behind a specific NFT-Application

ARCs/arc-0072.md
AustP: Sure but you can check if the box exists before accessing it. In fact, we have to do that for the `ownerOf` method call. The spec requires the zero address be returned rather than erroring out. It seems reasonable that someone could follow that pattern here and return a zero byte-string for the URI rather than erroring out. That's why I wondered if we should add into the spec what should happen if the `tokenId` is invalid and remove the ambiguity.

ARCs/arc-0072.md
AustP: I agree we need to be careful about getting too specific. However, if a contract implemented this spec as-is, then I could set myself as the operator for every NFT because there is nothing in the spec requiring authorization from current owners and/or operators. So I think something should be added into the spec to prevent that.

PR #165, opened 2023-01-26
add interface detection draft arc-73

willghatch:
This is the interface detection standard from #162 separated out into its own PR.

### DIFF
--- a/ARCs/arc-0073.md
+++ b/ARCs/arc-0073.md
@@ -38,6 +38,12 @@ This specification extends ARC-4 to define the concept of an interface selector.
 We define the interface selector as the XOR of all selectors in the interface.
 Selectors in the interface include selectors for methods, selectors for events as defined by [ARC-28](./arc-0028.md), and selectors for potential future kinds of interface components.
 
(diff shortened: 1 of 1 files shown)


ARCs/arc-0073.md
```diff
@@ -0,0 +1,86 @@
...
+    }
+  ]
+}
+```
+
+The `supportsInterface` method must be `readonly` as specified by [ARC-22](./arc-0022.md).
+
+The implementing contract must have a `supportsInterface` method that returns:
+
+* `true` when `interfaceID` is `0x4e22a3ba` (the selector for ARC-73, this interface)
```
SudoWeezy, line 68: Could you specify how the identifier was calculated?

willghatch: I've added some text explaining the algorithm.  Does this suffice?  Where are Enum functions defined?  I can't find any instances of “enum” when grepping through the ARC repository.  The method selector is the same as in ARC-4 (and also extended by ARC-28 for events), and the interface selector is just the XOR of all method selectors.

PR #168, opened 2023-02-16
ubigint alias for byte[]

joe-p:
Adds a type alias `biging` for `byte[]` to represent variable-width integers

### DIFF
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -531,7 +531,7 @@ For encoding purposes, the types are divided into two categories: static and dyn
 The dynamic types are:
 
 *  `<type>[]` for any `type`
-    * This includes `string` and `bigint` types since they are an alias for `byte[]`.
+    * This includes `string` and `bigint` types since they are aliases for `byte[]`.
 * `<type>[<N>]` for any dynamic `type`
 * `(T1,T2,...,TN)` if `Ti` is dynamic for some `1 <= i <= N`
 
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -516,7 +516,7 @@ any leading zeros, _unless_ `N` is zero, in which case only a single 0 character
 * `address`: Used to represent a 32-byte Algorand address. This is equivalent to `byte[32]`.
 * `<type>[]`: A variable-length array. `type` can be any other type.
 * `string`: A variable-length byte array (`byte[]`) assumed to contain UTF-8 encoded content.
-* `bigint`: A variable-length byte array (`byte[]`) assumed to contain a variable-width unsigned integer.
+* `ubigint`: A variable-length byte array (`byte[]`) assumed to contain a variable-width unsigned integer.
 * `(T1,T2,…,TN)`: A tuple of the types `T1`, `T2`, …, `TN`, `N >= 1`.
 * reference types `account`, `asset`, `application`: **MUST NOT** be used as the return type.
 For encoding purposes they are an alias for `uint8`. See section "Reference Types" below.
@@ -531,7 +531,7 @@ For encoding purposes, the types are divided into two categories: static and dyn
 The dynamic types are:
 
 *  `<type>[]` for any `type`
-    * This includes `string` and `bigint` types since they are aliases for `byte[]`.
+    * This includes `string` and `ubigint` types since they are aliases for `byte[]`.
 * `<type>[<N>]` for any dynamic `type`
 * `(T1,T2,...,TN)` if `Ti` is dynamic for some `1 <= i <= N`
 
@@ -578,9 +578,11 @@ For any ABI value `x`, we recursively define `enc(x)` to be as follows:
     * `enc(x) = enc(x * 10^M)`, where `x * 10^M` is interpreted as a `uint<N>`.
 * If `x` is a boolean value `bool`:
     * `enc(x)` is a single byte whose **most significant bit** is either 1 or 0, if `x` is true or false respectively. All other bits are 0. Note: this means that a value of true will be encoded as `0x80` (`10000000` in binary) and a value of false will be encoded as `0x00`. This is in contrast to most other encoding schemes, where a value of true is encoded as `0x01`.
+* If `x` is a unsigned variable-width integer, `ubigint` 
+  * `enc(x)` is encoded as `byte[]` containing the big-endian encoding of `x`. `enc(x)` SHOULD be as succint as possible without 0-padding beyond the highest MSB set.
 
 Other aliased types' encodings are already covered:
-- `string` and `bigint` are aliases for `byte[]`
+- `string` is an alias for `byte[]`
 - `address` is an alias for `byte[32]`
 - `byte` is an alias for `uint8`
 - each of the reference types is an alias for `uint8`
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -516,6 +516,7 @@ any leading zeros, _unless_ `N` is zero, in which case only a single 0 character
 * `address`: Used to represent a 32-byte Algorand address. This is equivalent to `byte[32]`.
 * `<type>[]`: A variable-length array. `type` can be any other type.
 * `string`: A variable-length byte array (`byte[]`) assumed to contain UTF-8 encoded content.
+* `bigint`: A variable-length byte array (`byte[]`) assumed to contain a variable-width unsigned integer.
 * `(T1,T2,…,TN)`: A tuple of the types `T1`, `T2`, …, `TN`, `N >= 1`.
 * reference types `account`, `asset`, `application`: **MUST NOT** be used as the return type.
 For encoding purposes they are an alias for `uint8`. See section "Reference Types" below.
@@ -530,7 +531,7 @@ For encoding purposes, the types are divided into two categories: static and dyn
 The dynamic types are:
 
 *  `<type>[]` for any `type`
-    * This includes `string` since it is an alias for `byte[]`.
+    * This includes `string` and `bigint` types since they are an alias for `byte[]`.
 * `<type>[<N>]` for any dynamic `type`
 * `(T1,T2,...,TN)` if `Ti` is dynamic for some `1 <= i <= N`
 
@@ -579,7 +580,8 @@ For any ABI value `x`, we recursively define `enc(x)` to be as follows:
     * `enc(x)` is a single byte whose **most significant bit** is either 1 or 0, if `x` is true or false respectively. All other bits are 0. Note: this means that a value of true will be encoded as `0x80` (`10000000` in binary) and a value of false will be encoded as `0x00`. This is in contrast to most other encoding schemes, where a value of true is encoded as `0x01`.
 
 Other aliased types' encodings are already covered:
-- `string` and `address` are aliases for `byte[]` and `byte[32]` respectively
+- `string` and `bigint` are aliases for `byte[]`
+- `address` is an alias for `byte[32]`
 - `byte` is an alias for `uint8`
 - each of the reference types is an alias for `uint8`
 


ARCs/arc-0004.md
```diff
@@ -516,6 +516,7 @@ any leading zeros, _unless_ `N` is zero, in which case only a single 0 character
 * `address`: Used to represent a 32-byte Algorand address. This is equivalent to `byte[32]`.
 * `<type>[]`: A variable-length array. `type` can be any other type.
 * `string`: A variable-length byte array (`byte[]`) assumed to contain UTF-8 encoded content.
+* `bigint`: A variable-length byte array (`byte[]`) assumed to contain a variable-width unsigned integer.
```
jasonpaulos, line 519: Perhaps this should be named `ubigint` to indicate that it's unsigned?

At some point in the future we may want to support signed ints, so it's good practice to avoid ambiguity.

ARCs/arc-0004.md
```diff
@@ -579,7 +580,8 @@ For any ABI value `x`, we recursively define `enc(x)` to be as follows:
     * `enc(x)` is a single byte whose **most significant bit** is either 1 or 0, if `x` is true or false respectively. All other bits are 0. Note: this means that a value of true will be encoded as `0x80` (`10000000` in binary) and a value of false will be encoded as `0x00`. This is in contrast to most other encoding schemes, where a value of true is encoded as `0x01`.
 
 Other aliased types' encodings are already covered:
-- `string` and `address` are aliases for `byte[]` and `byte[32]` respectively
+- `string` and `bigint` are aliases for `byte[]`
```
jasonpaulos, line 583: Two comments about encoding that I think should be in the spec:
* Bigints are encoded big-endian
* Variable-sized bigints don't have a single unique encoding. It's always possible to add more 0s to the beginning and encode a number as a larger byte string. The spec needs to decide if this should be allowed, or if only the most succinct encoding can be used.

jasonpaulos (commented):
I think it's in our best interest to properly finalize this ARC, so I'm inclined to say changes like this and #169 would be better suited in a different ARC which extends ARC-4.

Other than that, I left comments on some specifics.

ARCs/arc-0004.md
joe-p: Yeah good point. Fixed in 89f3d51

ARCs/arc-0004.md
joe-p: In 89f3d51 I specified big-endia encoding. I also addressed 0-padding and stated that `enc(x)` SHOULD be as succinct as possible.

joe-p:
@jasonpaulos The reason for these PRs now is because I think they would be good to have in the finalized ARC-0004. The primary reason being they directly map to values that are commonly used in TEAL/AVM. For example, `timestamp` relates to `global LatestTimestamp` and `ubigint` is essentially how big-endian numbers are already treated by the byte arithmetic opcodes in the AVM since there is no overflow panic.

FrankSzendzielarz:
> I think it's in our best interest to properly finalize this ARC, so I'm inclined to say changes like this and #169 would be better suited in a different ARC which extends ARC-4.
> 
> Other than that, I left comments on some specifics.

In the recent meeting with @SudoWeezy recently we agreed to close my larger PR (which contained the Datetime and Biginteger requests) in favour of making these smaller PRs in order to finalise ARC4.

Biginteger is a glaring omission from ARC4 at the moment as you cannot explicitly declare bigintegers without speciying uint512 , which would mean padding to 512 bits for arguments and return values.

FrankSzendzielarz:
Oh...didn't we also say we were going to limit UINTN to UINT8,16,32,64..?

SudoWeezy:
> Oh...didn't we also say we were going to limit UINTN to UINT8,16,32,64..?
A lot of dApps already use UINT > 64, so I don't think it is an option to reduce what's already there

FrankSzendzielarz:
> > Oh...didn't we also say we were going to limit UINTN to UINT8,16,32,64..?
> > A lot of dApps already use UINT > 64, so I don't think it is an option to reduce what's already there

You are missing the point. Uint 24/72/etc

SudoWeezy:
So uint for ^2 & ubigint otherwise?

FrankSzendzielarz:
yes , uints that work with the basic compiler types for fixed size integers, eg byte, short, int32, long, and possibly a uint512 to represent a fixed size teal bigint type I guess, and , yes, ubigint for the TEAL variable-length byte array integer type. Though that's my suggestion, my original question was that I thought we agreed that would be a change? But I don't recall exactly.

SudoWeezy:
I listened to the record again, indeed we discussed this at the end but no real agreement.

emg110:
Great idea but since this list can go on forever , how about another approach to make ABI able to accept user (dev) defined aliases through ABI specification JSON as they see fit? This way instead of having N PRs , we will need one forever!
Example:
`
{...
desc:"This ABI interface contains custom dynamic and configurable aliases"

aliases: [timestamp, my_nice_alias],

aliasTypes:[uint64, byte[]]
`

joe-p:
> Oh...didn't we also say we were going to limit UINTN to UINT8,16,32,64..?

I still don't fully understand the need for this. Can you expound upon why other sizes are problematic? Higher level languages might not natively support these types, but the important thing is the wire encoding.

> Great idea but since this list can go on forever , how about another approach to make ABI able to accept user (dev) defined aliases through ABI specification JSON as they see fit? This way instead of having N PRs , we will need one forever!

To me the benefit of aliases like `string`, `ubigint`, and `timestamp` is that is provides context to potential clients for how they should be encoded/decoded from/to native types. For example, if a method returns `String` the client knows it's UTF-8 encoded. If it's just `byte[]`, the client won't know this and will just return an array of bytes.

FrankSzendzielarz:
> > Oh...didn't we also say we were going to limit UINTN to UINT8,16,32,64..?
> 
> I still don't fully understand the need for this. Can you expound upon why other sizes are problematic? Higher level languages might not natively support these types, but the important thing is the wire encoding.

Because it complicates everything. The main point of ARC4 is not just the wire encoding but being able to specify Contract Interfaces so that you can generate clients on the basis of a contract.json.

If you send a contract.json to someone that uses types like uint48 then you have the additional complexity of having to a) pad results to uint64 to clients or b) truncate arguments from uint64 to the ABI method. Also you need compiler decorators on the field and property declarations saying ``[ABIByteLimit(48)]`` etc.

This is not a fictional issue. It's a major concern.

FrankSzendzielarz:
> 

Yeah. Saying ``string`` is alias for ``byte[]`` is a bit suss. It says it's a UTF8 encoded byte array, or should be. Same with bigint being an alias for bytes..... why would ufixed not be an alias?

joe-p:
> If you send a contract.json to someone that uses types like uint48 then you have the additional complexity of having to a) pad results to uint64 to clients or b) truncate arguments from uint64 to the ABI method. Also you need compiler decorators on the field and property declarations saying [ABIByteLimit(48)] etc.

I question whether padding/truncating binary data for a compiler or client is really that difficult of a task. If the language doesn't support uint48 natively, then convert the number to binary data and truncate accordingly. Doing this in the AVM is trivial and the same for all uintN types.

FrankSzendzielarz:
> > If you send a contract.json to someone that uses types like uint48 then you have the additional complexity of having to a) pad results to uint64 to clients or b) truncate arguments from uint64 to the ABI method. Also you need compiler decorators on the field and property declarations saying [ABIByteLimit(48)] etc.
> 
> I question whether padding/truncating binary data for a compiler or client is really that difficult of a task. If the language doesn't support uint48 natively, then convert the number to binary data and truncate accordingly. Doing this in the AVM is trivial and the same for all uintN types.

It's certainly not trivial. You would need to add or declare custom attributes over the main integer types that describe how they should be marshalled.

Well, it's not a *major* task but it does mean that auto-generated clients/proxies will need to take care of all that. 

For example, sending a uint64 C# or Java type to a smart contract will need :
a) overflow checks to see if the truncation is allowed
b) exceptions that are specific to this kind of serialisation error
c) etc

Ok when we do send uint48 to the contract, what do we expect compilers to do with mathematical operations on them? 

It's a pain in the ass. It's easier, more pragmatic, more reasonable, and overall better if we just support basic compiler types :

OR if we relegate wire encoding to types like : byte[]/ uint or arrays thereof and leave the HLL types to arc32

joe-p:
> For example, sending a uint64 C# or Java type to a smart contract will need :

But most high-level languages don't support uint8, 16, etc. either so the logic needed for uint8 support is the exact same as the logic needed for uint48 support. 

> a) overflow checks to see if the truncation is allowed

To me this isn't something that should be covered by the ABI. We are only talking about encoding/decoding here. Some implementations might want to panic on overflow some might not.

FrankSzendzielarz:
> > For example, sending a uint64 C# or Java type to a smart contract will need :
> 
> But most high-level languages don't support uint8, 16, etc. either so the logic needed for uint8 support is the exact same as the logic needed for uint48 support.
> 
> > a) overflow checks to see if the truncation is allowed
> 
> To me this isn't something that should be covered by the ABI. We are only talking about encoding/decoding here. Some implementations might want to panic on overflow some might not.

No, we are not talking about encoding/decoding only. We are talking about ARC4. Which makes the mistake of conflating wire encoding with a type system for the runtime and the contract.json client specs.

I really don't see where the resistance is coming from. It's obvious the UINTN idea was an oversight or over-reach. Just change the spec!! 

Anyway, I am kind of done thinking about this. I have got things to do, so until this ARC4 thing is sorted out, I will proceed with tentative support for elements of it.

joe-p:
> No, we are not talking about encoding/decoding only. We are talking about ARC4. Which makes the mistake of conflating wire encoding with a type system for the runtime and the contract.json client specs.

ARC4 is specifically about encoding/decoding. Current HLLs represent runtime types in accordance with ARC4 because that is the only standard we have, but a contract can be 100% ARC4 compliant while having a completely unique way of handling runtime logic and representation of these types. ARC4 only specifies the encoding of types for input arguments and return values, but nothing in between.

I agree that a standard should be made for handling different types during runtime, but that is a separate matter.

FrankSzendzielarz:
> > No, we are not talking about encoding/decoding only. We are talking about ARC4. Which makes the mistake of conflating wire encoding with a type system for the runtime and the contract.json client specs.
> 
> ARC4 is specifically about encoding/decoding. Current HLLs represent runtime types in accordance with ARC4 because that is the only standard we have, but a contract can be 100% ARC4 compliant while having a completely unique way of handling runtime logic and representation of these types. ARC4 only specifies the encoding of types for input arguments and return values, but nothing in between.
> 
> I agree that a standard should be made for handling different types during runtime, but that is a separate matter.

I also agree that something else should be constructed for dealing with runtime representations.

However, until then, ARC4 *specifies runtime types*!! Contract.json specifies types that your client must be able to convert to/from. Your high level language must be able to work with them. Your arrays and your structures must map to them and then you are forced to decide if to a) decode into something more useful b) work with the encoding directly.

I really don't understand where or why the dogma/resistance is coming from.

With the pragmatic aim of supporting the kind of activities in the paragraphs above, just *change the spec*. And if you really want to have a separate system of runtime types, then get the core devs to start working on it now. I've had PRs talking about datetime, bigint and god knows what else for almost a year.

joe-p:
> until then, ARC4 specifies runtime types!!

Just because there isn't an existing standard for runtime types doesn't mean we should expand the scope of ARC4 to cover runtime types.

>  decide if to a) decode into something more useful b) work with the encoding directly.

This is exactly my point. You can *decide* how want you want to do with the data after decoding and still be compliant with ARC4. What happens after decode and/or before encode is not within the scope of this ARC and expanding the scope to cover that opens an entirely different can of worms, which is why it makes sense to have runtime types be a separate ARC.

FrankSzendzielarz:
> 

We certainly agree that we need some method of abstracting the wire encoding away in the AVM ("runtime types"). 
 
I the other stuff is just semantics and points of view. I think until there is such a mechanism then the ABI specifies "ABI types" that for most pragmatic purposes have to be used as is. As a compiler, types like uint24 are either unsupported in smart contract hll definitions, or they are used as ways of limiting uint32 value range using some kind of marshalling annotation. 

Regardless though, it's clear we need some way of abstracting wire encoding in the AVM, and because I think it will take a long, long time for that to happen, I think it would be generally *helpful* if the uintN spec was limited to the classic powers of 2.

SudoWeezy:
Closing this PR and the issue is now addressed here: https://github.com/algorandfoundation/ARCs/pull/199

PR #169, opened 2023-02-16
timestamp alias for uint64

joe-p:
`timestamp` alias for `uint64` for representing a unix epoch timestamp. 

Useful when inputting/returning data that interacts with `global LatestTimestamp`


emg110:
Great idea but since this list can go on forever , how about another approach to make ABI able to accept user (dev) defined aliases through ABI specification JSON as they see fit? This way instead of having N PRs , we will need one forever!
Example:
`
{...
desc:"This ABI interface contains custom dynamic and configurable aliases"

aliases: [timestamp, my_nice_alias],

aliasTypes:[uint64, byte[]]
`

SudoWeezy:
Closing this PR and the issue is now addressed here: https://github.com/algorandfoundation/ARCs/pull/199

PR #181, opened 2023-03-29
ARC-4: Update ClearState behavior and revert change to tuple limits

jasonpaulos:
This PR makes two changes to ARC-4:
1. Revises how the ClearState OnCompletion action is handled with respect to ABI methods and bare app calls. We discovered in PyTeal that supporting ABI methods and bare app calls in clear state programs is actually a dangerous operation, since it introduces failure points to the clear state program (e.g. what if an unsupported method is called?). This was corrected in https://github.com/algorand/pyteal/issues/638, and we want to bring this change to ARC-4 as well.
2. Reverts a previous change which disallowed empty tuple types, e.g. `()`.

### DIFF
--- a/ARCs/arc-0004.md
+++ b/ARCs/arc-0004.md
@@ -266,10 +266,32 @@ from method invocations.
 
 The primary purpose of bare Application calls is to allow the
 execution of an OnCompletion (`apan`) action which requires no
-inputs and has no return value. A Contract **MAY** allow this
-for all OnCompletion actions, for only a subset of them, or for
-none at all. Great care should be taken when allowing these
-operations.
+inputs and has no return value. A Contract **MAY** allow this for
+all of the OnCompletion actions listed below, for only a subset of
+them, or for none at all. Great care should be taken when allowing
+these operations.
+
+Allowed OnCompletion actions:
+* 0: NoOp
+* 1: OptIn
+* 2: CloseOut
+* 4: UpdateApplication
+* 5: DeleteApplication
+
+Note that OnCompletion action 3, ClearState, is **NOT** allowed to
+be invoked as a bare Application call.
+
+> While ClearState is a valid OnCompletion action, its behavior differs
+> significantly from the other actions. Namely, an Application running during
+> ClearState which wishes to have any effect on the state of the chain
+> must never fail, since due to the unique behavior about ClearState
+> failure, doing so would revert any effect made by that Application.
+> Because of this, Applications running during ClearState are
+> incentivized to never fail. Accepting any user input, whether that
+> is an ABI method selector, method arguments, or even relying on the
+> absence of Application arguments to indicate a bare Application call,
+> is therefore a dangerous operation, since there is no way to enforce
+> properties or even presence about data that is supplied by the user.
 
 If a Contract elects to allow bare Application calls for some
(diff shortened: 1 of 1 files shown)


ARCs/arc-0004.md
```diff
@@ -266,10 +266,32 @@ from method invocations.
...
+* 2: CloseOut
+* 4: UpdateApplication
+* 5: DeleteApplication
+
+Note that OnCompletion action 3, ClearState, is **NOT** allowed to
+be invoked as a bare Application call.
+
+> While ClearState is a valid OnCompletion action, its behavior differs
+> significantly from the other actions. Namely, an Application running during
+> ClearState which wishes to have any effect on the state of the chain
```
bbroder-algo, line 286: "an application running during"

ARCs/arc-0004.md
```diff
@@ -266,10 +266,32 @@ from method invocations.
...
+> must never fail, since due to the unique behavior about ClearState
+> failure, doing so would revert any effect made by that Application.
+> Because of this, Applications running during ClearState are
+> incentivized to never fail. Accepting any user input, whether that
+> is an ABI method selector, method arguments, or even relying on the
+> absence of Application arguments to indicate a bare Application call,
+> is therefore a dangerous operation, since there is no way to enforce
+> properties or even presence about data that is supplied by the user.
```
bbroder-algo, line 294: presence = existance?

ARCs/arc-0004.md
jasonpaulos: Is the objection to the capital A? If so it was decided some time ago that Application should be capitalized in this doc

bbroder-algo: more that it is awkwardly phrased and you are describing something very delicate.

bbroder-algo:
in addition to what jason says, one usage could be to modify the ABI method with a zero length array "tag" that creates a secondary method signature for method overloading.

joe-p:
I do see value in a 0-byte type, but confused as to why we use arrays and tuples and not just a `null` type. With arrays and tuples, we basically have an infinite way of representing `0x`. `()`, `(uint64, uint512[], byte[], string)[0]`, and `byte[0]` are all the same which is a bit weird. 

Should mention I don't feel strongly about this for ARC4 since we're going to have to change the ABI anyways with AVM9, but think it's worth having the rationale mentioned on GH.

jasonpaulos:
For completeness, I'll explain my behind supporting 0-length tuples.
 * Utility: it's nice to have a type that is guaranteed to add zero bytes to the encoding. This might be useful if you need to differentiate two methods, and I think it could be useful as a building block for more advanced features in the future.
 * Consistency: 0-length static arrays provide the same feature as 0-length tuples, and for consistency it makes sense to offer this for tuples too.
 * Encoding spec correctness: I think we all agree that 0-length dynamic arrays are valuable. Well the official encoding spec explains how to encode dynamic arrays in terms of static arrays, which are themselves expressed in terms of tuples. So it's very important for everyone to agree that 0-length tuples are a valid construct.

And more practically, PyTeal and the Algorand SDKs are based on the initial version of ARC-4, which had support for 0-length tuples. It would be nice to keep those codebases valid by bringing back this feature.

PR #186, opened 2023-04-04
ARC - 33 xGov clarification

SudoWeezy:
I open this PR for any clarification needed on the process.

Feel free to comment.

### DIFF
--- a/ARCs/arc-0033.md
+++ b/ARCs/arc-0033.md
@@ -40,7 +40,7 @@ The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL 
   </tr>
   <tr>
     <td>Voting Power</td>
-    <td colspan="2">1 Algo commited = 1 Vote</td>
+    <td colspan="2">1 Algo commited = 1 Vote, as per REWARDS DEPOSIT</td>
   </tr>
   <tr>
     <td>Duty</td>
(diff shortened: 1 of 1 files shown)


nonfungibleab:
We must clarify that:

1. Only the GP rewards are deposited to the xGov Term Pool. Let's call it the "REWARDS DEPOSIT." The principal algo committed remains in the gov wallet (or DeFi protocol) and can be used in subsequent Governance Periods.
2. Voting power is 1 Algo = 1 Vote, as per REWARDS DEPOSIT, and it renews at the start of every quarter - provided the xGov remain eligible.

Niftgen:
Yes, this should be clarified.  Examples with numbers will make it more clear. For example how many votes a proposal needs to get to be approved for grant.

SudoWeezy:
> Yes, this should be clarified. Examples with numbers will make it more clear. For example how many votes a proposal needs to get to be approved for grant.

https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0034.md#threshold 
```
eg. 2 000 000 Algo are available to be given away as grants, 200 000 000 millions Algo are committed to the xGov Process (200 000 000 votes available):

Proposal A request 100 000 Algos (5 % of the Amount available)
Proposal A needs 5% of the votes (10 000 000 Votes) to go through
```

PR #190, opened 2023-04-05
fix(arc-6): TS definitions are not valid TS

kewde:
* `EnablNetworkOpts` -> `EnableNetworkOpts`: missing e
* TypeScript definitions were not actually valid, interface declaration shouldn't have =


kewde:
@SudoWeezy

PR #206, opened 2023-06-09
fix: make `stxn` optional

kewde:


kewde:
@SudoWeezy

SudoWeezy:
Thanks, good catch!

PR #210, opened 2023-06-15
ARC-37: ABI versioning

joe-p:
Adds a `version` key to ABI Interface and Contract descriptions

### DIFF
--- a/ARCs/arc-0037.md
+++ b/ARCs/arc-0037.md
@@ -1,14 +1,14 @@
 ---
-arc: <to be assigned>
+arc: 37
 title: ABI Versioning
 description: Add a versioning scheme to the ABI
 author: Joe Polny (@joe-p)
-discussions-to: <URL>
+discussions-to: https://github.com/algorandfoundation/ARCs/issues/210
 status: Draft
 type: Standards Track
 category: ARC
 created: 2023-07-15
-requires: [ARC-0004](./arc-0004.md)
+requires: 4
 ---
 
 ## Abstract
--- a/ARCs/arc-draft_abi_ver.md
+++ b/ARCs/arc-draft_abi_ver.md
@@ -1,0 +1,77 @@
+---
+arc: <to be assigned>
+title: ABI Versioning
+description: Add a versioning scheme to the ABI
+author: Joe Polny (@joe-p)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category (*only required for Standards Track): ARC
+created: 2023-07-15
+requires (*optional): [ARC-0004](./arc-0004.md)
+---
+
+## Abstract
+The Algorand ABI is defined in [ARC-0004](./arc-0004.md). As the protocol and desires of the ABI changes, there will be changes to the ABI. As such, there should be a versioning number to make it clear what versions of the ABI smart contracts adhere to and what versions of the ABI off-chain code supports.
+
+## Motivation
+Without any versioning scheme, it will be very hard to follow the evolution of the Algorand ABI.
+
+## Specification
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in <a href="https://www.ietf.org/rfc/rfc2119.txt">RFC-2119</a>.
+
+Every ABI change is versioned `MAJOR.MINOR`. `MAJOR` is incremeneted when there are breaking changes and `MINOR` is incremented for the addition of backwards-compatible features.
+
+Any ABI Interface or Contract JSON description that adheres to ABI standards beyond [ARC-0004](./arc-0004.md) **MUST** inlcude a `version` key in the top-level of the interface. Interface or Contract JSON descriptions that only adhere to [ARC-0004](./arc-0004.md) **SHOULD** have `"version": "1.0"` at the top level, but descriptions without a `version` key **MUST** be interepreted to implcitly be version 1.0.
+
+## Rationale
+The `MAJOR.MINOR` scheme is used to make it clear what features a given description supports and when clients will become incompatible with  the respective contract.
+
+Since `version` is a new key being added, all existing descriptions without will not have the `version` key, thus the omission of the key is implying version 1.0 ([ARC-0004](./arc-0004.md)).
+
+## Backwards Compatibility
+This ARC is fully backwards compatible with [ARC-0004](./arc-0004.md) and all clients. There will, however, need to be additional effort for clients to parse the `version` key for future evolutions of the ABI.
+
+## Test Cases
+N/A
+
+## Reference Implementation
+
+An [ARC-0004](./arc-0004.md) contract description witht the `version` key
+```json
+{
+  "version": "1.0",
+  "name": "Calculator",
+  "desc": "Contract of a basic calculator supporting additions and multiplications. Implements the Calculator interface.",
+  "networks": {
+    "wGHE2Pwdvd7S12BL5FaOP20EGYesN73ktiC1qzkkit8=": { "appID": 1234 },
+    "SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=": { "appID": 5678 },
+  },
+  "methods": [
+    {
+      "name": "add",
+      "desc": "Calculate the sum of two 64-bit integers",
+      "args": [
+        { "type": "uint64", "name": "a", "desc": "The first term to add" },
+        { "type": "uint64", "name": "b", "desc": "The second term to add" }
+      ],
+      "returns": { "type": "uint128", "desc": "The sum of a and b" }
+    },
+    {
+      "name": "multiply",
+      "desc": "Calculate the product of two 64-bit integers",
+      "args": [
+        { "type": "uint64", "name": "a", "desc": "The first factor to multiply" },
+        { "type": "uint64", "name": "b", "desc": "The second factor to multiply" }
+      ],
+      "returns": { "type": "uint128", "desc": "The product of a and b" }
+    }
+  ]
+}
+```
+
+## Security Considerations
+None
+
+## Copyright
+Copyright and related rights waived via <a href="https://creativecommons.org/publicdomain/zero/1.0/">CCO</a>.
--- a/ARCs/arc-draft_abi_ver.md
+++ b/ARCs/arc-draft_abi_ver.md
@@ -6,9 +6,9 @@ author: Joe Polny (@joe-p)
 discussions-to: <URL>
 status: Draft
 type: Standards Track
-category (*only required for Standards Track): ARC
+category: ARC
 created: 2023-07-15
-requires (*optional): [ARC-0004](./arc-0004.md)
+requires: [ARC-0004](./arc-0004.md)
 ---
 
 ## Abstract
--- a/ARCs/arc-draft_abi_ver.md
+++ b/ARCs/arc-draft_abi_ver.md
@@ -22,25 +22,25 @@ The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL 
 
 Every ABI change is versioned `MAJOR.MINOR`. `MAJOR` is incremeneted when there are breaking changes and `MINOR` is incremented for the addition of backwards-compatible features.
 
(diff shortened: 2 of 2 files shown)


ARCs/arc-0037.md
```diff
@@ -0,0 +1,77 @@
...
+## Abstract
+The Algorand ABI is defined in [ARC-0004](./arc-0004.md). As the protocol and desires of the ABI changes, there will be changes to the ABI. As such, there should be a versioning number to make it clear what versions of the ABI smart contracts adhere to and what versions of the ABI off-chain code supports.
+
+## Motivation
+Without any versioning scheme, it will be very hard to follow the evolution of the Algorand ABI.
+
+## Specification
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in <a href="https://www.ietf.org/rfc/rfc2119.txt">RFC-2119</a>.
+
+Every ABI change is versioned `MAJOR.MINOR`. `MAJOR` is incremeneted when there are breaking changes and `MINOR` is incremented for the addition of backwards-compatible features.
```
jasonpaulos, line 23: Some general questions about this scheme:
* Who decides what constitutes a new version?
* This version scheme is inherently linear. Why do you favor this approach instead of a more piecemeal one, e.g. where a contract lists the additional ARCs that it wants to use beyond ARC-4?

joe-p: > Who decides what constitutes a new version?

Any ARC that changes the ABI would have to include the proposed new version. 

> This version scheme is inherently linear. Why do you favor this approach instead of a more piecemeal one, e.g. where a contract lists the additional ARCs that it wants to use beyond ARC-4?

The main reason is simplicity. Changes to the ABI can be both additive and reductive so it can be hard to parse exactly what features support. For example, saying a contract is ARC-0004 compliant now becomes meaningless because you have to supply every single ABI-related ARC that it uses. Saying ABI 1.0 or ABI 2.0 is much more clear. It also makes it clear for when clients are backwards incompatible.

jasonpaulos: > Any ARC that changes the ABI would have to include the proposed new version.

The order in which ARCs are proposed is not necessarily the order in which they will be finalized. Maybe ARCs could suggest what type of update they would be (i.e. major vs minor), then a maintainer can assign an actual version at the time of finalization?

Regardless of the decision here, I think this process should definitely be mentioned in this ARC.

joe-p: Yeah very good point. I will definitely add the process to this ARC and perhaps it could be a living ARC that contains a list of version changes and their respective ARCs

ARCs/arc-0037.md
```diff
@@ -0,0 +1,77 @@
...
+
+Any ABI Interface or Contract JSON description that adheres to ABI standards beyond [ARC-0004](./arc-0004.md) **MUST** inlcude a `abi-version` key in the top-level of the interface. Interface or Contract JSON descriptions that only adhere to [ARC-0004](./arc-0004.md) **SHOULD** have `"abi-version": "1.0"` at the top level, but descriptions without a `abi-version` key **MUST** be interepreted to implcitly be version 1.0.
```
jasonpaulos, line 25: I wonder if a new key in the interface or contract description is enough. Right now if you want to call another contract's method in PyTeal, you only supply the app ID and the method signature (see [InnerTxnBuilder.MethodCall](https://pyteal.readthedocs.io/en/v0.25.0/api.html#pyteal.InnerTxnBuilder.MethodCall)). The SDKs and goal similarly only rely on method descriptions/signatures and app IDs.

Perhaps for major versions, it makes sense to somehow include this information in the method signature, e.g. `v2:foo()void`? This way, if an app I call decides to update its ABI version, my calls would start failing due to an unsupported method rather than one of us incorrectly decoding values. And if necessary, an app _could_ support the same method across multiple versions.

ARCs/arc-0037.md
jasonpaulos, line 23: I think it would be very useful to provide some examples of what constitutes a major vs minor update. Here's what I came up with so far.

#### Major
* Encoding changes
* Calling convention changes

All of these are certainly backwards incompatible and require the caller and callee to support the changes.

#### Minor:
* New types

Assume version 1.1 introduce a new type, `X`. Version 1.0 clients can still call methods on a version 1.1 contract, as long as those methods don't use `X` as an input or output type. So I _think_ this is still backwards compatible.

ARCs/arc-0037.md
joe-p: One potential problem is that this realistically would have to be managed by the developer per method when working with high-level languages.

For example, let's say a developer writes a contract with the method `foo(uint64)void` in accordance with ARC-0004. Then an ARC like ARC-0043 (#214) is finalized inducing a major version bump and affects other methods in the contract and the developer wants to use the new ABI for their methods. This means the high-level language would use ABI 2.0, but does that mean it should break all existing clients that call `foo(uint64)void` even though the method is completely unaffected by the version change?

I suppose the high-level language could have logic to determine if the change is breaking or not, but seems a bit complex.

pbennett: Since you mentioned it (but part of other ARCS), I think the method for resolution need to be decorated same as in any HLL at build time for RT disambiguation.
To start, methods should probably be namespaced by ARC: otherwise we can't have contracts that implement multiple ARCs.  Based on method names being used in some example ARCs, chance of collision is *extremely* high.
They should also be namespaced by ABI version - otherwise calling convention changes will obv. break things.
so isOwner()void really becomes ARCXXXX/ABIX.X/isOwner()void pre-hash [for example purposes only]

joe-p: What if instead of versioning the full signature only affected types are versioned, with no version implicitly meaning 1.0. 

For example: `foo(string, uint64)void` -> `foo(v2:string,uint64)void`

jasonpaulos: @joe-p interesting! That idea implies that there are no longer any backwards incompatible changes, since existing types are never changed, only antiquated by newer types.

Need to think through the implications of that more, but it seems promising.

joe-p: One other benefit to this is that you can have a method that takes in both v1 and v2 arguments. Perhaps you want to use v2 arguments internally in your methods logic but you need to pass some argument to an inner txn app call that only supports v1. So you could have a method like `foo(string, v2:string)void`. For something like ARC-0037 it's probably not needed since it's a matter of a few opcodes to convert between the two, but for future version that could have more complex changes this would be really useful for composability.

pbennett: I'm confused by the foo(string, v2:string) example.  Why would you have the same method use different encodings for two different arguments of the same type?
I would think having both:
foo(string, string)void
foo(v2:string, v2:string)void 
is what you'd see.

joe-p: > Perhaps you want to use v2 arguments internally in your methods logic but you need to pass some argument to an inner txn app call that only supports v1.

If you can't have both in one method, then you are forced to use whatever the lowest version supported by a potential inner txn app call is.

jannotti: You can accept arguments in any encoding you'd like, and then encode them differently for an inner call.

joe-p: Sure, but that could be costly for the opcode budget.

joe-p:
I want to make some progress on this ARC so we can also make progress on other ARCs this blocks. 

Here's the changes I'm thinking of making

* Don't have a version in the ABI JSON description
* Version new types. No version is v1. For example, `string` -> `string-v2` -> `string-v3`, etc.
* Have a version that follows semver as @jasonpaulos suggested [here](https://github.com/algorandfoundation/ARCs/pull/210#discussion_r1240036349). This version is only used by tooling to advertise what type versions are supported used by default. 


To explain the last point a bit more, an example is TEALScript. If TEALScript advertises support for ABI version 2.0 (and 2.0 includes a new string encoding) and there's a contract with the following method:

```typescript
foo(name: string): string {
    log('Hello, ' + name)
}
```

The ABI signature would be `foo(string-v2)string-v2`. If TEALScript advertised 3.0 support (which includes a new string encoding), the signature would be `foo(string-v3)string-v3`. 

This allows HLLs and tooling to use simple types like `string` will still being able to adopt new encoding types. If a developer really wanted to use a different version, they could specify it

```typescript
foo(name: stringv1): stringv1 {
    log('Hello, ' + name)
}
```

jannotti:
What ARCs are blocked by this?

joe-p:
> What ARCs are blocked by this?

#214 which updates the encoding for an existing type

#211 which would affect how tooling advertises which version of the ABI they support.

jannotti:
> > What ARCs are blocked by this?
> 
> #214 which updates the encoding for an existing type

As #214 itself notes, it is not an important change. It seems like a worse world to have version complexity than to waste two bytes in string lengths.

> #211 which would affect how tooling advertises which version of the ABI they support. 

I don't think there's any reason to _remove_ the reference types. Surely wallets and sdks would not _remove_ support for calling existing contracts that use them. It can be a reasonable thing to recommend against their use, as they preclude the use of shared resources. But they don't hurt anything to exist.

So I can't see why we would introduce the complexity of versioning to the ABI to support these.

joe-p:
> As https://github.com/algorandfoundation/ARCs/pull/214 itself notes, it is not an important change. It seems like a worse world to have version complexity than to waste two bytes in string lengths.

It's not just about saving bytes, but also saving opcode cost and compatability with underlying AVM opcodes and contracts that don't implement the ABI. 

The ARC meeting discussing this ARC had unanimous agreement that this was a sensible change. The versioning complexity you're referring to is a one-liner conditional statement in encoders/decoders. Most applications right now aren't even using the ABI so doing this change now rather than later makes sense imo. 

> I don't think there's any reason to remove the reference types. Surely wallets and sdks would not remove support for calling existing contracts that use them. It can be a reasonable thing to recommend against their use, as they preclude the use of shared resources. But they don't hurt anything to exist.

If they aren't recommended to be used then some tooling might not want to implement them. It would be nice to give them a proper way to advertise what they do or don't support.

jannotti:
> > As #214 itself notes, it is not an important change. It seems like a worse world to have version complexity than to waste two bytes in string lengths.
> 
> It's not just about saving bytes, but also saving opcode cost and compatability with underlying AVM opcodes and contracts that don't implement the ABI.

We concluded in that discussion that this is only about the 15 top-level arguments. If every single one of them meets the requirements of this change, you have saved exactly 15 opcodes, as each could be converted to this datatype by calling `extract 2 0` on them.  More likely, you have maybe two string arguments, so you have saved exactly 2 opcodes.

You talk about compatibility with contracts that don't implement the ABI. First, that's addressed by `extract 2 0`. Second, we should care about compatibility with contracts that _do_ implement the ABI. Third, by introducing this change, you now need both encoding paths in all four SDKs to handle the different version types.

It is _easy_ to turn an existing ABI string into what you want (`extract 2 0`). It is hard to do the reverse. So it makes more sense to hand in the uniform, easily converted form (which has the length prefix). That way, no matter what kind of contract you call, it is easy to produce the form it needs (whether it rakes it as a top-level argument or not).

Using this as a justification for adding the complexity of versioning to the ABI makes no sense.

> > I don't think there's any reason to remove the reference types. Surely wallets and sdks would not remove support for calling existing contracts that use them. It can be a reasonable thing to recommend against their use, as they preclude the use of shared resources. But they don't hurt anything to exist.
> 
> If they aren't recommended to be used then some tooling might not want to implement them. It would be nice to give them a proper way to advertise what they do or don't support.

It would be horrible mistake for any tool to "not implement" the reference types, as contracts that use them exist. So explorers need to show them, and languages need to be able to invoke them.  At most, tools might recommend against creating new contracts that take these types as arguments. But these tools still need to know about them and have code paths that understand them and potentially warn about them.  That is not an ABI version change, that is recommending against some types. But never, ever, removing support for them.

You can't argue against the complexity this introduces by saying "It's just one if statement". It's the entire idea of versioning an ABI that's complex.  If we leave things as is, tools don't ever discuss "ABI versions".  The idea and complexity of that discussion ceases to exist.

joe-p:
> We concluded in that discussion that this is only about the 15 top-level arguments. If every single one of them meets the requirements of this change, you have saved exactly 15 opcodes, as each could be converted to this datatype by calling extract 2 0 on them. More likely, you have maybe two string arguments, so you have saved exactly 2 opcodes.

Yeah I was initially thinking of the opcode cost for going from byteslice -> string but that is only be truly problematic  when encoding tuples, which would need the prefix anyways unless we had #212, and as mentioned there the usefulness isn't really evident because most people don't use complex data structures. If that becomes an evident problem then I think these sort of changes are worth looking at again, but I will be closing #211, #212, and #214 in the meantime.

PR #211, opened 2023-06-15
ARC-38: Remove ABI Reference Types

joe-p:
This ARC proposes the removal of ABI reference types from the ABI standard.


cusma:
Hi @joe-p, why you opted for removing the "alias" you originally suggested for the reference types? What are the cons?

joe-p:
> Hi @joe-p, why you opted for removing the "alias" you originally suggested for the reference types? What are the cons?

By removing the types entirely there's no worry for a client using the wrong encoding. These sort of type aliases that are solely for the purposes of forming transactions will be handled in ARC32.

PR #213, opened 2023-06-16
Arc 41 - NFT's vault of NFT, Algo & Fungible Token

SudoWeezy:
This standard is inspired by <a href="https://eips.ethereum.org/EIPS/eip-6551">EIP-6551</a>
The goal is to allow an ASA to detain & manage ASAs or Algo through a ”vault”.

### DIFF
--- a/ARCs/arc-0041.md
+++ b/ARCs/arc-0041.md
@@ -53,7 +53,6 @@ In order to allow wallet & dApps to identify the vault of the Authority NFT, an 
           "desc": "The Application ID of the authority NFT, (0 if ASA)"
         }
       ]
-    }
 }
 ```
 
--- a/ARCs/arc-0041.md
+++ b/ARCs/arc-0041.md
@@ -3,7 +3,7 @@ arc: 41
 title: ASA's vault of ASAs
 description: Application ID storing ASAs for another ASA
 author: Stéphane BARROSO (@SudoWeezy)
-discussions-to: https://github.com/algorandfoundation/ARCs/pull/213
+discussions-to: https://github.com/algorandfoundation/ARCs/issues/213
 status: Draft
 type: Standards Track
 category: Interface
@@ -98,7 +98,7 @@ Linking the vault ID to the Main ASA using the Metadata allows to have backwards
 The design using an application has vault rather than a logic sig is more flexible and secure.
 
 ## Backwards Compatibility
-This stantard is compatible with asa that can be updated [ARC-3](./arc-0003.md) + [ARC-19](./arc-0019.md) or [ARC-69](./arc-0069.md).
+This stantard is compatible with ASAs that can be updated [ARC-3](./arc-0003.md) + [ARC-19](./arc-0019.md) or [ARC-69](./arc-0069.md).
 
 ## Security Considerations
 The method `isOwner` **MUST** be called by every other method of the application to ensure that the caller detain the main ASA.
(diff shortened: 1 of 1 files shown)


ARCs/arc-0041.md
```diff
@@ -0,0 +1,107 @@
...
+```
+### Interface
+```json
+{
+  "name": "ARC-41",
+  "desc": "Main ASA's Vault Base Interface",
+  "methods": [
+    {
+      "name": "isOwner",
+      "desc": "Indicate if the caller detain the Main ASA of this vault, this method MUST be called before calling every other method of the application",
```
joe-p, line 59: The wording make it seem like the caller must always send two app call txns. Instead, the implementer can simply add call the `isOwner` logic internally when calling any of the other methods. I think this should be made more clear.

ARCs/arc-0041.md
```diff
@@ -0,0 +1,107 @@
...
+## Abstract
+This standard is inspired by <a href="https://eips.ethereum.org/EIPS/eip-6551">EIP-6551</a>
+The goal is to allow an ASA to detain & manage ASAs or Algo through a ”vault”.
+
+## Specification
+The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in <a href="https://www.ietf.org/rfc/rfc2119.txt">RFC-2119</a>.
+
+### Glossary
+In the document we will use following syntax:
+- Main ASA -> ASA that detain & manage others ASAs
```
joe-p, line 23: The wording here is weird. "Main ASA" seems very vague and it took me awhile to really understand what it meant. I would suggest something like "Authority ASA". 

Also, the ASA itself can't hold or manage ASAs, but rather indicate a specific account can do those things. I think that should be made more clear.

SudoWeezy: Good point, authority make more sense

ARCs/arc-0041.md
```diff
@@ -0,0 +1,155 @@
...
+          "type": "uint64",
+          "name": "AppId",
+          "desc": "The Application ID of the authority NFT, (0 if ASA)"
+        }
+      ]
+    }
+}
+```
+
+A `arc41_AccountCreated` event SHOULD be emitted, when a vault is created.
```
pbennett, line 60: This states 'should' be emitted, but line 27 says 'must' - which is correct?

ARCs/arc-0041.md
```diff
@@ -0,0 +1,155 @@
...
+  ]
+}
+```
+
+The `arc41_isOwner` method MUST be called internaly by every other methods.
+The `arc41_transferAlgoFrom` method MUST error when `from` is not the owner of `asaId` or `tokenId`.
+The `arc41_transferFrom` method MUST error when `from` is not the owner of `asaId` or `tokenId`.
+
+## Rationale
+ASAs NFT ([ARC-3](./arc-0003.md), [ARC-69](./arc-0069.md)) and Application NFT ([ARC-72](./arc-0072.md)) are compatible with This standard.
```
pbennett, line 146: Although ARC-3, ARC-19, ARC-69 are all just ASAs and nothing about this spec cares about their initial config or metatadata variations (and just tracks an ASA ID), I'd add ARC-19 to this list for completeness - otherwise some might think it doesn't apply.

pbennett (changes requested):
not commenting on substance of arc atm, but have some minor feedback

ARCs/arc-0041.md
SudoWeezy: Thanks, it's a must

SudoWeezy:
Closing this as it’s been 1.5 years old, and new things have come up since then.

Issue #42, opened 2021-11-09
Discussion for ARC-0014 - Algorand Standard for authentication


scholtz:
Please comment on https://github.com/algorandfoundation/ARCs/pull/41/files  - Algorand Standard for authentication

#41

At the Swiss Blockchain Hackathon as a side product we have developed the authentication mechanism for authentication of algorand accounts.

Person just have to sign the zero transaction, and use it the similar way as JWT tokens are being used. The backend has this way the information that the account which calls the methods is capable of signing the messages. It does not cost anything because the transaction must not go to the network, but can be validated by the backend service as valid signed transaction. Expiration of this token is handled by the suggested params in the transactions - the latest block parameter.

.net implemetation: https://github.com/scholtz/AlgorandAuthenticationDotNet

scholtz:
GraphQL authentication webhook for ARC0014 https://github.com/scholtz/HasuraAlgorandAuthWebHook

https://hub.docker.com/r/scholtz2/hasura-algorand-auth-web-hook

scholtz:
There has been another implementation using Algorand .net auth

https://github.com/scholtz/AlgorandKMDServer

Open api specs: http://kmd.h3.a-wallet.net/swagger/

Implemented in AWallet .. https://github.com/scholtz/wallet

Video on how to make account online with arc0014 auth: https://www.youtube.com/watch?v=cSJjmnzvvLY

scholtz:
For the record the activity diagram of the original ARC-0014

@startuml

User -> Website: User wish to login to the website 


Website -> AnyAlgorandNode : Request /v2/transactions/params
AnyAlgorandNode  -> Website : Returns transaction params

Website -> Wallet : Website produces through algosdk messege to be signed, note as realm, params as session timeout
Wallet -> Website: Signed tx, website stores tx in temporary memory, localstorage or cookie the same as jwt

Website --> User: Displays eg auth address

Website -> Backend: Request authorized method, use SigTx Authorization header, fe /getUser

Backend -> AnyAlgorandNode : Request /v2/transactions/params
AnyAlgorandNode  -> Backend : Returns current block round number

Backend -> Website: 401 if signature, genesis or round is wrong
Backend -> Website: 200 if authorized and returns content

Website --> User: Displays /getUser content, fe roles or username

@enduml

![image](https://user-images.githubusercontent.com/1223439/195995662-5d93f547-ff4a-4277-8546-74a627194dc8.png)

[...]

PR #218, opened 2023-06-25
ARC-0014 Algorand authentication using signed transaction

scholtz:
arc-0014 back to life

Following https://github.com/algorandfoundation/ARCs/pull/41

and https://github.com/algorandfoundation/ARCs/issues/42

ARC14 is widely used for example in these projects:

- https://www.nuget.org/packages/AlgorandAuthentication
- https://github.com/scholtz/AlgorandAuthenticationDotNet
- https://github.com/scholtz/AlgorandKMDServer
- https://kmd.h3.a-wallet.net/swagger/index.html
- https://github.com/scholtz/HasuraAlgorandAuthWebHook
- https://github.com/scholtz/Algorand2FAMultisig
- https://2fa-docs.a-wallet.net/
- https://2famsig.k8s.aramid.finance/swagger/index.html
- https://github.com/scholtz/wallet
- https://github.com/scholtz/AlgorandPublicData/

Newly it has full support for rekeying and multisig.

### DIFF
--- a/ARCs/arc-0014.md
+++ b/ARCs/arc-0014.md
@@ -1,6 +1,6 @@
 ---
 arc: 14
-title: Algorand authentication using signed transaction
+title: Algorand authentication
 description: Algorand authentication using signed transaction.
 author: Ludovit Scholtz (@scholtz)
 discussions-to: https://github.com/algorandfoundation/ARCs/pull/218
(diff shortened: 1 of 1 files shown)


emg110:
Hey ludo, great initiative! Thank you!
quick comments:

- This looks more like a method description rather than a standard (regardless of format).
- This method is been in use since a long time ago and is actually irrelevant to wallets and wallets should not do anything to support this (since it uses standard TXN object) and you can gatherSignatures() using ATC!
- The message construction is left unexplained!
- There are no specific recommendations or procedures on how to make TXNs unique 
- This is weakly resembling JOSE Standards (rfc7519 and rfc7515) and therefore I humbly suggest it to be 100% compliant to those living standards (JOSE and rfc7515 support ed25519)!

So in general  I humbly suggest, given your great technical abilities, that by complying to JWT standard(rfc7519), you make ARC14 the reference implementation of the rfc7519 standard on Algorand!

In order to do that I also humbly suggest:
- Use JWT construction format (header, payload and signature parts) and standard rfc7519 structures
- For payload claims , use `iss, exp, iat, nbf, jti, sub, aud` fields
- Use two types of standard TXNs for payload message construction: A 0 amount payment TXN From/To account address and An ABI call transaction from the account address to a smart contract complying to an ABI extension you propose in ARC14. These two types will cover all types of JWT scenarios from simplest to most complicated!
- Propose a nonce (using JWT recommendations) composition and use it in note field of transaction with a standardized format (e.g. nonce: base64 encoded of random 64 bytes) 
- Specify a TTL expiration (configurable with a default of 0 as never expires)
- Specify a revoke mechanism and interface to revoke credentials access or authorization or validity of authentication!
- Avoid using non-standard fields, routines, headers and properties like sigTxn or such because they will cause isolation for ARC standard and generally Algorand, IMHO!
- Make sure to involve the timestamp in the process of nonce or challenge creation!

Best of luck and thank you again for all the great contributions you delicately make! Let's keep rocking!

scholtz:
Sorry, but the purpose of this standard is not to use JWT.

JWT tokens
- requires custom signatures (either with single secret or using public/private keys) 
- it is not possible to sign the jwt tokens with current algorand wallets

If you want to use jwt tokens at data transfer, feel free to create different standard.

Note that this standard already proposes the TTL, nonce and timestamp.. All algorand messages are signed with valid from and valid until fields, and this is used to validate session validity.

This standard is already working, many projects has implemented it, it is live for at least 2 years, and there is no reason why algorand foundation should be postponing it further.

Please move it final state.

scholtz:
@emg110 , i dont get it.. what makes you to comment on this standard where it is explicitly said that this is for authentication using signed transactions.. feel free to create another standard and AF please put this one to the final stage

I understand that JWT is good, but i want to have the native wallet support, even from the wallets like metamask where you can just do dummy signing of transaction. I have already dozen projects on ARC14, and I may consider start the transition to some JWT algorand auth when i see it is working for my needs.

I dont get this conversasion where you do not use any argument what is wrong with the written text but you propose to put the whole thing to trash and go back 2 or 3 years.

scholtz:
Also notice that in this proposal is not mentioned ed25529 .. it is because algorand can come up with higher layer of security and make ed25529 accounts obsolete.. we want to be the quantum age safe.. right? 

i imagine it the way for example that algorand will force two different types of signatures on the accounts.. ed22529 and some other and valid transaction can be submitted to the blockchain only if both signatures are valid

in that times it will be still possible to sign transactions for arc14 and only valid signed txs will be ok

with jwt you stick to single signing mechanism, and btw you did not mention support for multisig or rekeying which in arc14 is implemented

SudoWeezy:
There are no consensus around this proposal.

scholtz:
But there was consensus that using the sign bytes it will take years to implement. So please keep this one open

PR #219, opened 2023-06-26
Adding arc number prefix to Application's method name

SudoWeezy:
There is a need to provide information about which ARCs have been implemented within a particular application. To address this requirement, it is proposed to introduce a standardized namespace for each implemented ARC number to be used before every method name.

### DIFF
--- a/ARCs/arc-0000.md
+++ b/ARCs/arc-0000.md
@@ -206,6 +206,7 @@ eg:
       "args": [
         { "type": "uint64", "name": "Number", "desc": "A number" },
       ]
+    },
     {
       "name": "arc0_Event2",
       "desc": "Method 2",
--- a/ARCs/arc-0000.md
+++ b/ARCs/arc-0000.md
@@ -165,9 +165,9 @@ To provide information about which ARCs has been implemented on a particular app
 > Where <ARC number> represents the specific ARC number associated with the implemented requirement change.
 
 eg:
-'''json
+```json
 {
-  "name": "ARC0",
+  "name": "Naming convention",
   "desc": "Example",
   "methods": [
     {
@@ -189,8 +189,6 @@ eg:
   ]
 }
 ```
-
-'''
 
 ## Rationale
 
--- a/ARCs/arc-0000.md
+++ b/ARCs/arc-0000.md
@@ -160,6 +160,37 @@ References to other ARCs should follow the format `ARC-N`, where `N` is the ARC 
 
 Images, diagrams, and auxiliary files should be included in a subdirectory of the `assets` folder for that ARC as follows: `assets/arc-N` (where **N** is to be replaced with the ARC number). When linking to an image in the ARC, use relative links such as `../assets/arc-1/image.png`.
 
+### Application's Methods name
+To provide information about which ARCs has been implemented on a particular application, namespace with the ARC number should be used before every method name: `arc<ARC number>_MethodName`.
+
+eg:
+'''json
+{
+  "name": "ARC0",
+  "desc": "Example",
+  "methods": [
+    {
+      "name": "arc0_method1",
+      "desc": "Method 1",
+      "args": [
+        { "type": "uint64", "name": "Number", "desc": "A number" },
+      ],
+      "returns": { "type": "void[]" }
+    },
+    {
+      "name": "arc0_method2",
+      "desc": "Method 2",
+      "args": [
+        { "type": "byte[]", "name": "user_data", "desc": "Some characters" }
+      ],
+      "returns": { "type": "void[]" }
+    }
+  ]
+}
+```
+
+'''
+
 ## Rationale
 
 This document was derived heavily from <a href="https://github.com/ethereum/eips">Ethereum's EIP-1</a>, which was written by Martin Becze and Hudson Jameson, which in turn was derived from  <a href="https://github.com/bitcoin/bips">Bitcoin's BIP-0001</a> written by Amir Taaki, which in turn was derived from <a href="https://www.python.org/dev/peps/">Python's PEP-0001</a>. In many places, text was copied and modified. Although the PEP-0001 text was written by Barry Warsaw, Jeremy Hylton, and David Goodger, they are not responsible for its use in the Algorand Request for Comments. They should not be bothered with technical questions specific to Algorand or the ARC. Please direct all comments to the ARC editors.
--- a/ARCs/arc-0000.md
+++ b/ARCs/arc-0000.md
@@ -161,13 +161,13 @@ References to other ARCs should follow the format `ARC-N`, where `N` is the ARC 
 Images, diagrams, and auxiliary files should be included in a subdirectory of the `assets` folder for that ARC as follows: `assets/arc-N` (where **N** is to be replaced with the ARC number). When linking to an image in the ARC, use relative links such as `../assets/arc-1/image.png`.
 
 ### Application's Methods name
-To provide information about which ARCs has been implemented on a particular application, namespace with the ARC number should be used before every method name: `arc<ARC number>_MethodName`.
-> Where <ARC number> represents the specific ARC number associated with the implemented requirement change.
+To provide information about which ARCs has been implemented on a particular application, namespace with the ARC number should be used before every method name: `arc<ARC number>_methodName`.
+> Where <ARC number> represents the specific ARC number associated to the standard.
 
 eg:
 ```json
 {
-  "name": "Naming convention",
+  "name": "Method naming convention",
   "desc": "Example",
   "methods": [
     {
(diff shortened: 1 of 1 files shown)


ARCs/arc-0000.md
```diff
@@ -160,6 +160,38 @@ References to other ARCs should follow the format `ARC-N`, where `N` is the ARC
 
 Images, diagrams, and auxiliary files should be included in a subdirectory of the `assets` folder for that ARC as follows: `assets/arc-N` (where **N** is to be replaced with the ARC number). When linking to an image in the ARC, use relative links such as `../assets/arc-1/image.png`.
 
+### Application's Methods name
+To provide information about which ARCs has been implemented on a particular application, namespace with the ARC number should be used before every method name: `arc<ARC number>_MethodName`.
+> Where <ARC number> represents the specific ARC number associated with the implemented requirement change.
+
+eg:
+'''json
```
joe-p, line 168: This codeblock is messed up. I think your ''' should be ```

temptemp3:
I would have to disagree with everyone on this page. I think that it is overkill and not useful. Providing information about the implementation in documentation such as which methods are implemented is sufficient. Also, I would prefer the ARC73 method (https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0073.md) over the proposed convention in this PR.
At the very least the choice of adopting this method should be optional. Furthermore, ARC73 should be optional in implementation, since interfaces should be mutually exclusive.
In the case that 2 specs share a similar method name which I have a hunch would be uncommon, 1 of 2 specs would be chosen to implement the method. In the case  that both method are required, usage of the namespace method would be useful.

pbennett:
@temptemp3 A single contract can trivially *implement* multiple interfaces.  

ARC-73 is independent - its role is simply to answer the qeustion: "DOES" contract X support interface Y?  That's it.  
It does NOT address a contract implementing multiple interfaces that happen to have overlap in method declarations.

Different interfaces could have the same method name and arguments (particularly those with no arguments, or even just different use of them - one might have uint64 be an asset id, another might reference an app id).  isOwner(uint64)
what's uint64 here?  Depends on the ARC - but that doesn't mean a single contract which fits many roles couldn't implement multiple contracts.

temptemp3:
> @temptemp3 A single contract can trivially _implement_ multiple interfaces.
> 
> ARC-73 is independent - its role is simply to answer the qeustion: "DOES" contract X support interface Y? That's it. It does NOT address a contract implementing multiple interfaces that happen to have overlap in method declarations.
> 
> Different interfaces could have the same method name and arguments (particularly those with no arguments, or even just different use of them - one might have uint64 be an asset id, another might reference an app id). isOwner(uint64) what's uint64 here? Depends on the ARC - but that doesn't mean a single contract which fits many roles couldn't implement multiple contracts.

@pbennett 

That sounds reasonable. I understand your concern about having to add the "arcXX_" prefix to every method and event when implementing interfaces. It can become cumbersome, especially when dealing with a single interface or non-overlapping interfaces.

Your suggestion of using application ARC tags seems like a good approach. By including ARC tags in the method names, you can differentiate between different ARC implementations within the same application. For example, if the application ARC tags are [ARC200][ARC0], the method names would be prefixed accordingly: "arc200_transfer(...)" and "arc200_transferFrom(...)". Conversely, if only [ARC200] is specified, the prefixes would be omitted: "transfer(...)" and "transferFrom(...)". This approach allows for clearer identification and organization of methods based on the ARC implementation.

Additionally, you raise a valid point about conflicts that may arise when combining ARC tags. In cases where there are conflicting method names, even if the interface IDs differ, it would be helpful to add the [ARC0] tag to indicate that the method is a generic implementation. For example, if you have [ARC72][ARC200], both of which implement a method called "transferFrom," it would be convenient to add the [ARC0] tag. In this scenario, you could use the following ARC tags: [ARC0][ARC72][ARC200]. The resulting method names would be: "arc200_transfer(...)", "arc200_transferFrom(...)", "arc72_transferFrom(...)", and so on.

In summary, your proposed solution of using application ARC tags and selectively adding [ARC0] when conflicts arise seems reasonable. It provides a clear and flexible naming convention that accommodates different ARC implementations while maintaining independence. I agree with your point that ARC0 should not be considered part of another ARC since ARCs are independent entities, varying in their level of independence.

pbennett:
I'm confused.  I don't recall suggesting tags and I don't have a concern about the arcXX prefix.
Or are you just saying that for the purposes of ARC-73 - it would implement the specified interfaces, although ARC-0 isn't an interface per-se.  A given ARC if honoring ARC0 would simply have its ARC prefixing the method names which would automatically alter the ARC-73 hash for that particular ARC.

temptemp3:
> I'm confused. I don't recall suggesting tags and I don't have a concern about the arcXX prefix. Or are you just saying that for the purposes of ARC-73 - it would implement the specified interfaces, although ARC-0 isn't an interface per-se. A given ARC if honoring ARC0 would simply have its ARC prefixing the method names which would automatically alter the ARC-73 hash for that particular ARC.

Oops. Sorry for the drop off and inventing tags. Also thanks for summarizing.

>  A given ARC if honoring ARC0 would simply have its ARC prefixing the method names 

Yes

> which would automatically alter the ARC-73 hash for that particular ARC.

Yes, the developer would be required to update those hashes for correctness

---

For backwards compatibility, I would be reasonable to honor ARC0. That is, not apply ARC prefixing if not necessary.

pbennett:
> For backwards compatibility, I would be reasonable to honor ARC0. That is, not apply ARC prefixing if not necessary.

Right, for ARCs and contracts *already deployed* I would expect them to be immutable at this point.  The majority of the proposed contract ARCs are still not final though so altering their method names for ARC 0 is reasonable IMO.

temptemp3:
All in all I think it is reasonable if you read between the lines.

> To provide information about which ARCs has been implemented on a particular application, namespace with the ARC number should be used before every method name: `arc<ARC number>_methodName`.

... namespace with the ARC number SHOULD be used before every method name

SudoWeezy:
Please upvote or downvote my comment: 
What if we add just one line saying something like
`If the application respects multiple ARC, then it MUST use arcxx_ prefix else, it is not requested.`

joe-p:
> ... namespace with the ARC number SHOULD be used before every method name

I would propose that this should be changed to a MUST

> If the application respects multiple ARC, then it MUST use arcxx_ prefix else, it is not requested.

So now clients must always check which for both possible methods before interacting with the app? This is unreasonable for inner transactions and hurts composability. 

Also, if two ARCs implement something simple like`isOwner()boolean` and a contract implements `isOwner()boolean` how do you know what ARC it corresponds to? ARC73 alone can't tell you this

pbennett:
> If the application respects multiple ARC, then it MUST use arcxx_ prefix else, it is not requested.

@SudoWeezy this doesn't make sense.  If an implementor implements multiple multiple ARCs then they need to change the method names of another ARC within their implementation ?  Now it won't match the actual ARC they're implementing, nor its ARC73 hash !

I agree with @joe-p, I don't see how this really shouldn't be a MUST.  It makes sense that *existing* (final!) ARCs and *existing* contracts won't be compliant but any new or in-process ARC should have this namespacing prefix so contracts can successfully be implementations of multiple ARCS.

temptemp3:
I agree with @SudoWeezy and think it makes sense to add a namespace when introducing ARCS that introduce conflicts such as overlapping method names. 
I don't think that we should blindly add ARCxx_ to every method and event name if not required. In particular, an application that is an implementation of a single should SHOULD not have to mark all its methods and events with ARCxx_. However, it would be crazy not to add ARCxx_ if an application respects multiple ARCS.
Regarding the ARC73 hash, there would be an ARC73 hash that does not include ARC namespacing and one that does. However, an application would have one or the other, not both.

pbennett:
It feels like there's a huge misunderstanding here.  What you're arguing doesn't make much sense to be honest.
If an application implements multiple ARCs those are NOT ITS ARCS !  It's implementing an interface defined elsewhere !  Those ARCs must be namespaced otherwise you can't implement multiple ARCS with overlapping methods !
It's not a 'decide later' sort of thing.  
You won't have conflicts until someone else uses the same names (which they will) - if they don't even look at other ARCs, or others are added later, they'll just assume, all good... no prefix and you've completely defeated the purpose.

temptemp3:
In my opinion, ARC namespacing should be applied if implementing multiple ARCS. It doesn't make sense to do when implementing a single ARC.

pbennett:
> In my opinion, ARC namespacing should be applied if implementing multiple ARCS. It doesn't make sense to do when implementing a single ARC.

It doesn't work that way.   A caller expecting to call contract A which implements ARCX is going to call the methods for that interface as defined by its spec (and which arc-73 would also match on).  Contract A calling contract B and expecting contract B (which could be dynamic - itself an argument to A!) to support a particular ARC is going to call that ARC's method names.  It's not method foo or... maybe... method ARCXX_foo.   What you're suggesting is completely unworkable.

temptemp3:
What I am _suggesting_ is to not apply ARC namespacing to anything and everything under the sun and that applying the ARC namespace method would come in handy when working with applications support multiple ARCS.

temptemp3:
I concede. I already have something on my plate that may benefit from the organization of ARC namespacing.

PR #220, opened 2023-06-27
ARC45: gtxn ABI Encoding

joe-p:
Introduces a new, more flexible, encoding for referencing other transactions in the group

### DIFF
--- a/ARCs/arc-0045.md
+++ b/ARCs/arc-0045.md
@@ -1,9 +1,9 @@
 ---
-arc: <to be assigned>
+arc: 45
 title: gtxn ABI Encoding
 description: Introduces a new, more flexible, encoding for referencing other transactions in the group
 author: Joe Polny (@joe-p)
-discussions-to: <URL>
+discussions-to: https://github.com/algorandfoundation/ARCs/issues/220
 status: Draft
 type: Standards Track
 category: ARC
--- a/ARCs/arc-draft_txn_types.md
+++ b/ARCs/arc-draft_txn_types.md
@@ -1,0 +1,90 @@
+---
+arc: <to be assigned>
+title: gtxn ABI Encoding
+description: Introduces a new, more flexible, encoding for referencing other transactions in the group
+author: Joe Polny (@joe-p)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: ARC
+created: 2023-07-26
+requires: 4
+---
+
+## Abstract
+This ARC proposes a flexible encoding scheme for referencing transactions in a group via either relative or direct indexing.
+
+## Motivation
+The [ARC-4](./arc-0004.md) encoding of transaction types is quite limiting due to the fact that it requires transactions to come before the application call in succession. This also means that if multiple app calls in a group require a specific transaction, those app calls must all be accompanied by a similar transaction, thus taking up extra space in the group. There needs to be a more flexible way to use transactions with ABI methods.
+
+## Specification
+
+### ARC-4 Transaction Type Removal
+
+The current [ARC-4](./arc-0004.md) transaction types (`txn`, `pay`, `keyreg`, `acfg`, `axfer`, `afrz`, `appl`) are no longer valid ABI types.
+
+### gtxn Encoding
+The `gtxn` type is a reference to a transaction in the same group as the current application call. `gtxn` **MUST** be encoded in 2 bytes. The least significant bytes is a number 0-16 encoded as `uint8`. The first byte is also a `uint8` that indicates the mode of the reference.
+
+| Mode | Description |
+| ---- | ----------- |
+| 0 | The least significant byte is the offet that should be *subtracted* from `txn GroupIndex` of the current application call |
+| 1 | The least significant byte is the offet that should be *added* from `txn GroupIndex` of the current application call |
+| 2 | The least significant byte is the exact index of the transaction being referenced in the current application call's group |
+
+
+## Rationale
+There were condierations for a more complex encoding that would leave room for referencing inner transactions. Due to the uncertainty of how this would look like in the AVM, however, it didn't seem possible to come up with a sensible encoding scheme for referencing inner transactions (particularly for relative indexing).
+
+Due to the above, this encoding scheme allows any transaction in the current group of the application call can be referenced, but inner transactions are still unable to be referenced.
+
+## Backwards Compatibility
+Due to the removal of the current transaction types, this ARC is not backwards compatiable with [ARC-4](./arc-0004.md).
+
+## Reference Implementation
+
+### `gtxn` Encoding Examples
+
+* `0x0003` signifies the referenced transactions comes three transactions *before* the current application call.
+* `0x0103` signifies the referenced transactions comes three transactions *after* the current application call.
+* `0x0203` signifies the referenced transactions is the 4th transaction in the current transaction group.
+
+### TEAL Decoding Example
+
+```
+deocde_gtxn:
+    txnas ApplicationArg     // [encoded_gtxn]
+    dup                      // [encoded_gtxn, encoded_gtxn]
+    extract 1 2              // [encoded_gtxn, byte(index)]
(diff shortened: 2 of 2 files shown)


jannotti (commented):
The aspect that you call "quite limiting" is *on purpose*. The goal of the ABI not to describe any possible application call requirement. It's to provide a reasonable way to encode an application call, particularly one that was designed with the ABI in mind in the first place. Thus, such app calls would expect their associated transactions to come before them in the group.

If we provide a way to reference an arbitrary transaction in the group, then inevitably we will have problems where two apps are called, and they are both given the _same_ txn as input. So they might both think, for example, that the pay they are looking at is "for them". But it's only one transaction! They shouldn't both be funded by the same pay.

Better to limit the form of app calls to limit such confusion, so that any app call will look *only* at the previous, adjacent transactions.

joe-p:
> The aspect that you call "quite limiting" is on purpose. The goal of the ABI not to describe any possible application call requirement. It's to provide a reasonable way to encode an application call, particularly one that was designed with the ABI in mind in the first place. Thus, such app calls would expect their associated transactions to come before them in the group.

The problem is that people are already writing non-compliant ABI apps for this use case. I understand the rationale, but the reality is that an standard that doesn't cover all of the the potential use cases is not useful because some apps will use the standard, some won't, and some will mix (for example, apps that require transactions to come before AND after the app call). This, in my opinion, is worse than not having a standard at all. 

We have a standard that people aren't using because it doesn't fit their applications. This is an attempt to standardize things that developers currently doing. 

> If we provide a way to reference an arbitrary transaction in the group, then inevitably we will have problems where two apps are called, and they are both given the same txn as input. So they might both think, for example, that the pay they are looking at is "for them". But it's only one transaction! They shouldn't both be funded by the same pay.

Sure, this adds additional logic that an app has implement but it's not insurmountable. Tracking number of calls in app state seems like a very simple way to handle this.

jannotti:
Imagine an app that needs to be paid by a simple pay to that app's app account. With the scheme you have proposed, I can call that app twice and point it at the same pay transaction. You can't expect apps to figure that out, if you give them a simple mechanism to just get a transaction and use it.

You are simultaneously claiming that non-ABI exists that must be described, AND that such apps will be rewritten to carefully discriminate transactions to avoid these "double pays"

If an app chooses to ignore the ABI, so be it. It's not the ABI's job to describe arbitrary requirements.

joe-p:
To give some more context... there's a reason this calls for the removal of the current transaction types from the ABI. By themselves, they provide very little information to the caller, which seems to be the entire purpose in the first place. 

If the `gtxn` encoding is not something people want, then I'd still argue that transaction types should be removed from the ABI all together. There needs to be more information to make these sort of arguments useful which becomes a problem much larger than the ABI (and something that should be tackled in ARC-32).

There can be higher-level types provided by ARC-32 that specify leading and proceeding transactions and the app logic must hardcode the relative index, but this encoding gives an option for more flexibility. I imagine *most* applications will use hard-coded relative indexes because as you mention it makes life easier. 

> Imagine an app that needs to be paid by a simple pay to that app's app account. With the scheme you have proposed, I can call that app twice and point it at the same pay transaction. You can't expect apps to figure that out, if you give them a simple mechanism to just get a transaction and use it.

If you want to have one-payment per call then use a bitmask in global state to mark which txn indexes have already been used. Or maybe you have an array of additional app calls and a counter for each app call to ensure you don't exceed the limit of the provided calls. I'm sure there are plenty of other reasonable ways to keep track as well. 

> You are simultaneously claiming that non-ABI exists that must be described, AND that such apps will be rewritten to carefully discriminate transactions to avoid these "double pays"
>
> If an app chooses to ignore the ABI, so be it. It's not the ABI's job to describe arbitrary requirements.

Perhaps we have a different view on what the purpose of the ABI should be. The ABI should be a way to standardize how common arguments are encoded when calling an app. Right now, there are apps that require additional transactions outside of the leading transactions. Thus, it is reasonable to assume that people would like to encode references to other transactions in the group. 

If we want a standard that enforces transactions should only be leading the current app call, that should come at a higher level than the ABI itself.

jannotti:
An ABI is not a way to describe arbitrary function behavior. (In our
context, transaction behavior.)

It is to provide a interface to which functions adhere. If they do, they
can be used together reasonably. If they don't, you can't combine them,
because you don't understand their behavior.

Imagine if a compiler compiled some functions to assume a particular
register is callee saved, and others assume it's caller saved. Or imagine
some pass a 16 byte structure argument on the stack, while others passed
the same structure in two registers. It would be chaos, and you could not
call between such functions freely. Thus, architectures have ABIs that
compilers adhere to. If you want to write a non-compliant function in
assembly, you can, but you are giving up any expectation that standard
tools can interact with you.

That can be fine! In the compiler world, maybe you are linking to such
functions only from other handcrafted functions. In our world, maybe you
are supplying your own sdk that creates  transaction (groups) in the form
you expect. But don't expect standard tools to understand your non-standard
choices.

The goal of the ABI is not to just allow any convention desired by the app
writers. That would mean apps can not be composed. It is to _provide_ a
convention that, if apps use it, allows your app to work with the rest of
the ecosystem.

An existing app that does things differently is not a justification to
change the ABI. Only if such an app was forced to do things differently
should we be thinking about an ABI change. And even then, we must be
careful that apps can easily understand how they were called and that no
single piece of data is interpreted differently by two different apps in
the same group. (the problem I described, in which one pay is seen by two
different transactions is one such example, and it is entirely unacceptable
to tell every app that it must implement elaborate checks to protect itself)

joe-p:
> That would mean apps can not be composed. It is to _provide_ a
convention that, if apps use it, allows your app to work with the rest of
the ecosystem.

The problem is that most of the ecosystem is not following the ABI because of composability issues that exist within it. One of which is how transaction args work. At the very least, I would propose that transactions are removed from method signatures which will allow for more flexible generic interfaces.

I still think there's value in encoding a reference to a grouped transaction, but it sounds like you'd want that to be at a higher level than the ABI. I think this ultimately means ABI-compliance will be meaningless for ecosystem integration and it will fully depend on that higher-level standard, which makes me question the value of the ABI in the first place.

emg110:
Hey @joe-p , Thank you so much for your great and impactful contributions , we community devs are very grateful for!

I have some issues to understand and relate to this ARC that I was hoping you can graciously make me comprehend correctly:
- What problem exactly is it going to solve ? Given the extent of change (it is a very big change) it needs to have very solid use-case to back it up , IMHO! Would you mind mentioning some TXN GROUP scenarios to be specifically in need of this ARC to be possible?
- I cannot understand how removing current ARC4 transaction types would do any good to simplify DX
- I cannot understand what is limiting about having the ABI call is the succession of all other group members! 
- Why no backward compatibility?

joe-p:
First I want to mention that I acknowledge this shouldn't be a breaking change if implemented. There's no reason we can't have a new encoding and still keep the old txn argument encoding as well. I will update this ARC in the near future to be backwards compatible. 

Second, I agree with JJ's reasoning for not including this in the ABI but I could imagine this still being useful as a higher-level standard.

Simple example that takes advantage of this encoding: Imagine there's a contract that implements a method `foo` which adds some data to box storage, thus requires a payment txn to the app address to cover MBR (and let's assume its 8KB box so inner transactions couldn't be used). Current ABI signature would be `foo(pay)void`. If one wanted to call this method repeatedly, the biggest transaction group they could do would only have 8 app calls: `[pay, appl, pay, appl, pay, appl, pay, appl, pay, appl, pay, appl, pay, appl, pay, appl]`. With the proposed encoding, however, one can increase the total number of atomic app calls by only having one payment and have 15 app calls that each introspect the group and that one payment: `[pay, appl, appl, appl, appl, appl, appl, appl, appl, appl, appl, appl, appl, appl, appl, appl, appl]`

Now this begs the questions of why would this be useful... and to that I don't have a great answer which is why I haven't continued to push this ARC.

@pbennett might have a better example.

PR #223, opened 2023-07-04
ARC-0200 draft

temptemp3:


joe-p:
If one of the goals is to mirror ERC20 to presumably allow for seamless bridging between the two chains, it would make more sense to use at least `uint256` for everything rather than `uint64`, no?

ARCs/arc-0200.md
```diff
@@ -0,0 +1,226 @@
...
+    },
+    {
+      "name": "symbol",
+      "desc": "Returns the symbol of the token",
+      "readonly": true,
+      "args": [],
+      "returns": { "type": "bytes[8]", "desc": "The symbol of the token" }
+    },
+    {
+      "name": "decimals",
```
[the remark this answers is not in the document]
temptemp3: The reason for treating `decimals` as an individual field is similar to how `name` or `symbol` are handled as separate fields. I prefer not to use `ufixed` encoded decimal because it would limit the implementation to a specific encoding method. The ABI specification only requires `decimals` to be returned as a uint8, but it allows the smart contract to store the decimal value in any desired format. The choice of implementation is left to the developer.

One possible implementation approach is to validate that `decimals` falls within an acceptable range. For compatibility with ASAs, the decimals value could be checked to be less than 19. For ERC20 compatibility, an implementation would check that it is less than 256.

Additionally, based on my understanding, Reach does not utilize `ufixed`. If a value is known to be less than 256 at compile-time, the compiler would automatically use uint8.

Considering the potential inconveniences and the availability of alternative methods, I don't believe using `ufixed` is necessary to achieve the desired purpose.

temptemp3:
> If one of the goals is to mirror ERC20 to presumably allow for seamless bridging between the two chains, it would make more sense to use at least `uint256` for everything rather than `uint64`, no?

If one of the goals is to mirror the ERC20 standard for seamless bridging between the two chains, it does raise the question of whether using uint256 for all fields would be more appropriate than uint64.

While aligning with ERC20 would provide a high level of compatibility, we must consider the trade-offs involved. Using uint256 for all fields would significantly increase the cost of global storage and boxes. This is an important factor to consider, as it could affect the overall cost and adoption of the token.

However, it is indeed possible to create a smart contract token of this type using uint256. By replacing uint64 with uint256 and ensuring that the decimals value remains within the range of 256, you could achieve alignment with ERC20.

At present, sticking with uint64 seems appropriate, considering the goals and requirements of the project. However, if there is a strong demand or a specific use case that necessitates the use of uint256, it might be worth considering submitting a separate proposal that outlines the alternative token specification using uint256.

joe-p:
> Using uint256 for all fields would significantly increase the cost of global storage and boxes. This is an important factor to consider, as it could affect the overall cost and adoption of the token.

There is no reason that the contract would need to store the full 32 bytes if not all bytes are used. Truncating the bytes for box storage but encoding as uint256 for arg/return values seems like a reasonable solution. The 32 bytes are there if you need them, but not a requirement. You can use uint64 if you want to save on opcodes and storage. It's entirely the developers choice. 

If the developer really wanted to use global storage, it is a 0.02 ALGO extra cost over uint64, but not sure why you'd use global storage when box storage is cheaper. 

> However, it is indeed possible to create a smart contract token of this type using uint256. By replacing uint64 with uint256 and ensuring that the decimals value remains within the range of 256, you could achieve alignment with ERC20.

A contract that wants to use uint64 integers can follow a standard that uses uint256 encoding, but not the other way around. Having one standard that supports both use cases seems better than having two different standards.

ARCs/arc-0200.md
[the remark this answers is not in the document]
joe-p: Yeah the more I think about it the more it makes sense to keep it as a separate method since clients will need an easy way to discover how many decimals are actually supported regardless of encoding. 

I will say though that I think the only language that should dictate the direction of standards is TEAL. Standards should dictate language features, not the other way around. Obviously it is extra dev effort and will take longer to implement/support across the ecosystem. But long term, having a standard that everyone feels is best suited for its purpose is better than having multiple evolving standards superseding each other.

pbennett:
If you want compatibility with ERC20, uint256 will be kind of a must.  The issues with bridging ETH tokens over to ASAs and decimals has been an issue.  PEPE has 18 decimals for eg.  Its supply with those decimals isn't representable in a uint64 - in fact if brought over with that many decimals it'd only have about $100K maximum mcap based on an earlier price of PEPE.

temptemp3:
@joe-p @temptemp3 
Thank you for the feedback! I appreciate your openness to reassess the approach. If there is a consensus and everyone agrees, updating the specification to use uint256 where applicable could be a beneficial decision. By aligning with uint256, you would ensure a higher level of compatibility with existing standards and enhance interoperability between chains.

joe-p:
Can you provide some more rationale to the timestamp even arguments? What is this expected to be? The timestamp the client sends the transaction or the time of the last block? If one is parsing the logs to read events, it would be trivial to implicitly get the time of the block, no?

temptemp3:
> Can you provide some more rationale to the timestamp even arguments? What is this expected to be? The timestamp the client sends the transaction or the time of the last block? If one is parsing the logs to read events, it would be trivial to implicitly get the time of the block, no?

The rationale to adding the timestamp event argument is that currently the library that I use to replay events does not include round time. So my workaround was to include the round time at confirmation as an argument of the event.

> What is this expected to be? 

It is to be expected as the round time at confirmation. 

> The timestamp the client sends the transaction or the time of the last block?

The timestamp at confirmation

> If one is parsing the logs to read events, it would be trivial to implicitly get the time of the block, no?

Yes, I may be able to do something while parsing the logs to figure the time of the block.

With that said it may be viable to remove the timestamp and recover it elsewhere.

ARCs/arc-0200.md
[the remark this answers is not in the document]
temptemp3: Here since this values may use UInt256, `uint8` may be preferable `decimals`

temptemp3:
After consideration, I have updated the spec draft to use uint256 and not include a timestamp argument in events. Now the spec is identical to ERC20.

ARCs/arc-0200.md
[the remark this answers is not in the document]
temptemp3: > Yeah the more I think about it the more it makes sense to keep it as a separate method since clients will need an easy way to discover how many decimals are actually supported regardless of encoding.

Yup. It is an individual field. Disregard my tangent regarding decimals and uint8. Similar to using uint256 for totalSupply, balances, and allowance values, the decimals method should return a uint8.

The rest are implementation details which I hope we all agree is out of the scope of the spec.

joe-p:
What the rationale for adding the `_` after each arg?

temptemp3:
To prevent overlap between reserved words in language of implementation and
to not be confused with _ prefix which by convention is supposed to imply
some meaning.

2023年7月21日(金) 5:04 Joe Polny ***@***.***>:

ARCs/arc-0200.md
```diff
@@ -0,0 +1,232 @@
...
+
+Ownership of a token by a zero address indicates that a token is out of circulation indefinitely, or otherwise burned or destroyed.
+
+The methods `transfer` and `transferFrom` method MUST error when the balance of `from_` is insufficient.
+The `transferFrom` method MUST error unless called by an approved spender as defined by an extension defined in this ARC.
+The methods `transfer` and `transferFrom` MUST emit a `Transfer` event that a transfer is successful.
+A `Transfer` event SHOULD be emitted, with `from_` being the zero address, when a token is first minted.
+A `Transfer` event SHOULD be emitted, with `to_` being the zero address, when a token is destroyed.
+
+The `Approval` event MUST be emitted when the `approve` method is called successfully.
```
jannotti, line 198: Do you want to "fix" the Ethereum misfeature here, and put `increaseAllowance` and `decreaseAllowance` into the standard?
https://www.adrianhetman.com/unboxing-erc20-approve-issues/

ARCs/arc-0200.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Rationale
+
+This specification is based on <a href="https://eips.ethereum.org/EIPS/eip-20">ERC-20</a>.
+
+### Core Specification
+
+The core specification identical to ERC-20 with the exception of:
+
+- Method and event arguments are appended, rather than prefixed, with \_.
```
jannotti, line 214: I don't think there's any reason to do this in an ARC.  There is no semantic meaning to the variable names here - if some HLL compiled the json down to something that could be called from the HLL and use the argument names from the standard, it ought to be the convention of the language in question how to handle name conflicts.

ARCs/arc-0200.md
```diff
@@ -0,0 +1,232 @@
...
+        }
+      ]
+    }
+  ]
+}
+```
+
+Ownership of a token by a zero address indicates that a token is out of circulation indefinitely, or otherwise burned or destroyed.
+
+The methods `transfer` and `transferFrom` method MUST error when the balance of `from_` is insufficient.
```
jannotti, line 192: `transfer` has no `from_` argument. (I get your point, but the language needs to be changed.)

ARCs/arc-0200.md
```diff
@@ -0,0 +1,232 @@
...
+The `transferFrom` method MUST error unless called by an approved spender as defined by an extension defined in this ARC.
```
jannotti, line 193: Why does this say "approved spender as defined by an extension defined in this ARC"? I think you're just referring to the `approve` system, right?

ARCs/arc-0200.md
```diff
@@ -0,0 +1,232 @@
...
+The methods `transfer` and `transferFrom` MUST emit a `Transfer` event that a transfer is successful.
```
jannotti, line 194:
```suggestion
The methods `transfer` and `transferFrom` MUST emit a `Transfer` event.
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,232 @@
...
+A `Transfer` event SHOULD be emitted, with `from_` being the zero address, when a token is first minted.
```
jannotti, line 195:
```suggestion
A `Transfer` event SHOULD be emitted, with `from_` being the zero address, when a token is minted.
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,232 @@
...
+A `Transfer` event SHOULD be emitted, with `to_` being the zero address, when a token is destroyed.
+
+The `Approval` event MUST be emitted when the `approve` method is called successfully.
+
+A value of zero for the `approve` method and the `Approval` event indicates no approval.
+When a `Transfer` event emits following the `transferFrom` method, this may also indicate that the approved value for the token is decremented.
```
jannotti, line 201: It seems this should be stronger.  The `transferFrom` method MUST decrement the current approval value for the `from` address of the `to` address's tokens.

I don't know if you want to require that a new `Approval` event is emitted.

ARCs/arc-0200.md
```diff
@@ -0,0 +1,232 @@
...
+The contract MUST allow multiple operators per owner.
```
jannotti, line 202: I think this means that, for example, owner A can approval both account's B and C simultaneously.  Is that right?  Does the MUST here imply an unbounded number of such approvals must be allowed?

ARCs/arc-0200.md
```diff
@@ -0,0 +1,232 @@
...
+
+# Algorand Smart Contract Token Specification
+
+## Abstract
+
+This ARC (Algorand Request for Comments) specifies an interface for tokens to be implemented on Algorand as smart contracts. The interface defines a minimal interface required for tokens to be held and transferred, with the potential for further augmentation through additional standard interfaces and custom methods.
+
+## Motivation
+
+Currently, most tokens in the Algorand ecosystem are represented by ASAs (Algorand Standard Assets). However, to provide rich extra functionality, it can be desirable implementing tokens as a smart contract instead. To foster an interoperable token ecosystem, it is necessary that the core interfaces for tokens be standardized.
```
jannotti, line 22:
```suggestion
Currently, most tokens in the Algorand ecosystem are represented by ASAs (Algorand Standard Assets). However, to provide rich extra functionality, it can be desirable to implement tokens as smart contracts instead. To foster an interoperable token ecosystem, it is necessary that the core interfaces for tokens be standardized.
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,232 @@
...
+          "name": "value_",
+          "desc": "The amount of tokens approve"
+        }
+      ]
+    }
+  ]
+}
+```
+
+Ownership of a token by a zero address indicates that a token is out of circulation indefinitely, or otherwise burned or destroyed.
```
jannotti, line 190: Is there a requirement that `totalSupply` reflect this convention?  Should `totalSupply` decrease after a transfer to the ZeroAddress?

jannotti (commented):
I think this is an important standard.  I hope my comments don't seem to nitpicky.

ARCs/arc-0200.md
temptemp3: I agree. It doesn't matter because when it comes does to it names are not translated in to the ABI where it counts. I'm going to revert this to the original state, ie. _from _to just like the ERC-20 spec.

ARCs/arc-0200.md
temptemp3: Yes, the `approve` system allows for multiple operators to spend on the behalf of other account if approved. I supposed we can revise it as follows.

approved spender -> `spender`

This may be more understandable since it matches the usage in the surrounding language and ABI spec.

ARCs/arc-0200.md
joe-p: What's the reason for `_` at all? (outside of aligning with ERC20)

temptemp3:
> I think this is an important standard. I hope my comments don't seem to nitpicky.

Not one bit. I appreciate your feedback and attention to detail.

ARCs/arc-0200.md
temptemp3: This one could be revised to be more clear.

The approval event must be emitted in case of `approve` and `transferFrom`

`transferFrom` emits both `Approval` and `Transfer`

`approve` emits only `Approval`

Going to come back to this one. I see your point. Thx!

ARCs/arc-0200.md
temptemp3: Yes, there is no limit on approvals for any account.

For example,

approval[A][B] and
approval[A][C] 

are mutually exclusive.

On another note

approval[D][B] and
approval[D][C] 

MUST be allowed

that is there are no bounds on approval relations.

---

> I think this means that, for example, owner A can have an outstanding approval for both account's B and C to spend their tokens simultaneously. Is that right?

Yes, if A has approval for both B and C, B and C can be spent by A simultaneously.

---

Hope that helped

ARCs/arc-0200.md
temptemp3: Currently no. In my mind, `totalSupply` is immutable. That is total supply starts at the mint address. Later on it may find itself in the zero address. However, depending on how you look at it circulating supply may differ. Circulating supply could be:

1) Tokens outside of mint address and zero address * I like this one

2) Tokens outside of zero address

The rational here is that a total supply reflecting the convention can be computed by

totalSupply - balanceOf(zeroAddress) if needed

on and off chain

temptemp3:
@jannotti Thank you for making a thorough pass through. I appreciate it!

ARCs/arc-0200.md
temptemp3: I may want to fix this issue from the get-go because the ARC-200 spec implementation may easily be ported to a EVM based chain with this fix already applied. I haven't heard of any issues with front running on Algorand yet but this seems like a great way to future-proof the spec. Will be considering.

jannotti:
I think we should specify some more. It's common to think of the ABI JSON as sufficient to document a contract. And I wish it was. But it's not, at the moment. For example, I'm sure it is you intent that ARC-200 authors use box storage to store balances and approvals.  Therefore, a caller will need to set the appropriate box refs to make any of these calls work.

While I would sort of like it if ARC200 contracts were free to use any storage scheme behind the scenes, I think there would be a benefit to standardization is the ARC constrained them and specified the box refs.

So, for example, since balances are keyed by owner, the `transfer` method is certainly going to need at least two box refs - wherever the balance is stored.  As written, one ARC200 contract might use the box named by the 32 bytes of the address in question, so the caller would need to add `{0, <sender address bytes>}` and `{0, to address bytes}`. But another contract, perhaps thinking to the future, decides to prefix their balance records with "B:" or something like that.

This means callers can't know how to invoke a random ARC-200 contract.

While I would not go so far as to specify the format of the box contents, I think the ARC should specify the box names it uses. And, I think that all ARCs should start doing this, so we should use prefixes to distinguish them. I propose, for example, that this ARC specifies two box names.  For balances:

"ARC200:<32 bytes of address>" must be added to transfers and balance checks.
"ARC200:<32 bytes of owner><32 bytes of spender>" must be added to `transferFrom` and `approve`

Perhaps this should be an ARC about this convention.

ARCs/arc-0200.md
temptemp3: @jannotti I am on the side of not adding `increaseAllowance` and `decreaseAllowance` to the standard for simplicity sakes. I do think that implementations may add non-standard methods similar to the way OpenZeppelin does it with ERC-20 (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol)

I have a working ARC200 implementation that methods like  `increaseAllowance` and `decreaseAllowance` for convenience.

ARCs/arc-0200.md
temptemp3: Yes. I will exterminate the tick ; )

ARCs/arc-0200.md
temptemp3: Here `from` is implied as the signer of the transaction

ARCs/arc-0200.md
temptemp3: Reworded here
https://github.com/algorandfoundation/ARCs/pull/223/commits/125f2342c4ab3bbb3e73918223811840d2fdfb5e

ARCs/arc-0200.md
temptemp3: Revision here
https://github.com/algorandfoundation/ARCs/pull/223/commits/a9174bc215e756fb12858db4c0657cc0b73d2527

Reworded as `spender` approved by `owner` instead of approved spender. 
Added note that `from` is `owner` in case of `transfer` method

ARCs/arc-0200.md
temptemp3: See also
https://github.com/algorandfoundation/ARCs/pull/223/commits/a9174bc215e756fb12858db4c0657cc0b73d2527

temptemp3:
> I think we should specify some more. It's common to think of the ABI JSON as sufficient to document a contract. And I wish it was. But it's not, at the moment. For example, I'm sure it is your intent that ARC-200 authors use box storage to store balances and approvals. Therefore, a caller will need to set the appropriate box refs to make any of these calls work.
> 
> While I would sort of like it if ARC200 contracts were free to use any storage scheme they'd like behind the scenes, I think there would be a benefit to standardization if the ARC constrained them and specified the box refs.
> 
> So, for example, since balances are keyed by owner, the `transfer` method is certainly going to need at least two box refs - wherever the balances are stored. As written, one ARC200 contract might use the box named by the 32 bytes of the address in question, so the caller would need to add `{0, <sender address bytes>}` and `{0, to address bytes}`. But another contract, perhaps thinking to the future, decides to prefix their balance records with "B:" or something like that.
> 
> This means callers can't know how to invoke a random ARC-200 contract.
> 
> While I would not go so far as to specify the format of the box contents, I think the ARC should specify the box names it uses. And, I think that all ARCs should start doing this, so we should start using prefixes to distinguish them. I propose, for example, that this ARC specifies two box names.
> 
> "ARC200:<32 bytes of address>" must be added to transfers and balance checks. "ARC200:<32 bytes of owner><32 bytes of spender>" must be added to `transferFrom` and `approve`
> 
> Perhaps there should be an ARC about this convention.

It seems like there should be an ARC about this convention. I would not go as far as standardizing what to do with boxes in the spec. I think the spec should not include anything regarding boxes.

What I can tell you about the implementation I have on hand is

index
**0** **Address | UInt256**
**1**  **Address Address | UInt256**

where index 0 is for balances and index 1 is for allowances.

pbennett:
I don't think Box references belong in an interface spec spec - because now you also need to document the *storage* requirements of the boxes themselves and which operations will create or replace with larger size so that callers can fund the MBR prior to calling.  This feels like recreating a new version of what's being taken out of ARC-4.
To construct a single transaction to call the contract using the presumed 'simple' box references (not accounting for common things like calculated hashes as box keys), how do you call that method which has to create a box without the MBR being accounted prior to the call?  You don't.

Required 'at runtime' foreign references for apps, assets, accounts, boxes (and even multiple empty box references depending on how much data has to be touched) and the requirements specific to an interface implementation should be independent.   If I call contract A implementing ARC200, contract B may implement it completely differently and that will certainly include boxes.  Boxes are opaque and accessible only by the owning contract itself so why would that implementation detail be surfaced and made part of its interface specification?

The future simulate endpoint (which hopefully can give us required MBR adjustments as well) should solve all these builder nightmares.  I'm hopeful at least.

SilentRhetoric:
It seems to me that simulate is the real answer to enable calling other people’s apps, given the challenges Patrick mentioned.  My app uses hashes as box references and also strictly enforces payment to the app to cover incremental box MBR, so it is pretty much impossible to call successfully unless you know exactly how my app works or I provide an SDK.

temptemp3:
> I tried to implement it in pyteal, but we are facing issues with the `from` keyword that is already reserved. We cannot use `sender` either because it is a reserved keyword for calling applications with SDKs.

I will go ahead and add a suffix to the argument names. For example, `from` becomes `from_`,

jannotti:
> I tried to implement it in pyteal, but we are facing issues with the `from` keyword that is already reserved. We cannot use `sender` either because it is a reserved keyword for calling applications with SDKs.

I don't see why this should be fixed in the ARC.  A pyteal library calling this can call the arguments anything it wants.

I don't want to beat a dead horse, but why are people so focused on this "implementation detail" even though it has no effect on the ability of calling code to invoke such a contract, but are completely ignoring the fact that no code can call any of these contracts unless the box and fee payments are specified?

PS. Fee payments will _not_ be fixed by `simulate`.  Simulate might tell you how the MBRs change, but only the spec can tell you if the caller is supposed to pay for those changes, and where that payment transaction should appear.

temptemp3:
I agree. It is not important to me either even if we are talking about implementation. I'm going to revert the change because it shouldn't matter.

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+          "name": "to",
+          "desc": "The destination of the transfer"
+        },
+        {
+          "type": "uint256",
+          "name": "value",
+          "desc": "Amount of tokens to transfer"
+        }
+      ],
+      "returns": { "type": "bool", "desc": "Success" }
```
jasonpaulos, line 97: Do you see any advantage in making `transfer` and `transferFrom` return void instead of booleans? My impression is that for ERC-20 implementations, the safer thing to do is to only ever return true or error if the operation isn't possible. Otherwise it's very important that callers check the returned value, which is just another opportunity for a bug.

temptemp3: It does trivially return True. To be on the safe side you would check that it does despite the fact that in practice the only alternative is to throw an error. 
I think there is no advantage of returning void instead of boolean. Same problem more uncertainty.
I know we are not going into implementation but if we were I would advise having both methods return true if not error and don't even bother having code returning false.

temptemp3: In other words, it is not written in the spec but these methods only return true, which is a bool

jasonpaulos: Then why not write into the spec that the methods should return only true? Or better yet, remove all uncertainty and have it return void?

True, this would be a departure from the ERC-20 spec, but a very small one that's easily bridged.

joe-p: I think the argument for `bool` is that one can attempt to do a transfer which is a feature that isn't otherwise possible. The counterargument is that one should just check `allowance` before attempting a transfer.

temptemp3: I think bool is fine the way it is. I don't think we should bake into the spec the fact that on the underlying network if something does not match up, assertions fail, it will crash and burn.

SudoWeezy:
> > I tried to implement it in pyteal, but we are facing issues with the `from` keyword that is already reserved. We cannot use `sender` either because it is a reserved keyword for calling applications with SDKs.
> 
> I don't see why this should be fixed in the ARC. A pyteal library calling this can call the arguments anything it wants.

Maybe I am missing something but, if we use don't fix it in the ARC,  people will not be able to compile the application using Algokit & Beaker. In my opinion, it's an issue.

jasonpaulos:
> Maybe I am missing something but, if we use don't fix it in the ARC, people will not be able to compile the application using Algokit & Beaker. In my opinion, it's an issue.

Implementing contracts do not need to use the exact same method argument names as the interface spec. These names are informational only. It would be perfectly fine to use some variant like `_from` in PyTeal.

temptemp3:
I have a hunch that our algokit & beaker boys will be okay changes the argument names as mentioned by @jasonpaulos

temptemp3:
We should confirm that this an issue. Ideally, the argument names should not be important, the same way you can call function arguments whatever you like. 
If push comes to shove, we could do something in this spec if it is agreed.

temptemp3:
I do not see any argument names in this ABI
`_reachp_0((uint64,address,address,byte,(uint64,byte[32],byte[8],uint256)))void`, `_reachp_2((uint64,(byte,byte[96])))void`, `approve(address,uint256)byte`, `deleteAllowanceBox(address,address)byte`, `deleteBalanceBox(address)byte`, `destroy()void`, `transfer(address,uint256)byte`, `transferFrom(address,address,uint256)byte`

jannotti:
> It seems like there should be an ARC about this convention. I would not go as far as standardizing what to do with boxes in the spec. I think the spec should not include anything regarding boxes.
> 
> What I can tell you about the implementation I have on hand is
> 
> index **0** **Address | UInt256** **1** **Address Address | UInt256**
> 
> where index 0 is for balances and index 1 is for allowances.

Can you say more about what you mean by "index" here? I thought I was reading you to say that you balance boxes are named by the sender address, and the allowance boxes by the concatenation of owner and spender.  I follow that (if that's what you mean). But I don't follow how "index" of any kind enters into the explanation.

temptemp3:
> > It seems like there should be an ARC about this convention. I would not go as far as standardizing what to do with boxes in the spec. I think the spec should not include anything regarding boxes.
> > What I can tell you about the implementation I have on hand is
> > index **0** **Address | UInt256** **1** **Address Address | UInt256**
> > where index 0 is for balances and index 1 is for allowances.
> 
> Can you say more about what you mean by "index" here? I thought I was reading you to say that you balance boxes are named by the sender address, and the allowance boxes by the concatenation of owner and spender. I follow that (if that's what you mean). But I don't follow how "index" of any kind enters into the explanation.

No problem. I wasn't aware at the time but it makes sense now. In an application created by reach box names are generated. See reference here https://docs.reach.sh/networks/#p_8

pbennett:
If this to define an 'interface', then foreign references (which are a *runtime* requirement specific to an *implementation*) are independent and don't belong here.  Defining specific box names as part of the interface would 100% tie the hands of all implementations.  Each implementation could define what its requirement are in its own spec or docs - but ideally, this shouldn't be necessary either.

Algorand dApps currently have to provide SDKs or APIs for callers in order for the 'right' transactions to be created to call their contracts because of foreign references, MBRs, and specific atomic transaction requirements.  
No current interface specification covers this and simply adding box names to this doesn't cover it either.

Hopefully the future simulate endpoint will be able to handle the foreign reference issues at a minimum (and also MBR impacts through asset/contract/box adds).

jannotti:
> If this to define an 'interface', then foreign references (which are a runtime requirement specific to an implementation) are independent and don't belong here.

Please stop saying "don't belong here" unless you have a precise definition of "interface" in mind, because in my definition, these things belong, and I will explain why. This document is an attempt to make it possible to invoke a class of similar contracts in a uniform way.  It doesn't matter if _you_ call some part of it "interface" and some part of it "implementation". They all go toward telling callers the exact bytes that must be sent so that the contract does something useful.

If the arc said, "All ARC-200 implementations need a `transfer` method, but they may choose to call it anything they'd like", you'd be outraged! You'd say you *must* tell people what to call the method.  That's "interface"! And some jabroney might come along and say, "Don't tie my hands, man, I want to call it xfer because it's cool". Wallets, explorers, and dapps will just have to call me up on the phone and ask me what I called my transfer function.  I also reserve the right to change the name once in a while, because I don't want my hands tied!

That jabroney would be wrong. If wallets and explorer implementors can't read this ARC and expect to call ARC-200 contracts, then this spec is literally, and I do mean _literally_, not a spec.  Whether you like it or not, the box refs and the possible required payments _are_ part of these contracts' interface - you need to get them right, or the contract doesn't work.

I want to ask @temptemp3 a serious, non-rhetorical question, as he initiated this ARC.  Do you want wallets to be able to show balances for ARC-200 contracts?  How do you propose they do that?  Must they continue to use hand crafted SDKs, a separate SDK for each such asset?

Patrick claims that SDKs are always required today and:
> No current interface specification covers this and simply adding box names to this doesn't cover it either.

Then stop writing what Patrick is calling an "interface specification" and write an actual specification that *does* allow a wallet to call all members of the class you are defining. In this case, it would mean writing the box names down AND writing down how the payment model must work.  After that, wallets can call every single conforming ARC-200 contract.  That's what a specification is for!  (And Again - `simulate` will NEVER fix the payment model question. It must be specified!)

This ARC says: "To foster an interoperable token ecosystem, it is necessary that the core interfaces for tokens be standardized."

I beg you to do so.

temptemp3:
Sure @jannotti 
Very constructive feedback, I appreciate it. Constructive feedback from everyone ; )
Allow me to process and come back with a serious response as soon as I can. Thx!

joe-p:
>  Do you want wallets to be able to show balances for ARC-200 contracts? How do you propose they do that? Must they continue to use hand crafted SDKs, a separate SDK for each such asset?

Is it not the plan for simulate to be able to run method calls without specific resources? If so, then wallets can simply use simulate to get balances and know what references something like transfer would need.

jannotti:
> > Do you want wallets to be able to show balances for ARC-200 contracts? How do you propose they do that? Must they continue to use hand crafted SDKs, a separate SDK for each such asset?
> 
> Is it not the plan for simulate to be able to run method calls without specific resources? If so, then wallets can simply use simulate to get balances and know what references something like transfer would need.

FFS, then do you want wallets to be able to *transfer* these tokens?

Just do the work. Specify the boxes and the payment model.

joe-p:
> Specify the boxes and the payment model., don't wait for simulate to fix half of the problem.

Payments and boxes are two completely separate issues. Simulate completely fixes the box problem, so why would we not take that into account when writing an ARC? Or do you think simulate is not a solution to the boxes problem? If so, why?

algoanne:
> Payments and boxes are two completely separate issues. Simulate completely fixes the box problem, so why would we not take that into account when writing an ARC? Or do you think simulate is not a solution to the boxes problem? If so, why?

I'm glad people are enthusiastic, but I think it's a little premature to assume that simulate will "fix the box problem". Yes, there is an idea that we might start using simulate without naming resources before submitting any transaction. But we (or at least I) haven't thought it all the way through, let alone planned out and executed the work. Will people really want to use simulate before every app call they make? Doing so will introduce a delay, so perhaps if people know the foreign refs they would prefer to explicitly fill them in rather than have simulate do it. I can imagine that high frequency traders will certainly prefer to fill things in manually when possible. If they want to trade ARC-200 assets they will need to know how to fill those arrays. If people do adopt the pattern of using simulate all the time, what should simulate do exactly? These are answerable questions, I'm just a little nervous about the assumptions being made in the context of ARC-200 design.

joe-p:
> Will people really want to use simulate before every app call they make? 

To have true composability with ARC200, yes. What about contract that *might* interact with an ARC200 contract. Should the caller always just add the standardize box even if its not used?

> Doing so will introduce a delay, so perhaps if people know the foreign refs they would prefer to explicitly fill them in rather than have simulate do it. I can imagine that high frequency traders will certainly prefer to fill things in manually when possible. 

And those people can still fill in those fields manually. Just because the box key isn't standardized doesn't mean it can't be known ahead of time. *most* implementations will probably use simple box keys. But enforcing a standardized box key limits potential implementations

> If people do adopt the pattern of using simulate all the time, what should simulate do exactly?

Presumably let an app be called without a panic on an unavailable resource. Returning the references would obviously be nice, but even without that you could have methods that manually returns references. 

I think enforcing boxes in the standard is a mistake because it limits potential implementations. Maybe the ecosystem becomes a mess for a bit, but if it's really a problem an additive standard could be proposed to improve the situation. Maybe the solution is simulate, maybe it's something else. 

We've already seen what happens when standard aren't suitable for use cases. People just ignore the standard and we're in a no better situation than before. 

I think it'd be reasonable for the standard to RECOMMEND box names, but don't think it should be enforced.

algoanne:
>I think it'd be reasonable for the standard to RECOMMEND box names, but don't think it should be enforced.

That sounds like a potentially reasonable compromise.

joe-p:
I have proposed a new arc in #230 that standardizes app reference discoverability. It will essentially be deprecated once we have unnamed resources with simulate, but it seems like a good intermediate solution to me.

jannotti:
> I have proposed a new arc in #230 that standardizes app reference discoverability. It will essentially be deprecated once we have unnamed resources with simulate, but it seems like a good intermediate solution to me.

I love it.  I have been thinking along similar lines. I'll comment there on some additional things this might do.

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+
+A smart contract token that is compliant with this standard MUST implement the following interface:
+
+```json
+{
+  "name": "ARC-200",
+  "desc": "Smart Contract Token Base Interface",
+  "methods": [
+    {
+      "name": "name",
```
SudoWeezy, line 38:
```suggestion
      "name": "arc200_name",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+      "desc": "Returns the name of the token",
+      "readonly": true,
+      "args": [],
+      "returns": { "type": "bytes[32]", "desc": "The name of the token" }
+    },
+    {
+      "name": "symbol",
```
SudoWeezy, line 45:
```suggestion
      "name": "arc200_symbol",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+      "desc": "Returns the symbol of the token",
+      "readonly": true,
+      "args": [],
+      "returns": { "type": "bytes[8]", "desc": "The symbol of the token" }
+    },
+    {
+      "name": "decimals",
```
SudoWeezy, line 52:
```suggestion
      "name": "arc200_decimals",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+      "desc": "Returns the decimals of the token",
+      "readonly": true,
+      "args": [],
+      "returns": { "type": "uint8", "desc": "The decimals of the token" }
+    },
+    {
+      "name": "totalSupply",
```
SudoWeezy, line 59:
```suggestion
      "name": "arc200_totalSupply",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+      "desc": "Returns the total supply of the token",
+      "readonly": true,
+      "args": [],
+      "returns": { "type": "uint256", "desc": "The total supply of the token" }
+    },
+    {
+      "name": "balanceOf",
```
SudoWeezy, line 66:
```suggestion
      "name": "arc200_balanceOf",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+          "desc": "The address of the owner of the token"
+        }
+      ],
+      "returns": {
+        "type": "uint256",
+        "desc": "The current balance of the holder of the token"
+      }
+    },
+    {
+      "name": "transfer",
```
SudoWeezy, line 82:
```suggestion
      "name": "arc200_transfer",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+        {
+          "type": "uint256",
+          "name": "value",
+          "desc": "Amount of tokens to transfer"
+        }
+      ],
+      "returns": { "type": "bool", "desc": "Success" }
+    },
+    {
+      "name": "transferFrom",
```
SudoWeezy, line 100:
```suggestion
      "name": "arc200_transferFrom",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+        {
+          "type": "uint256",
+          "name": "value",
+          "desc": "Amount of tokens to transfer"
+        }
+      ],
+      "returns": { "type": "bool", "desc": "Success" }
+    },
+    {
+      "name": "approve",
```
SudoWeezy, line 123:
```suggestion
      "name": "arc200_approve",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+      "desc": "Approve spender for a token",
+      "readonly": false,
+      "args": [
+        { "type": "address", "name": "spender" },
+        { "type": "uint256", "name": "value" }
+      ],
+      "returns": { "type": "bool", "desc": "Success" }
+    },
+    {
+      "name": "allowance",
```
SudoWeezy, line 133:
```suggestion
      "name": "arc200_allowance",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+      "args": [
+        { "type": "address", "name": "owner" },
+        { "type": "address", "name": "spender" }
+      ],
+      "returns": { "type": "uint256", "desc": "The remaining allowance" }
+    }
+  ],
+  "events": [
+    {
+      "name": "Transfer",
```
SudoWeezy, line 145:
```suggestion
      "name": "arc200_Transfer",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+        },
+        {
+          "type": "uint256",
+          "name": "value",
+          "desc": "The amount of tokens transferred"
+        }
+      ]
+    },
+    {
+      "name": "Approval",
```
SudoWeezy, line 166:
```suggestion
      "name": "arc200_Approval",
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+        }
+      ]
+    }
+  ]
+}
+```
+
+Ownership of a token by a zero address indicates that a token is out of circulation indefinitely, or otherwise burned or destroyed.
+
+The methods `transfer` and `transferFrom` method MUST error when the balance of `from` is insufficient. In the case of the `transfer` method, from is implied as the `owner` of the token.
```
SudoWeezy, line 192:
```suggestion
The methods `arc200_transfer` and `arc200_transferFrom` method MUST error when the balance of `from` is insufficient. In the case of the `arc200_transfer` method, from is implied as the `owner` of the token.
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+The `transferFrom` method MUST error unless called by an `spender` approved by an `owner`.
```
SudoWeezy, line 193:
```suggestion
The `arc200_transferFrom` method MUST error unless called by an `spender` approved by an `owner`.
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+The methods `transfer` and `transferFrom` MUST emit a `Transfer` event.
```
SudoWeezy, line 194:
```suggestion
The methods `arc200_transfer` and `arc200_transferFrom` MUST emit a `Transfer` event.
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+A `Transfer` event SHOULD be emitted, with `from` being the zero address, when a token is minted.
```
SudoWeezy, line 195:
```suggestion
A `arc200_Transfer` event SHOULD be emitted, with `from` being the zero address, when a token is minted.
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+A `Transfer` event SHOULD be emitted, with `to` being the zero address, when a token is destroyed.
```
SudoWeezy, line 196:
```suggestion
A `arc200_Transfer` event SHOULD be emitted, with `to` being the zero address, when a token is destroyed.
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+
+The `Approval` event MUST be emitted when an `approve` or `transferFrom` method is called successfully.
```
SudoWeezy, line 198:
```suggestion
The `arc200_Approval` event MUST be emitted when an `arc200_approve` or `arc200_transferFrom` method is called successfully.
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+
+A value of zero for the `approve` method and the `Approval` event indicates no approval.
```
SudoWeezy, line 200:
```suggestion
A value of zero for the `arc200_approve` method and the `arc200_Approval` event indicates no approval.
```

ARCs/arc-0200.md
```diff
@@ -0,0 +1,231 @@
...
+The `transferFrom` method and the `Approval` event indicates the approval value after it is decremented.
```
SudoWeezy, line 201:
```suggestion
The `arc200_transferFrom` method and the `arc200_Approval` event indicates the approval value after it is decremented.
```

SudoWeezy (changes requested):
To be consistent with https://github.com/algorandfoundation/ARCs/pull/219
Please accept suggested changes

BunsanMuchi:
Some of the pushback I've seen in the community has been w/r to opt-ins. How anyone can airdrop an ERC-20, whereas for an ASA the user has to explicitly state they're willing to accept the asset. The airdropping could be annoying, a baiting/phishing attempt or potentially create an unforeseen tax-liability. 

Is there any way we could improve on this by following the opt-in ethos?

 I understand if it's seen as unimportant due to keeping consistency with ERCs but figured it'd be worth considering it.

jannotti:
> Is there any way we could improve on this by following the opt-in ethos?
> 
> I understand if it's seen as unimportant due to keeping consistency with ERCs but figured it'd be worth considering it.

I'm firmly in _favor_ of allowing airdrops without opt-in for these assets.  It's not clear to me what the legal ramifications of airdropping tokens in Ethereum is.  Have you really received property just because someone wrote down that you own it? I'm also not sure what the UI/UX should be for wallets when someone could be "spammed" by airdrops.

But I do know that, in both cases, the entire crypto ecosystem is getting along in the context of a model with no opt-ins.  The fact that Algorand is different has caused more than one project to throw up their hands and decide they don't want to have to settle those difficult questions in a new context.

So I think there's room for a token standard that operates the way "everyone else" is doing it.

BunsanMuchi:
Yeah, I understand that it's not really clear. I do feel sending a message of approval can act as an acceptance of ownership, but then again an opt-in isn't really that and there's just no real standard for this. 

Maybe the answer is more Off-Chain, for example on other chains you tend to add a token to your wallet to have it reflected in your UI, and that's good enough, it should avoid most of the baiting attempt. Plus, since I imagine indexing balances is going to be a pain for most wallets I feel they'll end up doing something like this and avoid the other route of displaying all balances of ARC-200 due to the overhead.  

I agree with the your assessment of there being room for a standard to be as close as "what everyone else is doing" as possible, so it's fair to just dismiss this.

temptemp3 (commented):
Accept suggested changes

SudoWeezy (approved):

PR #229, opened 2023-07-25
ARC-50: Delegated Asset Opt Ins

joe-p:
A reference implementation can be seen here: https://github.com/joe-p/delegated_optin_arc/tree/main

This implementation will be added to the ARC repo and this ARC once it's reviewed and finalized.


ARCs/arc-0050.md
```diff
@@ -0,0 +1,210 @@
...
+This ARC provides a standardized logic signature intended to be used to delegate asset opt ins. An atomic app call is used to provide modular functionality.
+
+## Motivation
+Asset opt ins has been a long-standing point of friction for the Algorand ecosystem, especially for onboarding new users. As such, there needs to be a standardized way to allow delegation of opt ins while still allowing users to have control over their accounts.
+
+## Specification
+
+There are two logic signature programs that apply to this ARC. *Open Opt In Program* refers to a logic signature program that can be used by anyone to opt the signing account into any asset. *Address Opt In Delegation* refers to a logic signature program that can only be used by a specific account to opt the signing account into an asset. 
+
+Both logic signatures must be sent alongside an appcall to an app that implements a specific ABI method `openOptIn(pay,axfer)void` or `addressOptIn(pay,axfer)void`. These methods should determine whether the opt-in is allowed.
```
jannotti, line 24: Why have the two methods? At one level, the reason is obvious - to constraint opt-ins to a given sender. But why is _that_ particular check important enough to create complexity here? Isn't it the app's job to check things?  It could check against a list of addresses, or it could not.  Seems like the user signs this logicsig because it trusts the app to do the checking.  Don't do any checking here.

joe-p:
Because of ARC59 (#285) I am withdrawing this ARC.

PR #230, opened 2023-07-28
ARC-51: Method Reference Discovery

joe-p:
This ARC proposes a way to discover references required for a method. Eventually this will be superceeded by simulation endpoint's unnamed resources functionality, but it serves as a viable intermediate solution for ARCs like ARC200 (#223) and ARC72 (#147)

### DIFF
--- a/ARCs/arc-draft_get_refs.md
+++ b/ARCs/arc-draft_get_refs.md
@@ -19,7 +19,7 @@ As of the time of this ARC, it can be hard to know which resources a caller need
 
 ### Simulation Consideration
 
-It should be noted that it is currently planned for the algod simulate endpoint to allow readonly execution of a method without providing in references. The response to this endpoint will return the necessary resources. Once this functionality is availible, this ARC will no longer be needed. The primary intent of this ARC is to serve as an intermediate solution.
+It should be noted that it is currently planned for the algod simulate endpoint to allow readonly execution of a method without providing references. The response to this endpoint will return the necessary resources. Once this functionality is availible, this ARC will no longer be needed. The primary intent of this ARC is to serve as an intermediate solution.
 
 ## Specification
 If an application has a method and the contract wants to make the required resources for calling the method discoverable, it **MUST** implement a readonly method with the same exact signature with an `arcXXXX_` prefix and a return type of `(address[],uint64[],uint64[],(uint64,byte[]))`. 
--- a/ARCs/arc-draft_get_refs.md
+++ b/ARCs/arc-draft_get_refs.md
@@ -1,0 +1,60 @@
+---
+arc: <to be assigned>
+title: Method Reference Discovery
+description: A standardized way contracts can reveal which references will be needed to call a specific method
+author: Joe Polny (@joe-p)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: ARC
+created: 2023-07-28
+requires (*optional): 4
+--- 
+
+## Abstract
+A contract caller needs to know which resources a contract needs to be availible before calling a method. This ARC proposes a standard way to make that information discoverable.
+
+## Motivation
+As of the time of this ARC, it can be hard to know which resources a caller needs to make availible when calling an application. The current solution typically involves proprietary SDKs which can make app usage and composability difficult.
+
+### Simulation Consideration
+
+It should be noted that it is currently planned for the algod simulate endpoint to allow readonly execution of a method without providing in references. The response to this endpoint will return the necessary resources. Once this functionality is availible, this ARC will no longer be needed. The primary intent of this ARC is to serve as an intermediate solution.
+
+## Specification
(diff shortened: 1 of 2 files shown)


joe-p:
I think a similar approach could be used for fee and MBR discoverability as well, but I think that should be a separate ARC since the reference discoverability will eventually be provided algod.

jannotti:
This is great.  I think you might consider a couple additions.

Report information on box _sizes_ so that callers can know whether they need to pack extra box refs for quota.  You might also want to report whether a box is being read or written, as that can matter.

Report "cross-product" requirements separately. With resource sharing coming into play, there's a difference between needing account A and asset X, vs needing holding AxX, as the caller could put A and X in separate transactions.

Report on payment expectations.  If the contract will only work if the caller gives it some extra algos for MBR, it would be nice to report that. I don't think it's worth the complexity of coming up with some scheme where the discovery call explains where the payment has to go - I would say this is only for contracts that don't care about that, they aren't explicitly *looking* for the algo payment, they just won't work if they don't have enough balance.

Consider reporting each requirement as `log` statement, as it will be more extensible. Each such requirement might be something like `<type bye> || <requirement data>` so that, for example, you get `0x01 <asset id>`, `0x01 <another asset>`, `0x2 <app id>` etc.

jannotti:
>>You might also want to report whether a box is being read or written, as that can matter.
>How does this affect references?

Suppose you intend to create a box named X of size 5k.  You also intend to read a box named Y, of size 4k (but not modify it).  You only need 5 box refs, one for X, one for Y, and three empties to increase your read/write budget from 2k up to 5k.

But if you were writing to Y, you'd need 7, as you'd need a 9k write budget.

jannotti:
>>Consider reporting each requirement as log statement, as it will be more extensible. Each such requirement might be something like || so that, for example, you get 0x01 , 0x01 , 0x2 etc.

>Love this approach. Makes it easier to add future information as well if we wanted to without affecting the method signature.

If you go this way, I'd avoid arcXXX_ in the prefix. Instead, something generic like "explain_",  as the prefix. ARCs in the future might simply add a new thing a method can report.  The new tag byte would be defined in a new arc, and I wouldn't want to imply they should create a _new_ `arcYYY_method()` to convey it.

ARCs/arc-draft_get_refs.md
```diff
@@ -0,0 +1,60 @@
...
+As of the time of this ARC, it can be hard to know which resources a caller needs to make availible when calling an application. The current solution typically involves proprietary SDKs which can make app usage and composability difficult.
+
+### Simulation Consideration
+
+It should be noted that it is currently planned for the algod simulate endpoint to allow readonly execution of a method without providing references. The response to this endpoint will return the necessary resources. Once this functionality is availible, this ARC will no longer be needed. The primary intent of this ARC is to serve as an intermediate solution.
+
+## Specification
+If an application has a method and the contract wants to make the required resources for calling the method discoverable, it **MUST** implement a readonly method with the same exact signature with an `arcXXXX_` prefix and a return type of `(address[],uint64[],uint64[],(uint64,byte[]))`. 
+
+The return value corresponds to arrays containing the required account, application, asset, and box references respectively.
```
jasonpaulos, line 27: It would be nice to provide clarity in the case of a method which takes reference type arguments.

I.e. does its `arcXXXX_` function need to return the already specified reference types, or can it omit them?

ARCs/arc-draft_get_refs.md
```diff
@@ -0,0 +1,60 @@
...
+N/A
+
+## Test Cases
+N/A
+
+## Reference Implementation
+
+### ARC200 example
+
+In this example, let's say `arc200_totalSupply()uint256` requires two box refences `"baseSupply"` and `"supplyMultiplier"`, each encoded as `byte[]`.
```
jasonpaulos, line 46: nit: it would be nice to show an example with arguments to reenforce the restriction that the `arcXXXX_` must be declared with the same set of arguments.

jasonpaulos:
+1 to the event mechanism idea. (Do we have an ARC that defines this already?)

pbennett:
When contract A needs to call contract B as part of its implementation, would the expectation be that A hard-codes B's requirements as part of what it returns for this discovery method?  
Ideally, it would get B's requirements and add them to its own, but if it requires A to 'parse an ARC28' message then I think that might get awkward (or at least make sure it's part of delivered tooling for developers to use in pyteal, etc.)

joe-p:
> When contract A needs to call contract B as part of its implementation, would the expectation be that A hard-codes B's requirements as part of what it returns for this discovery method?

The contract can handle this however it sees fit. Could be hard-coded or could use the `requirements_` method of the inner app if implemented. 

> Ideally, it would get B's requirements and add them to its own, but if it requires A to 'parse an ARC28' message then I think that might get awkward (or at least make sure it's part of delivered tooling for developers to use in pyteal, etc.)

The latest version uses the event specification, so if languages support events then it will trivial to parse the events. To be fair, I don't believe events are supported by PyTeal right now

emg110:
When this ARC would be living standard?

cusma:
@joe-p: is there any intention or plan to push this ARC forward?

Roseymr (approved):

PR #233, opened 2023-08-15
ARC-34 Update Threshold

ShaperOfEntropy:
With the current calculation of the threshold for whether a proposal gathered sufficient number of votes to be approved, it is possible that a large portion of funds end up undistributed if any voters forget to cast their votes since their votes still count towards the total. Moreover, voters that did not vote become ineligible in the xGov program but their Algo remains in the Term Pool. Therefore, only the part of the votes that belong to xGovs still eligible at the end of the voting period should count towards the threshold (similarly as is with regular Governance and the cool down period).

### DIFF
--- a/ARCs/arc-0034.md
+++ b/ARCs/arc-0034.md
@@ -39,11 +39,11 @@ The voting system for the xGov program will be the following:
 In order for a proposal to be approved, it is necessary for the number of votes in favor of the proposal to be proportionate to the amount of funds requested. This ensures that the allocation of funds is in line with the community's consensus and in accordance with democratic principles.
 
 The formula to calculate the voting power needed to pass a proposal is as follows:
-`Voting Power Needed = (Amount Requested) / (Amount Available) * (Total Number Of Algo In Term Pools)`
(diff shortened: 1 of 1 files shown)


SudoWeezy:
We cannot change anything for the current period.
However, we have two options available for the next ones, and it should be discussed with the community: 
- we follow your proposal and move the requested amount base on the number of voters.
- we could take the bet that the next session will be more efficient and will not have a massive gap between participants & voters and not try to solve this with maths (people who register for xGov will be more careful to spend their votes & proposals will be more thoughtfully created)

ShaperOfEntropy:
Theoretically, you could redo the threshold recaluclation off-chain and distribute project funds based on that already for this period. Nothing would change to the voting as it is still recorded by the smart contract. It would just aid the program as a whole and I think community would support such a change even during the period, especially as I do not see any downside to it.

> we could take the bet that the next session will be more efficient and will not have a massive gap between participants & voters and not try to solve this with maths (people who register for xGov will be more careful to spend their votes & proposals will be more thoughtfully created)

If this problem is left unsolved, there will always be the possibility that some funds end up blocked, which would be inefficient, especially since the solution is a very simple one.
The problem with the approach that we hope the situation will improve with the next session is also the fact that the unspent votes in this session will continue blocking funds for the next 3 sessions.

PR #239, opened 2023-08-29
ARC-52 Contextual Wallet API

ehanoc:
Embracing wallets ability to be (or use) a KMS for more functions that are web3 friendly / compliant.

This also addresses problems for some projects in the ecosystem where wallets aren't able to generate, hold and recover keys that related to something else than transactions (i.e Identity / DIDs, Authentication, Secret Sharing, etc)


ehanoc:
> One thing that needs to be considered when it comes to signing is that currently the Algorand SDK does NOT support signing arbitrary bytes unless they are prefixed with `MX`. All the current signing contexts (txns, messages, lsigs, etc.) have their own prefix for domain seperation. It seems like we want more general signing capabilities, in which case the SDK should be expanded to support signing ANY arbitrary bytes. I imagine this function should still check to ensure that the bytes being signed are NOT a program/txn.

I think this is okay. I think we can just defined the API (or Data Models) that any wallet could, if it chooses to, implement.  The SDK can choose to update (or not). But there's not really specific to Algorand about key-generation, bip39/32/44 derivations or signing. In fact i think most Contextual keygen/signing past transactions (&logicsigs maybe) will be done with dedicated keys, not directly associated with addresses / accounts. Which i think it would make cryptographic sense to have dedicated keys for dedicated purposes; but still all of them recoverable from the same seed.

k13n:
I think method `sign` should take some extra parameters that allows an application to communicate to the wallet & user what is being signed. This additional information can be structured or unstructured. If we consider the example of signing logicsig templates (ARC-47), how would the dapp communicate the ARC-47 template and the necessary placeholder replacements to the wallet?

Why is it only possible to export GPG keys but not any other kind of keys? It seems like we should either support exporting all keys or no keys.

Why is the interface written in typescript? Not all wallets are web-wallets written in TS/JS. How is a mobile wallet supposed to implement this ARC? Maybe this is meant as some abstract pseudo-code, but if so, it seems too specific to typescript since promises are returned. This also applies to many other ARCs that are typically written with some JS runtime in mind.

pbennett:
> One thing that needs to be considered when it comes to signing is that currently the Algorand SDK does NOT support signing arbitrary bytes unless they are prefixed with `MX`.

The SDK has a sign bytes function but *IT* prepends MX prior to the received bytes and then hashes it.  The bytes passed into it can be anything.

emg110:
Great ARC Proposal! So excited to see how it goes!

ehanoc:
> I think method `sign` should take some extra parameters that allows an application to communicate to the wallet & user what is being signed. This additional information can be structured or unstructured. If we consider the example of signing logicsig templates (ARC-47), how would the dapp communicate the ARC-47 template and the necessary placeholder replacements to the wallet?
> 
> Why is it only possible to export GPG keys but not any other kind of keys? It seems like we should either support exporting all keys or no keys.
> 
> Why is the interface written in typescript? Not all wallets are web-wallets written in TS/JS. How is a mobile wallet supposed to implement this ARC? Maybe this is meant as some abstract pseudo-code, but if so, it seems too specific to typescript since promises are returned. This also applies to many other ARCs that are typically written with some JS runtime in mind.

Taken this into account @k13n . Thanks for this

jasonpaulos:
@ehanoc do you know how this proposed derivation scheme compares to [SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md)'s support for ed25519? It seems like there are more libraries and tooling around that version of BIP-32 for ed25519 than the one introduced by the paper in this PR.

ehanoc:
> A few questions and comments
> 
> * The ARC title says "Wallet API" and the contents define a Typescript API.  However it seems the spec's main purpose is to define a hierarchical key derivation and signature scheme for wallets to adopt, not a Typescript API. Could it be rewritten to focus on the new cryptographic standards it is proposing? The Typescript API seems like an implementation detail to me and the spec is confusing to follow, since it lacks technical detail and points the reader at the TS reference implementation code rather than any formal description. The kind of specs I'm thinking of are more like
>   
>   * https://docs.oasis.io/adrs/0008-standard-account-key-generation/
>   * https://docs.sui.io/concepts/cryptography/transaction-auth/keys-addresses
>   * https://blog.sui.io/wallet-cryptography-specifications/
>   * https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0005.md
>   * https://cips.cardano.org/cips/cip1852/
>   * https://cips.cardano.org/cips/cip3/
>   * https://zips.z.cash/zip-0032
> * The reference implementation README mentions "note: Bip32-ed25519 is meant to be compatible with our Ledger implementation." So is the Typescript implementation the same as the Ledger SDK call `os_derive_bip32_with_seed_no_throw(HDW_NORMAL, CX_CURVE_Ed25519, ...)` or are there any differences?
> * From an initial read it wasn't clear to me whether you are defining a new scheme unique to Algorand or just making the existing Ledger standard official. (Reading more carefully AFAICT this is just establishing the existing Ledger HDW_NORMAL bip32-ed25519 method as the standard?) Ideally this ARC recommends a key derivation method already implemented by well-used third-party libraries. If so, then IMO a reference implementation might not be needed for this ARC; just test vectors and a English description should be enough, plus links to existing libraries that support this spec.
> * I do see a lot of blockchains using either SLIP-0010 or bip32-ed25519 variations, I don't really have any feedback about usability of hard vs soft derivations, but choosing an existing standard with support from third-party libraries is what I believe is most important.

Good questions! Thanks! 

- The main purpose of the proposal is the API. Although it does have dependency of being able to derive keys. The BIP39 / BIP32 / BIP44 were already flagged in the wallet council as standards that we should adopt. Although for ed25519 there are variants, and my reference implementation here is just to comply with the one that we need / should use. It makes ECDH safe to use and supports soft derivations. We could separate these; but creating an ARC proposal to propose to adopt another standard is a bit weird.  

You can find our compatibility matrix here: [Compatibility Matrix](https://arc.algorand.foundation/wallets) and notice that some wallets (namely Exodus (and Ledger) already support these anyway. )

- The API is defined generically in the specification section, although the examples are indeed in Typescript. I'm happy to adopt a particular format to do the specifications if there's hunger for it. 

- The Ledger implementation does not fully comply with that paper; they use SLIP-0010 and the root key creation, which is technically fine but SLIP-0010 by itself only supports hard derivations. There is also extra bit manipulation after the derivations that doesn't comply with the paper. Specially the Child Keys don't receive the `kl + 8 * truncate28BytesOf(zl)` treatment, but get the regular clamping. I don't think it's insecure, but as part of this API proposal we want to support soft derivations and ECDH; which will allows to do key agreements between 2 algorand addresses / keys and have e2e encryption. The paper + clamping makes the use of these keys and mappings from ed25519 -> X25519 secure.

Ledger also does not take advantage fully of [BIP44] (https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki). It only derives at the `account` level with hard derivations and not at the `address_index` level; which would be a soft derivation. Meaning we can simply derive public keys with the previous node public key without having to access private key material.  

- I'm happy to take an existing compliant bip32-ed25519 lib / package instead of having to apply it again; But i think we need test vectors from a reference implementation that we can verify that they all work the same. The goal should be interoperability between wallets; so that we can import / export from one to the other and they all rebuild the same wallet state. 

Hope it makes sense

ehanoc:
> > The main purpose of the proposal is the API. Although it does have dependency of being able to derive keys. The BIP39 / BIP32 / BIP44 were already flagged in the wallet council as standards that we should adopt. Although for ed25519 there are variants, and my reference implementation here is just to comply with the one that we need / should use. It makes ECDH safe to use and supports soft derivations. We could separate these; but creating an ARC proposal to propose to adopt another standard is a bit weird.
> 
> Thanks for the additional explanation — if the focus here is the Typescript API, IMO the complexity of the different variations of bip32-ed25519 deserves a separate ARC, or at least separate section in this ARC, formally defining the key derivation scheme and including test vectors, like the examples from other chains I cited. It doesn't have to be super long but especially since you are defining a new standard for deriving keys and signing non-transaction data that (I assume?) will end up in SDKs for different languages — plus given the heightened security and correctness requirements of such a standard, this feels like the kind of thing that should be somewhat more formally written out, so implementors have a clearer spec than how it's described here. Either way I don't think it's weird to have an ARC for key derivation that is cited by your wallet API ARC.
> 
> > The Ledger implementation does not fully comply with that paper; they use SLIP-0010 and the root key creation, which is technically fine but SLIP-0010 by itself only supports hard derivations. ... as part of this API proposal we want to support soft derivations and ECDH; which will allows to do key agreements between 2 algorand addresses / keys ...
> 
> Ohhh I see, for linking child accounts to parent accounts and stuff like that. I missed this in the ledger app code, setting the 0x8... bit to get hardened child keys for use with HDW_NORMAL: https://github.com/algorand/ledger-app-algorand/blob/f55c4e6ead8aafe3b8789ab905991f7286ffcb47/app/src/coin.h#L28-L31 https://github.com/algorand/ledger-app-algorand/blob/f55c4e6ead8aafe3b8789ab905991f7286ffcb47/app/src/crypto.c#L36-L38
> 
> I am a little confused though, where in the ledger SDK or the ledger-app-algorand source does it say the root key creation or HDW_NORMAL uses SLIP-0010?

- Test Vectors; good point; I've added them to the reference implementation. 

- Ledger and legacy SLIP-0010; yeah, so their main code is closed source. Although i've managed to look through their sample codes and reverse engineer some bits, i have a Ledger compliant implementation of BIP32. And it's very non standard. You can check one Ledger's reference impl here: https://github.com/LedgerHQ/orakolo/blob/0b2d5e669ec61df9a824df9fa1a363060116b490/src/python/orakolo/HDEd25519.py#L10-L41

- I'm not trying to define a new BIP32 standard just adopt what's already spec'ed and define. It is true that some ecosystems and libs do not adhere to it by the book; some for technical reasons, others due to legacy implementations and backwards compatibility and others i don't understand why. I don't think it needs to be "formally" written since the BIP32-ed25519 paper does that already. And since we have a "fresh start" and it seems the wallet council members are okay with us let Ledger-be-Ledger and the rest adopt the standard as it was defined. 

- This ARC isn't trying to be TypeScript prescriptive, outside the examples the definition is trying to be agnostic. I'm happy if someone adds other language examples. If i can i might add others too.

assets/arc-0052/bip32-ed25519.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  // clamp
+  //Set the bits in kL as follows:
+  // little Endianess
+  kL[0] &= 0b11_11_10_00; // the lowest 3 bits of the first byte of kL are cleared
+  kL[31] &= 0b01_11_11_11; // the highest bit of the last byte is cleared
+  kL[31] |= 0b01_00_00_00; // the second highest bit of the last byte is set
+
+  // chain root code
+  const c: Buffer = createHash("sha256").update(Buffer.concat([new Uint8Array(0x01), seed])).digest();
```
k13n, line 40: This line seems to refer to formula $H_{256}(0x01||k)$ in the referenced paper, which is the concatenation of byte strings `0x01` and $k$.

The line `new Uint8Array(0x01)` does not generate a byte string `[0x01]`, but rather an array of zeros with length `0x01`, i.e., `[0x00]`. 

Therefore, should the above line be changed to `new Uint8Array([0x01])`?

ehanoc: Great catch. That's definitely an oversight. Thank you!

emg110: @k13n, Kevin, what an eye!!!!

HashMapsData2Value:
Do we need to consider further the fact that coin type 0' is registered for Bitcoin? Is this a standard for generating identity keys, or is it something we will be introducing as part of this ARC?

Granted Bitcoin is not Ed25519 compatible and as such there shouldn't be a way to link someone's Ed25519 ARC-52 derived identity key with their Secp256k1 Bitcoin address.

I did a (cursory) check through the top blockchains listed in [SLIP-44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) and compared it with [this list of elliptic curves](http://ethanfast.com/top-crypto.html). Most of the cryptocurrencies in the list are quite obscure and might not even exist anymore, so it's difficult to find the curves quickly. But at the very least, NEM (XEM) chain, which has snagged coin type 44, also uses Ed25519. Monero follows with coin type 128, XRP 144, Stellar 148, DOT 354 and Cardano 1815.

emg110:
@HashMapsData2Value , I just wanted to humbly mention (as a side note regarding linking Ed25519 ARC-52 derived identity key with Secp256k1 Bitcoin address) that there is an actual web standard way of linking different curve-based or derived identities together and that is by using DID context and bringing all different identities (using different curves ) as verification methods under that and by resolving that master identity all those under are actually linked together as well and can be used for different occasions (different chain signings for example) ! As another approach where there is a need for separate identities just being linked and not sharing a composed identity, there is as well  "alsoKnownAs" field in DID document which contains an array of other identities (based on different curves or chains or,...) that can be linked!
 See [ARC13](https://github.com/algorandfoundation/ARCs/assets/1900448/c6d6d19e-6aed-495a-91d4-fadd60c66905) for more details! 
Since mosaic (composable) identities based on Algorand identities and also ARC52-based derived identities are fundamental to DIDs and therefore ARC13, users can have unlimited number of different linked identity scenarios for different purposes!

ehanoc:
> Do we need to consider further the fact that coin type 0', which we want to use for identity keys, is registered for Bitcoin? Is this an established standard for generating identity keys, or is it something we will be introducing as part of this ARC?
> 
> Granted Bitcoin is not Ed25519 compatible and as such there shouldn't be a way to link someone's Ed25519 ARC-52 derived identity key with their Secp256k1 Bitcoin address.
> 
> I did a (cursory) check through the top blockchains listed in [SLIP-44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) and compared it with [this list of elliptic curves](http://ethanfast.com/top-crypto.html). Most of the cryptocurrencies in the list are quite obscure and might not even exist anymore, so it's difficult to find the curves quickly. But at the very least, NEM (XEM) chain, which has snagged coin type 44, also uses Ed25519. Monero follows with coin type 128, XRP 144, Stellar 148, DOT 354 and Cardano 1815.

The `coin_type` being expected to be `0` is a result of trying to comply with the current interoperability spec of W3C for digital wallets; [W3C Interop Universal Wallet Spec](https://w3c-ccg.github.io/universal-wallet-interop-spec/#hd-wallets).

Being two separate curves, do you still see a problem?

ehanoc:
> I think the ECDH example, while having correct syntax, might be a little confusing since it assumes that Alice and Bob have the same underlying root, with Alice key index 0 and Bob key index 1. In reality Alice would receive Bob's pk, so I think the example should reflect that.

Good point. I updated the tests to reflect that: https://github.com/algorandfoundation/ARCs/pull/239/files#diff-707d012926dd86071e1759dbce441d11c620617a409190612ea813493aadf369R351

assets/arc-0052/bip32-ed25519.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  const chainCode = childChainCode.subarray(32, 64);
+  const zl = z.subarray(0, 32);
+  const zr = z.subarray(32, 64);
+
+  // left = kl + 8 * trunc28(zl)
+  // right = zr + kr
+
+  const left = new BN(kl, 16, "le").add(new BN(zl.subarray(0, 28), 16, "le").mul(new BN(8))).toArrayLike(Buffer, "le", 32);
+  let right = new BN(kr, 16, "le").add(new BN(zr, 16, "le")).toArrayLike(Buffer, "le").slice(0, 32);
```
HashMapsData2Value, line 76: K_R is supposed to be taken mod 2^256. I assume the .slice(0, 32) accomplishes this?

assets/arc-0052/bip32-ed25519.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  // just padding
+  if (right.length !== 32) {
```
HashMapsData2Value, line 79: If the length isn't 32, we are sure that exactly "00" will be enough to bring it to 32? Or could it be e.g. 31 length?

assets/arc-0052/bip32-ed25519.ts
```diff
@@ -0,0 +1,181 @@
...
+  const { z, childChainCode } = index < 0x80000000 ? derivedNonHardened(kl, cc, index) : deriveHardened(kl, kr, cc, index);
+
+  const chainCode = childChainCode.subarray(32, 64);
+  const zl = z.subarray(0, 32);
+  const zr = z.subarray(32, 64);
+
+  // left = kl + 8 * trunc28(zl)
+  // right = zr + kr
+
+  const left = new BN(kl, 16, "le").add(new BN(zl.subarray(0, 28), 16, "le").mul(new BN(8))).toArrayLike(Buffer, "le", 32);
```
HashMapsData2Value, line 75: According to the paper, `if kL is divisible by the base order n, discard the child`. Where do we check if this is divisible?

assets/arc-0052/bip32-ed25519.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    // Section V. BIP32-Ed25519: Specification; subsection D) Public Child Key Derivation
+    const chainCode: Buffer = i.subarray(32, 64);
+    const zl: Buffer = z.subarray(0, 32);
+
+    // left = 8 * 28bytesOf(zl)
+    const left = new BN(zl.subarray(0, 28), 16, 'le').mul(new BN(8)).toArrayLike(Buffer, 'le', 32);
+
+    const p: Uint8Array = crypto_scalarmult_ed25519_base_noclamp(left);
+    return Buffer.concat([crypto_core_ed25519_add(p, pk), chainCode]);
```
HashMapsData2Value, line 122: In the specification it says `If Ai is the identity point (0,1), discard the child`. Where do we check for this?

ehanoc:
> The advantage of this is that it would be much easier to do address discoverability. As soon as you point your wallet (Pera, Defly, etc) to an application, and that application uses ed25519 keys for whatever reason (privacy, identity, etc), it will be much easier for the wallet to go to the application's designated coin type and start looping through the account indices and then key indices to find previous activity and make it available to the user.

That should happen anyway with BIP39/32/44 compliant wallets. You scan the BIP44 until accounts + addresses are empty. 

> In this case however, if let's say coin type 148 is allocated to [insert Algorand Application] AND Stellar Lumens, and an individual also owns XLM, that user might inadvertently link their history with the two together.

This would be quite a deviation from [BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki), if the point is to be compliant, perhaps avoiding using `coin_type` to specific applications would be best. We could in theory use `change` instead; which isn't commonly used at all. Was initially considered to differentiate between outgoing / incoming addresses.   

Overall, i think we can brainstorm this. But if you have in mind specific designs to tackle ring signatures, maybe we can do this in another ARC? Unless you think this would prevent you from doing those

assets/arc-0052/bip32-ed25519.ts
ehanoc: Right; 

How about we replace it with: 

```ts
  const rightBuffer = Buffer.alloc(32);
  Buffer.from(right).copy(rightBuffer, 0, 0, right.length)
 ```
 
 Should guarantee 32 bytes. Any remaining bytes will be zero

File: assets/arc-0052/bip32-ed25519.ts (hunk 4 above)
Comment (ehanoc) [reply assets/arc-0052/bip32-ed25519.ts:75]: This is where i went and checked all other known implementations (i.e Nano, Cardano, bip32-ed25519) and none of them do anything obvious related to this. The paper doesn't state what you should do if you discard the child. 

Derivations are deterministic, if you were to reject a child, you wouldn't be able to derive further keys on that path. 

I've messaged the authors trying to get a clarification. Do you have a suggestion?

**Comment by HashMapsData2Value** [comment]
> > The advantage of this is that it would be much easier to do address discoverability. As soon as you point your wallet (Pera, Defly, etc) to an application, and that application uses ed25519 keys for whatever reason (privacy, identity, etc), it will be much easier for the wallet to go to the application's designated coin type and start looping through the account indices and then key indices to find previous activity and make it available to the user.
> 
> That should happen anyway with BIP39/32/44 compliant wallets. You scan the BIP44 until accounts + addresses are empty.

The wallet would scan for addresses on the Algorand blockchain and see if they have any transaction history. It could do this blindly. However, if the public key is used as part of an application the wallet would need to know to scan inside the box elements of specific smart contracts, perhaps even going back into old Algorand history.

> > In this case however, if let's say coin type 148 is allocated to [insert Algorand Application] AND Stellar Lumens, and an individual also owns XLM, that user might inadvertently link their history with the two together.
> 
> This would be quite a deviation from [BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki), if the point is to be compliant, perhaps avoiding using `coin_type` to specific applications would be best. We could in theory use `change` instead; which isn't commonly used at all. Was initially considered to differentiate between outgoing / incoming addresses.
> 
> Overall, i think we can brainstorm this. But if you have in mind specific designs to tackle ring signatures, maybe we can do this in another ARC? Unless you think this would prevent you from doing those

Sure let's brainstorm it.

Ultimately the worst case scenario that would happen is that, just like how the Web3 standard expects people to simply use two sets of mnemonics mapping to coin type 0 (one being the mnemonics used for Bitcoin, the other being used exclusively for identity), something application-specific could be done here as well. But it would be nice to be able to avoid it.

File: assets/arc-0052/bip32-ed25519.ts (hunk 4 above)
Comment (HashMapsData2Value) [reply assets/arc-0052/bip32-ed25519.ts:75]: I assume inside the wallet itself you'd simply have to flag that index as bad and move to the next index?

File: assets/arc-0052/contextual.api.crypto.ts [hunk 6]
```diff
@@ -0,0 +1,292 @@
...
+import { deriveChildNodePrivate, fromSeed } from './bip32-ed25519';
+
+
+/**
+ * 
+ */
+export enum KeyContext {
+    Address = 0,
+    Identity = 1,
+    Cardano = 2,
```
HashMapsData2Value, line 23: Do we really need Cardano as a key context

ehanoc: Ah we don't. This was just to try to use some of their test vectors to match implementations. I can remove it

ARCs/arc-0052.md
```diff
@@ -0,0 +1,210 @@
...
+
+The current wallet APIs available only support key derivation for addresses and signing for transactions. In the broader context of Algorand and web3 ecosystems, there is a need for a more flexible API that allows for contextual key generation and signing. An example of this could be identity dedicated keys (i.e DIDs), or passkeys for authentication.
+
+## Reference Implementation
+
+Reference Implementation is included in the `assets` folder <a href= "https://github.com/algorandfoundation/ARCs/tree/main/assets/arc-0052">here</a>.
+
+## Security Considerations
+
+As this functionality is based on EdDSA, BIP32 HD derivations and ECDH, the security considerations of those specifications apply. Particularly, the Ed25519 derived keys must be clamped to avoid known attacks
```
HashMapsData2Value, line 206: Since we are including an ECDH function to produce a shared secret we should point out that the shared secret should NOT be used directly for encryption. Instead the output of ECDH should be concatenated with the bytes of the public keys and then hashed to produce a new secret.

> secret = Hash(concatenation(ECDH(skA, pkB), pkA, pkB))

(Where skA and pkA are of course produced following the BIP39/ARC-52 standard.)

Note that Alice and Bob need to agree beforehand on whose public key is concatenated first and whose second.

An explanation for why is given here: https://github.com/jedisct1/libsodium/issues/586#issuecomment-325182090

Scalarmult produces a point, a group element. While we use 32 bytes, corresponding to 2^256 possible values, in truth the order of the group is roughly ~2^252 and the output is not "uniformly random".

Hashing (e.g. with SHA512/256) makes the output uniformly random, but while the output would occupy the full 32 bytes space the range of inputs to the hash function is still constrained by ~2^252 possible inputs and the same number of outputs since hashes map (input) data to (output) value. By concatenating with something extra  increase the range of outputs.

Keep in mind that the different combinations of n,p in scalarmult produce the same new point as well. Since Alice's pk is x*G and Bob's y*G, their shared secret is xy*G. x=10, y = 50 would produce the same shared secret as x=20 and y = 25. However, Hash(Concat(150*G, 10*G, 50*G)) !== Hash(Concat(150*G, 20*G, 25*G)). So the space of potential secrets is boosted.

**Alternatively** we should update the ECDH so it doesn't return scalarmult(skA, pkB) directly. Instead we update it with an additional parameter "input order" and then returns either Hash(Concat(scalarmult(skA, otherPubKey), pkA, otherPubKey) or Hash(Concat(scalarmult(skA, otherPubKey), otherPubKey, pkA) depending on the order.

ehanoc: > Since we are including an ECDH function to produce a shared secret we should point out that the shared secret as produced by scalarmult(., .) should NOT be used directly for encryption. Instead the output of ECDH should be concatenated with the bytes of the public keys and then hashed to produce a new secret.
> 
> > secret = Hash(concatenation(ECDH(skA, pkB), pkA, pkB))
> 
> (Where skA and pkA are of course produced following the BIP39/ARC-52 standard.)
> 
> Note that Alice and Bob need to agree beforehand on whose public key is concatenated first and whose second.
> 
> An explanation for why is given here: [jedisct1/libsodium#586 (comment)](https://github.com/jedisct1/libsodium/issues/586#issuecomment-325182090)
> 
> Scalarmult produces a point, a group element. While we use 32 bytes, corresponding to 2^256 possible values, in truth the order of the group is roughly ~2^252 and the output is not "uniformly random".
> 
> Hashing (e.g. with SHA512/256) makes the output uniformly random, but while the output would occupy the full 32 bytes space the range of inputs to the hash function is still constrained by ~2^252 possible inputs and the same number of outputs since hashes map (input) data to (output) value. By concatenating with something extra increase the range of outputs.
> 
> Keep in mind that the different combinations of n,p in scalarmult produce the same new point as well. Since Alice's pk is aG and Bob's bG, their shared secret is abG. a = 10, b = 50 would produce the same shared secret as a=20 and b = 25. However, Hash(Concat(150G, 10G, 50G)) !== Hash(Concat(150G, 20G, 25G)).
> 
> **Alternatively** we should update the ECDH so it doesn't return scalarmult(skA, pkB) directly. Instead we update it with an additional parameter "input order" and then returns either Hash(Concat(scalarmult(skA, otherPubKey), pkA, otherPubKey)) or Hash(Concat(scalarmult(skA, otherPubKey), otherPubKey, pkA)) depending on the order.

Great find. Update is needed then

HashMapsData2Value: Also mentioned here: https://github.com/jedisct1/libsodium/blob/master/src/libsodium/include/sodium/crypto_scalarmult.h#L33

In crypto_kx() https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_kx/crypto_kx.c#L17 they're calling on "crypto_generichash".

Navigating to it shows that they're using blake2b: https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_generichash/crypto_generichash.c#L71

Regarding the input order, they're specifically concatenating the client pk first over the server pk. It might be clear to the two entities engaging in which is which, and if not they can negotiate (perhaps by comparing the bits of the PK and sorting the PK before the upper PK).

assets/arc-0052/contextual.api.crypto.ts
```diff
@@ -0,0 +1,292 @@
...
+        }
+        
+        await ready // libsodium
+
+        const rootKey: Uint8Array = fromSeed(this.seed)
+        const bip44Path: number[] = GetBIP44PathFromContext(context, account, keyIndex)
+        const raw: Uint8Array = await this.deriveKey(rootKey, bip44Path, true)
+
+        const scalar = raw.slice(0, 32);
+        const c = raw.slice(32, 64);
```
HashMapsData2Value, line 150: I believe there is a bug here. c is supposed to be the raw.slice(32, 64), but deriveKey will always return raw as a 32 byte length Uin8Array. You can see this inside deriveKey function on line 97 that we take scalar = derived.subarray(0, 32).


When raw.slice(32, 64) gets run here on a size 32 array, rather than fail it will simply return a 0 length Uint8Array. Later on when concatenating c and data to produce the first hash for r (currently line 156), c is empty and as a result only data gets hashed.

I'm a little unsure on how to proceed. According to the [RFC 8032 section 3.3](https://datatracker.ietf.org/doc/html/rfc8032#page-8), also implemented [noble-curves](https://github.com/paulmillr/noble-curves/blob/6a85252dc3f9d8bef7e407177c6c81a3a67b185b/src/abstract/edwards.ts#L449) (c is called "prefix" and is calculated on line 426) you would grab your private key, hash it to get 64 bytes. The first 32 bytes would make up the scalar to calculate the scalar*B = publicKey (which we do). And then the last 32 bytes would be used for this.

Should we change deriveKey so instead of isPrivate (true/false), it could have three modes: get scalar, get publicykey and get raw (full 64 bytes)?

HashMapsData2Value: Btw, for those confused, the code still works, it's just that R (the first 32 bytes of the signature) will not be produced correctly. The problem is that r will be calculable as a function of the message data, a known value. If you know r you can take use the other half the signature, S, and calculate the private key as such: (S - r)*h^-1 = k

assets/arc-0052/contextual.api.crypto.ts
```diff
@@ -0,0 +1,292 @@
...
+
+        // \(1): pubKey = scalar * G (base point, no clamp)
+        const publicKey = crypto_scalarmult_ed25519_base_noclamp(scalar);
+
+        // \(2): h = hash(c + msg) mod q
+        const hash: bigint = Buffer.from(crypto_hash_sha512(Buffer.concat([c, data]))).readBigInt64LE()
```
HashMapsData2Value, line 156: I'm not sure about this line here. 

I called signData in the following way:

const message = Buffer.from("Hello World")
const signature: Uint8Array = await cryptoService.signData(KeyContext.Address, 0, 0, message, { encoding: Encoding.NONE, schema: {} })

When I print out the hash it looks like this:

Screenshot 2024-02-21 at 10 29 34

And the calculated r looks like this:
Screenshot 2024-02-21 at 10 30 12

To me I think the hash looks strange, I would expect it to have the full length, and then we simply do a r = scalar_reduce(hash) to get it down under q with some reversal of bytes to get it into Little Endian mode.

ARCs/arc-0052.md
```diff
@@ -0,0 +1,210 @@
...
+}
+```
+
+#### BIP44 Context Derivation Paths
+
+There is several reasons why we want to avoid key re-use in different contexts. One of the reasons is that we want to avoid the possibility of **leaking** information about the identity of the user. Another reason is that there is some cryptographic security concers when re-using keys or in case of keys being compromised the impact is **compartmentalized**.
+
+For this reason is advisable to have different derivation paths for different contexts supporting recoveribility of the keys by using the same mnemonic as before.
+
+| Purpose | Coin Type | Account | Change | Address Index |    Context    |
```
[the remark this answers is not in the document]
ehanoc: > A few points here:
> 
> 1. Is there any particular reason why we opted for `purpose` 44 equal to BIP-44? IMO we should use a different `purpose` from Bitcoin's wallets.
> 2. Coin Type: we should provide default registered coin types (like [here](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#registered-coin-types)), at least for ALGO, and common ASAs e.g. USDC, gALGO, goETH, goBTC, ...
> 3. Are we enforcing the same account discovery algorithm ad BIP-44 here? Does it make sense to put a limit on account indexes that are valid only if previously active on-chain? What if I want to reserve 2/3 consecutive derivation paths for accounts that I will never use on-chain (maybe one for document encryption, another for authentication, erc..)

Hey Stefano,

So, regarding

1) We have discussed this before. It doesn't really matter that is the same path as in bitcoin, it's 2 different elliptic curves. And this is following the spec for the W3C Universal Wallet API(https://github.com/w3c-ccg/universal-wallet-interop-spec) 

2) Yea, we could extend it in the future to other ecosystems / applications

3) The discovery algorithm should be the same as any other chain wallet. Recursive finding until no account / address has any activity. For off-chain use, if you need to recovery, all is still recoverable, depending on the context you need to loop through them and see if you can use them, or can include them in wallet backups.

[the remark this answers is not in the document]
deanstef: 1. Yes we use different curves. However, this ARC aims at providing "Contextual Wallet APIs" for different purposes and contexts. In that case, we might want to use another `purpose` as a derivation path. For example, [SLIP-0013](https://github.com/satoshilabs/slips/blob/ef6d7700cc/slip-0013.md) provides authentication for HD wallets and uses `purpose = 13'` for authentication with deterministic hierarchy.
2. Yes! Maybe providing a table with "reserved" values could be useful before having fragmented implementations in the ecosystem. WDYT?
3. It depends on the purpose and context. While I think the discovery algorithm is needed for multi-account HD wallets (BIP-44-like) it is not necessary in other cases like authentication.

ARCs/arc-0052.md
```diff
@@ -0,0 +1,210 @@
...
+#### Wallet New Interface
+
+This new interface is for the wallet to implement. It is an extension of the current wallet interface.
+
+This extension is **OPTIONAL** and separate from the current interface that handles address generation and transaction signing.
+
+```ts
+interface Wallet {
+    ...
+    keyGen(context: KeyContext, account:number, keyIndex: number): Promise<Uint8Array>;
```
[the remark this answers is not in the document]
deanstef: I am ok with having those well specified in the implementation. However, I would like to be mandatory in the spec as well.

IMHO those constraints should be clarified in the ARC-52 spec, or at least should be redirected to BIP-32. My issue with the current text is that an Algorand wallet might be compliant with ARC-52 but not compliant with BIP-32.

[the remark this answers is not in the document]
ehanoc: BIP-44 defines the hardening at account / index level. 

I don't think we should expose that here as this is a wallet API. Anyone that would like to do BIP32 but not BIP44 compliant, should just use a low level BIP32 API to derive keys that way (i.e non hardened accounts or hardened indexes instead)

ARCs/arc-0052.md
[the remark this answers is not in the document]
HashMapsData2Value: I don't understand the point about assigning the coin type to different ASAs. Given that Algo is needed to opt in and transact with an ASA regardless, how would assigning ASA IDs to coin type work? And what if you want/need to hold multiple ASAs in one address?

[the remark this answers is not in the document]
deanstef: What I like of different coin types for ASAs is that it force users to split their holdings into different addresses. This would improve security in case of compromised keys.

Maybe changing the coin type is not the best approach because of ALGO though.

As you mentioned, ALGO opt-in is a protocol requirement, so any address should be intended to hold ALGOs as well as another ASA. However, we could think of another derivation level dedicated to ASAs. Some related discussion [here](https://github.com/ethereum/EIPs/issues/85).

ARCs/arc-0052.md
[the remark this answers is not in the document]
ehanoc: I understand, although i see problems with that. 

We could expose more options or a more "advanced" API. problem is not every integrator understands the nuances of technical standards that are used underneath. Good API's in my opinion are strict and don't offer too many options that can be miss used. 

There can be an integrator that understands BIP-32 and another that doesn't. This would lead for implementations that might not be interoperable and recovery in one wallet wouldn't work on another wallet.

ARCs/arc-0052.md
ehanoc: Updated

assets/arc-0052/contextual.api.crypto.ts
ehanoc: Updated

assets/arc-0052/contextual.api.crypto.ts
ehanoc: updated

ARCs/arc-0052.md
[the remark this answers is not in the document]
HashMapsData2Value: I feel that the biggest risk here is that the seed gets compromised, in which case all addresses are compromised regardless of the derivation path. I think it is unlikely that a single derived private key (i.e. a scalar output of the deriveKey function) could be compromised without the seed or many addresses also being compromised in the process.

If we want users want to split their funds wallets can encourage them and make it easy to use different indices or different accounts for accepting different tokens.

assets/arc-0052/contextual.api.crypto.ts
```diff
@@ -0,0 +1,344 @@
...
+     * @param context
+     * - context of the key (i.e Address, Identity)
+     * @param account
+     * - account number. This value will be hardened as part of BIP44
+     * @param keyIndex
+     * - key index. This value will be a SOFT derivation as part of BIP44.
+     * @param prefixEncodedTx
+     * - Encoded transaction object
+     *
+     * @returns stx
```
ori-shem-tov, line 203:
```suggestion
     * @returns sig
```

assets/arc-0052/package.json
```diff
@@ -29,7 +29,8 @@
     "js-sha512": "^0.8.0",
     "libsodium-wrappers-sumo": "^0.7.13",
     "ts-custom-error": "^3.2.0",
-    "ts-log": "^2.2.4"
+    "ts-log": "^2.2.4",
+    "tweetnacl": "1.0.3"
```
HashMapsData2Value, line 33: Duplicate dependencies with tweetnacl defined in devDependencies

Missjones2829 (commented):
Merge

cusma:

Roseymr (approved):

PR #259, opened 2023-11-15
Update arc-0073.md

temptemp3:
Add arc namespacing


temptemp3:
Thinking that arc73 interface selector may be added as well

SudoWeezy:
This ARC has been finalized and won't be subject to further changes. Given this, I believe updating the method in question is unnecessary, as its purpose is to indicate the currently supported Interface/ARC.

PR #260, opened 2023-11-20
ARC-54 - Simpler reference contract + TealScript source

SilentRhetoric:
Based on feedback from other developers, two changes are being proposed to the reference implementation:

1. The contract will accept any ASAs even if they have clawback enabled.  This simplifies the process of existing smart contracts being able to interact with this contract.  It also means that there can be just this one burning account for any ASA, with the caveat that clawback-enabled ASAs can never be considered burned.

2. The contract does not enforce that the caller must pay the contract to cover the MBR increase at the time of requesting an opt in.  This enables people to donate Algos to the account to cover the MBR for others to use for burning without incurring the MBR cost.

The net effect is also that the contract is simpler and easier to audit.

I have also included the TealScript source code used to generate the TEAL.


ctibo:
I agree on the fact that the contract should accept anything (clawback or not). Displaying an asset as burned is really just a display thing. Explorers should check if the asset has clawback and add the appropriate labels.

Preventing clawback-enabled assets would mean that the only way to get rid of such an asset is by sending it back to the creator. While this approach works, it's not what I'd prefer to do in the case of scam to rug projects. I'd prefer to send them to a burn contract, and have them even temporarily marked as burned.

This new version is really simple to read. I love it.

ARCs/arc-0054.md
```diff
@@ -14,35 +14,39 @@ created: 2023-09-15
 This ARC provides TEAL which would deploy a application that can be used for burning Algorand Standard Assets. The goal is to have the apps deployed on the public networks using this TEAL to provide a standardized burn address and app ID.
 
 ## Motivation
-Currently there is no official way to burn ASAs. While one can currently deploy their own app or rekey an account holding the asset to a vanity address, having a standardized address for burned assets allows explorers and dapps to display burnt supply for every ASA.
+Currently there is no official way to burn ASAs. While one can currently deploy their own app or rekey an account holding the asset to the zero address, having a standardized address for burned assets enables explorers and dapps to easily calculate and display burnt supply for any ASA burned here.
```
d13co, line 17: > rekey an account ... to the zero address

The zero address is the only address that can not be rekeyed-to. It is interpreted as "no rekey"

ARCs/arc-0054.md
```diff
@@ -83,142 +93,120 @@ switch create_NoOp NOT_IMPLEMENTED NOT_IMPLEMENTED NOT_IMPLEMENTED NOT_IMPLEMENT
...
+class ARC54 extends Contract {
+  /*
+   * Sends an inner transaction to opt the contract account into an ASA.
+   * Will fail if the contract is already opted in to the asset.
+   * The fee for the inner transaction must be covered by the caller.
+   *
+   * @param asa The ASA to opt in to
+   */
+  optIntoASA(asa: Asset): void {
+    assert(!globals.currentApplicationAddress.hasAsset(asa));
```
pbennett, line 194: I'd remove this assert.  Its simply friction for callers - with this they have to know/check in advance whether the burn contract has already 'burned' the specified asset.  It should always work, opted-in or not.  If it's already opted-in, it's a no-op, return success - whether the caller paid MBR is irrelevant.  Once they send to the burn contract account, it still succeeds which is all they care about.

ARCs/arc-0054.md
SilentRhetoric: Doh!  You are right... which is why there really is no good, trustless way to use a rekeyed account as a burn location.  Fixed the commentary here.

ARCs/arc-0054.md
SilentRhetoric: Good suggestion--I have made this change.

SudoWeezy:
I am not sure if this fits in this PR or not, but we should also add the `arc54_ prefix` to each method

ARCs/arc-0054.md
d13co: Good spot @pbennett . The optin state may change between the client reading the optins and the txn hitting the chain, which could cause contract calls to fail unnecessarily. Given the 0 fee enforcement, this is safe to do

joe-p:
LGTM. I would just add a section under the specification about displaying the values. Primarily noting that interfaces MUST check the clawback address before displaying burn amount.

pbennett:
> LGTM. I would just add a section under the specification about displaying the values. Primarily noting that interfaces MUST check the clawback address before displaying burn amount.

How do you mean?  Clawback address has nothing to do with the burn amount though.  Circulating amounts just need to subtract any held in this burner contract account (take total amount - whats in creator/reserve address - whats in burner contract).  Circulating supply is completely liquid, including clawback assets.  The burn account just becomes another reserve account of sorts - needs subtracted from circulating.  Explorers should show definitely that an asset has clawback  but it still doesn't change what's circulating.

USDT has clawback for eg and they can claw it back anytime.  It doesn't change it having a legitimate (current) circulating value though.

pbennett:
> To me, this standard is about showing tokens that are irreversibly burnt.

There's no such thing on Algorand.  Anything with clawback set can be clawed back at will and there's nothing a contract can do about it.  Are you saying USDT isn't a real asset?  It's a display issue for the circulating supply - that's it.  
The contract refusing to accept 'some' assets causes other issues as some pointed out.  Users can 'burn' tokens and it always works.  If it happens to be clawed back, then that's all visible on-chain - user A sent asset X to the burn contract.  If asset X's owner then clawed it back - that is all visible on-chain.  A 'burn' contract refusing to accept it means the user can only send it back to the creator - possibly implying they're connected accounts (particularly notable for scam tokens) when they're not.

SilentRhetoric:
> > The burn account just becomes another reserve account of sorts
> 
> If a ASA with clawback wants to have a circulating supply shown on explorers/wallets then they should use the reserve address. To me, this standard is about showing tokens that are irreversibly burnt.

I would slightly reframe that as enabling clarity of burned supply for non-CB assets _and also_ enabling convenience of burning any ASA.

joe-p:
> There's no such thing on Algorand. Anything with clawback set can be clawed back at will and there's nothing a contract can do about it.

Exactly, which is why the standard should enforce that clients check the clawback address before displaying the burnt amount, because ASAs with clawback cannot be burnt. 

> Are you saying USDT isn't a real asset?

This has nothing to do with "real assets", but rather what the standard is trying to achieve: standardized asset burning. Asset burning is not possible for assets with clawback, but that doesn't make them any more or less "real". 

What value does this standard offer to users with clawback-enabled ASAs that isn't available through the reserve address?

pbennett:
If you add unnecessary friction to the contract, it won't be used.  User or contract wants to cleanly 'discard' token - they can just sent to that contract and - done.  It's automatically excluded from circulating as being in a known 'burn/discard [if clawback] pile' and anyone looking at on-chain movement can see intent of sender.

joe-p:
Maybe we just need to be clear about burnt supply vs non-circulating supply. 

Burnt supply: Amount of tokens held by ARC54 contract address. Not applicable to clawback. 

Circulating supply: (total amount of tokens) - (amount in reserve address + amount in ARC54 contract)

SilentRhetoric:
> What value does this standard offer to users with clawback-enabled ASAs that isn't available through the reserve address?

A unified experience for users and contracts for "trashing" tokens, whether they can be considered "burned" or not.

joe-p:
@SilentRhetoric would you be open to adding the definitions for non-circulating supply and burnt supply to the ARC?

d13co:
Let me start by saying that I am all for the current contract implementation - it has a good chance of becoming The One Black Hole for tokens (even if some can be yanked back by definition.) **Accepting everything means that I can just count on this contract to burn tokens from another contract** - which is an actual use case I am working on right now. If we had kept the "reject clawback" and even the "fail if opted in" I would think thrice (and twice, respectively) because that would mean more effort on my part to branch my code based on token clawback attribute, or  a chance that the opt-in state change of the contract would make my call fail.

Regarding the circulating supply discussion, I just want to point out that circulating supply if a very soft metric, as it is extremely rarely the case that the Reserve address is locked (e.g. a Vestige Vault.) As such, a transaction from the reserve address would change the circulating supply immediately. Counting the bonfire address as "outside the circulating supply" would not change this metric/mechanic much: a clawback from the bonfire address would also change the circulating supply immediately.

So the circulating supply is a currently a point-in-time "snapshot" metric of the % funds not-in-reserve, and it is reasonably to include not-in-burn-contract in the same metric. When there are funds in the reserve address, and when there is clawback, the circulating supply can be changed at the whim of the reserve and clawback address holders, but this doesn't mean that we should not include the bonfire assets as non-circulating _while they are still there_. Right?

d13co:
I have to ask though: now that we have simplified this to the point of just opting in, would an logic sig suit better than a contract? It would have the benefit of being the same on all 3 networks, so contracts on testnet and mainnet would share the same code for burning without needing to switch out the Bonfire contract address

SilentRhetoric:
> I have to ask though: now that we have simplified this to the point of just opting in, would an logic sig suit better than a contract? It would have the benefit of being the same on all 3 networks, so contracts on testnet and mainnet would share the same code for burning without needing to switch out the burn contract address

I think using a LogicSig here somewhat complicates the use of the contract when being called by existing smart contracts or through interfaces like Dappflow that can provide interactivity for apps with ARC-4 metadata.  My preference would be to let it be a simple app.

We could update the ARC with the three networks' contract addresses; this is typical for other one-per network smart contract architectures like bridges.

acfunk:
> I have to ask though: now that we have simplified this to the point of just opting in, would an logic sig suit better than a contract? It would have the benefit of being the same on all 3 networks, so contracts on testnet and mainnet would share the same code for burning without needing to switch out the burn contract address

A logic sig also has the benefit of being simpler to read/audit. Here is an implementation of a burning lsig (this was done prior to this PR).

https://algotools.org/burn

joe-p:
The main problem with an lsig at this point is that it's not usable by smart contracts. To me, it's not worth the tradeoff

acfunk:
Granted, a contract can't make the lsig opt-in, but it is useable for burning after the opt-in. But while the contract-to-contract approach does allow for programmatic opt-in, the logic and funds must be present for it to work properly.  It might be just as easy, or easier, for a contract/asset creator to pre-opt-in the burn account.

SilentRhetoric:
The contract is **extremely** simple, so it is not clear to me what is really gained in audit-ability via the logicSig approach.  

I think interoperability is already an area of friction in the on-chain ecosystem and so am going to hold the line on using an application approach which enables easier contract-to-contract composability.

PR #264, opened 2024-01-02
ARC-57: Application Action Routing

joe-p:
This app provides standardized TEAL that a developer can use to allow tools such as explorers or wallets easily determine if an app can be updated or deleted (among other actions)

### DIFF
--- a/ARCs/arc-0057.md
+++ b/ARCs/arc-0057.md
@@ -26,7 +26,7 @@ The TEAL for an application **MUST** start with the following lines:
 ```
 txn ApplicationID
 !
-pushint 6
+pushints 6
 *
 txn OnCompletion
 +
--- a/ARCs/arc-0057.md
+++ b/ARCs/arc-0057.md
@@ -1,9 +1,9 @@
 ---
-arc: <to be assigned>
+arc: 57
 title: Application Action Routing
-description: Standardized handling of OnCompletion and ApplicationID actions 
+description: Standardized handling of OnCompletion and ApplicationID actions
 author: Joe Polny (@joe-p)
-discussions-to: <URL>
+discussions-to: https://github.com/algorandfoundation/ARCs/issues/264
 status: Draft
 type: Standards Track
 category: ARC
--- a/ARCs/arc-draft_action_routing.md
+++ b/ARCs/arc-draft_action_routing.md
@@ -1,0 +1,67 @@
+---
+arc: <to be assigned>
+title: Application Action Routing
+description: Standardized handling of OnCompletion and ApplicationID actions 
+author: Joe Polny (@joe-p)
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: ARC
+created: 2024-01-02
+---
+
+## Abstract
+This ARC provides a standard way to handle an application call depending on the combination of the `OnCompletion` and `ApplicationID`.
+
+## Motivation
+Currently there is no way for someone to tell whether an app can be updated or deleted without doing a full static analysis on the TEAL.
+
+## Specification
+
+### TEAL
+The TEAL for an application **MUST** start with the following lines:
+
+```
+txn ApplicationID
+int 0
+>
+int 6
+*
+txn OnCompletion
++
+switch create_NoOp create_OptIn NOT_IMPLEMENTED NOT_IMPLEMENTED NOT_IMPLEMENTED create_DeleteApplication call_NoOp call_OptIn call_CloseOut NOT_IMPLEMENTED call_UpdateApplication call_DeleteApplication
+NOT_IMPLEMENTED:
+err
+```
+
+For every action (combination of `ApplicationID === 0` and `OnCompletion`) there is a label that corresponds to that action. If that action is not implemented in the contract, its label should be replaced by a label that simply contains `err`. In the above TEAL, this is the `NOT_IMPLEMENTED` label.
+
+The label containing the `err` opcode **MUST** come right after the `switch` opcode.
+
+The specific label names in the pre-compiled TEAL **MAY** be different, but it is **RECOMMENDED** to use the label names here for consistency across tooling.
+
+The `switch` statement in a contract **SHOULD** only include labels until the last supported action is reached. For example, if an application only supports `NoOp` creations, the switch can simply be `switch create_NoOp`. All other actions will simply go to the next opcode, which **MUST** be `err` as specified above.
+
+### Client Parsing
+
+A client **MUST** parse the decompiled TEAL to ensure that the firt ten opcodes match the opcodes specified in the TEAL specification. `intcblock` and `bytecblock` opcodes **MUST** be ignored if they are the first or second opcodes executed.
+
+The client **MUST** follow the label for a specific action. If that label does not exist in the `switch` opcode OR if it leads to the `err` label, then that action is known to not be supported by the TEAL.
+
+## Rationale
+Standardizing the initial opcodes of the TEAL program makes it easy for anyone to parse the TEAL, follow a label for a specific action, and and determine whether that action is supported or not. 
+
+## Backwards Compatibility
+N/A
+
+## Test Cases
+N/A
+
+## Reference Implementation
+TODO: Client parsing example
+
+## Security Considerations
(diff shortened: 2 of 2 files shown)


ARCs/arc-draft_action_routing.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Specification
+
+### TEAL
+The TEAL for an application **MUST** start with the following lines:
+
+```
+txn ApplicationID
+int 0
+>
```
nullun, line 27: You could save an opcode by doing the following. But I think it would swap the "create_" and "call_" labels the other way round on the `switch`, if that's acceptable.

```suggestion
txn ApplicationID
!
```

joe-p: Nice! Changed in [cdfbdb7](https://github.com/algorandfoundation/ARCs/commit/cdfbdb7fb335667b24b20735ce498095eaaac4c1)

ARCs/arc-draft_action_routing.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Specification
+
+### TEAL
+The TEAL for an application **MUST** start with the following lines:
+
+```
+txn ApplicationID
+!
+int 6
```
jannotti, line 27: When you disassemble a contract, you won't see `int 6`.  You might see `intc X` or `pushint 6`

ARCs/arc-draft_action_routing.md
```diff
@@ -0,0 +1,66 @@
...
+*
+txn OnCompletion
++
+switch call_NoOp call_OptIn call_CloseOut NOT_IMPLEMENTED call_UpdateApplication call_DeleteApplication create_NoOp create_OptIn NOT_IMPLEMENTED NOT_IMPLEMENTED NOT_IMPLEMENTED create_DeleteApplication
+NOT_IMPLEMENTED:
+err
+```
+
+For every action (combination of `ApplicationID === 0` and `OnCompletion`) there is a label that corresponds to that action. If that action is not implemented in the contract, its label should be replaced by a label that simply contains `err`. In the above TEAL, this is the `NOT_IMPLEMENTED` label.
```
jannotti, line 36: An offchain parser has to check the destination of the label, that it contains `err` only?  This is going to lead to extreme fragility in that we now have to worry about badly written parsers of TEAL code. If we change the disassembly output slightly (for example, add comments to a line, as we sometimes do), will they break?  If not, they must be implementing a fully correct TEAL parser?

Since you're asking them to skip constant blocks, they have to handle all possible ways to write constants.  There's base32 and base64 strings, for example.

You have to use this switch that takes two bytes per label, 26 bytes, even if you have much simpler needs, like an immutable contract?

This entry point is certainly one way to do things, and I know TealScript does. But nothing else does, and it doesn't seem best for immutable contracts.

ARCs/arc-draft_action_routing.md
```diff
@@ -0,0 +1,66 @@
...
+
+The label containing the `err` opcode **MUST** come right after the `switch` opcode.
+
+The specific label names in the pre-compiled TEAL **MAY** be different, but it is **RECOMMENDED** to use the label names here for consistency across tooling.
+
+The `switch` statement in a contract **SHOULD** only include labels until the last supported action is reached. For example, if an application only supports `NoOp` creations, the switch can simply be `switch create_NoOp`. All other actions will simply go to the next opcode, which **MUST** be `err` as specified above.
+
+### Client Parsing
+
+A client **MUST** parse the decompiled TEAL to ensure that the firt nine lines match the lines specified in the TEAL specification. `intcblock` and `bytecblock` opcodes **MUST** be ignored if they are the first or second opcodes executed.
```
jannotti, line 46: I read the second sentence as a contradiction to the first.

ARCs/arc-draft_action_routing.md
joe-p: Ah right. My understanding is that an explicit `pushint` doesn't get added to the constant block, is that true? If so I can change the TEAL to use `pushint`

ARCs/arc-draft_action_routing.md
joe-p: > An offchain parser has to check the destination of the label, that it contains err only? This is going to lead to extreme fragility in that we now have to worry about badly written parsers of TEAL code. If we change the disassembly output slightly (for example, add comments to a line, as we sometimes do), will they break? If not, they must be implementing a fully correct TEAL parser?

As long as nothing is added *before* the opcodes then parses shouldn't have a problem if they check to ensure only the beginning matches. I can update the ARC to be more clear about this. 

> Since you're asking them to skip constant blocks, they have to handle all possible ways to write constants. There's base32 and base64 strings, for example.

I'm not sure I follow. If parsers ignore `bytecblock` and `intcblock` lines there aren't other types of constant blocks, are there? I know there are pseudo-ops for different encodings, but aren't those ultimately compiled to either `pushbytes` or put in the `bytecblock`?

> This entry point is certainly one way to do things, and I know TealScript does. But nothing else does, and it doesn't seem best for immutable contracts.

A side benefit of this approach is that you can check what other "actions" are supported, but I don't think that's very important. 

The alternative, as you suggested on discord would be asserts in the beginning. Parsers would check for `txn OnComplete; pushint DeleteApplication; !=; assert` or `txn OnComplete; dup; pushint DeleteApplication; !=; assert; pushint UpdateApplication; != assert` (I think checking for update is just as important as being able to check for delete). The main downside of this approach is extra ops/bytes in the program, but it's not a lot.

ARCs/arc-draft_action_routing.md
joe-p: Yeah I can make this more clear

FrankSzendzielarz:
What's the motivation behind this? The stated motivation doesn't explain what problem is being solved or goal is being achieved.

joe-p:
Just updated the motivation and rationale sections

ARCs/arc-draft_action_routing.md
```diff
@@ -0,0 +1,72 @@
...
+### Client Parsing
+
+A client **MUST** parse the decompiled TEAL to ensure that the firt nine lines match the lines specified in the TEAL specification. Prior to parsing, `inctblock` or `bytecblock` opcodes **MUST** be removed from the TEAL because they may come before the opcode in this ARC, but don't affect functionality.
+
+To support future changes where comments may get added to the decompiled TEAL, clients **MUST** only check the beginning of lines to ensure the opcode and its argument(s) match the specification.
+
+The client **MUST** follow the label for a specific action. If that label does not exist in the `switch` opcode OR if it leads to the `err` label, then that action is known to not be supported by the TEAL.
+
+## Rationale
+Standardizing the initial opcodes of the TEAL program makes it easy for anyone to parse the TEAL, follow a label for a specific action, and and determine whether that action is supported or not. This makes it possible for tools in the ecosystem, such as explorers and wallets, to tell end-users whether apps they are interacting with are mutable or not.
```
nullun, line 55:
```suggestion
Standardizing the initial opcodes of the TEAL program makes it easy for anyone to parse the TEAL, follow a label for a specific action, and determine whether that action is supported or not. This makes it possible for tools in the ecosystem, such as explorers and wallets, to tell end-users whether apps they are interacting with are mutable or not.
```

ARCs/arc-0057.md
```diff
@@ -0,0 +1,72 @@
...
+type: Standards Track
+category: ARC
+created: 2024-01-02
+---
+
+## Abstract
+This ARC provides a standard way to handle an application call depending on the combination of the `OnCompletion` and `ApplicationID`.
+
+## Motivation
+App mutability is an important factor in a rational actors decision making when it comes to interacting with an app and sending funds to a contract account. If a contract is mutable, it is possible that funds could be stolen or lost after the initial program is changed or deleted. Currently, there is no easy way for an end-user to determine the mutability of a contract. It is possible to use static analysis on the decompiled TEAL, but this is not easily accessible and unrealistic for wallets and explorers to implement for every program.
```
tristanmenzel, line 17: > App mutability is an important factor in a rational actors decision making when it comes to interacting with an app and sending funds to a contract account. If a contract is mutable, it is possible that funds could be stolen or lost after the initial program is changed or deleted.

I'm not really sure I see the point of this ARC as even an immutable contract could steal your funds if you don't understand all the code. It might just provide a false sense of security.

joe-p: There is no way to have 100% security guarantees. A bug in algod could result in loss of funds, but that doesn't mean we should discount all other security considerations. 

> if you don't understand all the code

The goal of this ARC is to make it easier to understand critical parts of the code.

FrankSzendzielarz: Support the idea of dealing with this at avm level. I have seen chat recently about legal issues of mutable contracts, so the requirement seems to be one of some consequence.

joe-p:
No longer feel this is needed and there may be better approaches.

PR #269, opened 2024-01-09
ARC-0058: Plugin-Based Account Abstraction

joe-p:
This ARC proposes a standard for using stateful applications and rekey transactions to enable account abstraction on Algorand. The abstracted account is controlled by a single stateful application which is the auth address of the abstracted account. Other applications can be used as plugin to provide additional functionality to the abstracted account.


ARCs/arc-draft_plugin_abstraction.md
```diff
@@ -0,0 +1,177 @@
...
+      "name": "verifyAppAuthAddr",
+      "desc": "Verify the abstracted account is rekeyed to this app",
+      "args": [],
+      "returns": {
+        "type": "void"
+      }
+    },
+    {
+      "name": "rekeyTo",
+      "desc": "Rekey the abstracted account to another address. Primarily useful for rekeying to an EOA.",
```
robdmoore, line 69: EOA?

ARCs/arc-draft_plugin_abstraction.md
```diff
@@ -0,0 +1,177 @@
...
+### Plugins
+TODO
+
+### Wallet and dApp Support
+TODO
+
+## Rationale
+
+### App vs Logic Sig
+There have similar propsoals for reducing end-user friction, such as [ARC47](./arc-0047.md) which enables safer usage of delegated logic signatures. The major downside of logic signatures is that they are not
```
robdmoore, line 154: Half finished sentence ;)

joe-p: Fixed in f797187

ARCs/arc-draft_plugin_abstraction.md
joe-p: Externally Owned Account. Added definition in f797187

jannotti (commented):
I didn't really follow how this is going to be used, I mostly just have spelling corrections.

I do think the prose would read better if you removed all instances of "auth-addr" and just talking about an account being rekeyed.

joe-p:
> I didn't really follow how this is going to be used, I mostly just have spelling corrections.

I really need to enable a spell checker in my VSCode...

But the main idea of this ARC is to allow an account to be controlled by a smart contract but still have a safe way to rekey back to an EOA so that it can be used with any application. Taking a look at tests might give an idea on how its used, granted they are bit lacking in comments/explanation. https://github.com/joe-p/account_abstraction/blob/main/__test__/abstract_account_plugins.test.ts

ARCs/arc-0058.md
```diff
@@ -0,0 +1,410 @@
...
+### Definitions
+**External Owned Account (EOA)** - An account that is NOT controlled by a smart contract.
+
+**Abstracted Account** - An account that has functionality beyond a typical keypair-based account.
+
+**Abstracted Account App** - The stateuful application used to control the abstracted account. This app's address is the `auth-addr` of the abstracted account.
+
+**Plugin** - An additional application that adds functionality to the **Abstracted Account App** (and thus the **Abstracted Account**).
+
+**Admin** - An account, sepererate from the **Abstracted Account**, that controls the **Abstracted Account App**. In patiular, this app can initiate rekeys, add plugins, and transfer admin.
```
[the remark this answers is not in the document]
pbennett: That description doesn't seem correct.  The 'abstract account' should have a name here as it's the account actually being transacted (later).  So when you say " If Alice wants to rekey ALICE, she uses ADMIN to call rekeyTo which will rekey ALICE." - ALICE isn't rekeyed.  Instead, the Abstracted Account Application, AAA(') **contract account** (which we'll call Alice') is rekeyed to ALICE, allowing ALICE to then sign all txns for Alice'.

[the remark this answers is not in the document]
kylebeee: > The 'abstract account' should have a name here as it's the account actually being transacted (later)

Not necessarily, the abstract account contract may control itself or a regular account. in Joe's example the account its controlling ( and thus managing rekeys for ) is the standard account `ALICE`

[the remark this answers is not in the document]
pbennett: Deleted my comment.  Strange - I wonder if I saw a different version but I missed ALICE being rekeyed to the AAA prior to the rekeyTo.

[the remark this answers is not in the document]
joe-p: Yeah on creation you can supply the address the app controls. If you give it a zeroAddress then it just uses the app account. This is what's currently shown in the tests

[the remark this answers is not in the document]
joe-p: I will say though I'm back and forth on whether I should just remove that zeroAddress feature. Ie. a address must always be specified on create.

ARCs/arc-0058.md
```diff
@@ -0,0 +1,410 @@
...
+
+    const app = this.namedPlugins(name).value;
+    this.namedPlugins(name).delete();
+    this.plugins(app).delete();
+  }
+}
+```
+https://github.com/joe-p/account_abstraction.git
+
+TODO: Migrate to ARC repo, but waiting until development has settled.
```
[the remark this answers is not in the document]
joe-p: Eventually I'd like to link to the TEAL in the ARC but don't want to merge it into this repo yet because it'll be a PITA to continuously update it as there are changes being made, thus the TODO and link to the github

ARCs/arc-0058.md
```diff
@@ -0,0 +1,520 @@
...
+  namedPlugins = BoxMap<bytes, PluginsKey>({ prefix: 'n' });
+
+  /** The address this app controls */
+  controlledAddress = GlobalStateKey<Address>();
+
+  /**
+   * Ensure that by the end of the group the abstracted account has control of its address
+   */
+  private verifyRekeyToAbstractedAccount(): void {
+    const lastTxn = this.txnGroup[this.txnGroup.length - 1];
```
k13n, line 302: For better composability, could we make this more dynamic and allow the rekey to happen before the last transaction in a group?

As an example, one could do it like Tinyman V2 does it in its flash loans: they have two methods`flash_loan` and `verify_flash_loan` that reference each other and check that the counterpart exists.

kylebeee: Hey! If you look just below this comment you'll see the implementation does allow for rekeys earlier in the group and if so it just asserts that its been rekeyed back properly

joe-p: Yeah I would imagine `arc58_verifyAuthAddr` is the equivalent to  `verify_flash_loan`. The last txn must either be an explicit rekey or a call to `arc58_verifyAuthAddr`.

k13n: Thanks for the clarification. The difference is that `verify_flash_loan` doesn't have to be the last transaction in the group, while here either an explicit rekeying or `arc58_verifyAuthAddr` have to be the last transaction in the group (if I understand it correctly). My point is that for better composability it would be nice if one could take ARC-58 transactions and freely add more transactions in front or at the end of a group.

joe-p:
Made some changes today:

* As per @k13n suggestion, the rekey back or the `arc58_ verifyAuthAddr` can now be anywhere in the group after the initial rekey
* Renamed some variables/args to be a little bit more descriptive
* Added some additional context to the diagrams in the ARC and potential use cases making it clear they should be seperate discussions but included to demonstrate what potential usage of this ARC might look like.

ARCs/arc-0058.md
k13n: Awesome! One question: do you think running out of opcode budget could be a problem because of the loop in case the transaction group is big? A small optimization could be to start the for loop at one past the current group index.

joe-p: Ah yeah good catch on the starting index. I was originally thinking that additional txns in the group would be app calls but I suppose that's not always the case. I will add a worst-case scenario to my tests and see if the budget is exceeded.

ARCs/arc-0058.md
```diff
@@ -0,0 +1,538 @@
...
+#### Viewing Plugins
+
+For a given Abstracted Account, a wallet **SHOULD** allow the user to view all of the plugins added to their Abstracted Account App. Wallets also **SHOULD** provide the user a way to manually add and remove plugins from their Abstracted Account App.
+
+
+#### Supporting EOA Rekeys
+
+If a user connects to an app with an Abstracted Account the app **SHOULD** allow the user to easily sign transactions with an externally owned account. It is easy for the admin to manually call `rekeyTo` prior to interacting with an app, but this does not gurantee the user will rekey back to the Abstracted Account (this breaking plugin functionality). As such, to improve user experience, applications **SHOULD** be able to send transactions groups that start with a `rekeyTo` call that rekeys to the admin admin account with `flash` set to true. Because `flash` is set to true, the last transaction in the group **MUST** be a rekey from the admin address to the Abstracted Account address.
+
+This requires wallets to be able to tell apps that the connected account is an Abstracted Account.
```
k13n, line 232: This is probably one of the hardest problems to tackle to gain adoption of abstracted accounts. Currently, it's not possible to tell a dapp that an account is an abstracted account. But let's assume there is, then every dapp would have to special-case abstracted accounts and sandwich their transactions in between `arc58_rekeyTo` and `arc58_verifyAuthAddr`.

This got me thinking if there's a possibility to make abstracted accounts oblivious to dapps. Right now, by default, the auth-addr of the abstracted account is the smart contract, while the admin and plugins can temporarily gain control over the account. Could it be done the other way around? That is, by default the auth-addr of the abstracted account is the admin, while the SC & plugins can temporarily gain control over the account.

This way, the admin could interact with dapps normally and sign on behalf of the abstracted account, and plugins can gain temporary access to the account if needed.

Implementation-wise, I'm not sure what the most elegant solution is, but one approach could be to use logic-sigs. The abstracted account starts as a regular key-based account. Then, the account deploys the ARC-58 smart contract and signs a logic-sig that anyone can use to call `arc58_rekeyToPlugin` or `arc58_rekeyToNamedPlugin`. This signed logic-sig could be stored in a box in the smart contract. The admin, and only the admin, can add or remove any (named) plugin as they do now with ARC-58, nothing would change in that regard. If anyone wants to call a plugin, they would have to use the signed logic-sig that would ensure that by the end of the transaction group the admin is again the auth-addr of the abstracted account.

This approach has of course its downsides, too. For one, it requires logic-sigs with all their implied problems. Plugins couldn't be called app-to-app, thus limiting composability. Logic-sigs couldn't be revoked (but plugins could be revoked at any time). I don't know if there's a solution that wouldn't require a logic-sig.

joe-p: `arc58_rekeyTo` doesn't need an atomic rekey back. It can be required by the caller, but not necessary. So wallets can receive an incoming txn group, see that the user will need to sign it and form three different txn groups

1. `arc58_rekeyTo(admin)`
2. ...dappTxns
3. Rekey back to app

Since we are rekeying to the admin, its ok that the rekey back isn't atomic since we know the user presumably has control over the admin account. I think the lsig approach would have too many compromises.

k13n: That's true, but I'm afraid the UX isn't great:
- A user may close the wallet midway through the process or the wallet might crash for some reason (OS kills it because of OOM, etc). Then the account is still rekeyed to the admin, which wouldn't be too problematic but the plugins wouldn't work anymore.
- A dapp may gather signatures now but submit a transaction later. Waiting for an unknown amount of time to rekey the transaction back is not good. Right now, some dapps, for example, let users sign a 0-ALGO txn for authentication but never submit it to the network.
- Users are told that rekeying is powerful and potentially dangerous. Wallets can detect `arc58_rekeyTo(admin)` and in that case show no / less / different warnings for the rekey back to app, but if the user signs on a hardware ledger, there will always be this warning.

These may be edge cases, but they do happen. I'm just trying to figure out how the smoothest UX could look like for end users to adopt abstracted account (which I hope they do)

kylebeee:
Amazing work @k13n! Would love to link up and discuss arc58's potential and how you envision it being used even further than these plugins here if you have some time!

k13n:
> Would love to link up and discuss arc58's potential and how you envision it being used even further than these plugins here if you have some time!

Absolutely, I'd love to.

k13n:
I've opened a [security-relevant issue](https://github.com/blockshake-io/account_abstraction/issues/1) in my fork, it would be great to get your feedback on that! Maybe I should create the issue in this repo here instead?

emg110:
Hello everyone, when I first checked this reading of "Rekeying account to stateful contract" was enough to make me very happy but recently I dug in details and have some concerns I list here:
1- Aside from opt-in remediation what specific benefits proposed abstract accounts bring to users (I have red @k13n fork as well and saw that subscription plugin but that is 100% achievable without abstract accounts and with one signature for recurrent payments as well and this proposal does not add any advantage to that).
2- Given the fact that each rekey is risky and this ARC proposing all operations based on rekeys and as of results many risk prone events happen, one may ask: Isn't there a simpler and less risky way?
3- One of the main usecases in this proposal is onboarding: Using such a complex and risk-prone (each plugin acceptance may be an existential threat to the initial account) is not proportional to the issue it is going to solve, IMHO. The risk vs reward does not match.
4- In ARC content there are no propositions on Undoing/Canceling account abstraction and no methods for it.
5- The mechanics of Plugin verification is unclear and ARC just invites users to DYOR which is contradictional with onboarding mission of this ARC. I personally think this ARC at best can be for advanced users and most certainly not for onboarding.
6- For Onboarding simplicity is the king and explaining about how this all works to a simple user is a hard job.
7- This ARCs needs to be either supported by all wallets (since it is dealing with an account type) or none to mitigate the risk of Algorand blockchain foundamental features consistency (e.g. all wallets import and export accounts using Mnemonics but those will not work with abstract accounts if a wallet does not support this ARC). I personally think that foundamental features specifications and custom implementations targeting those, should only be allowed if accepted and implemented by all (and absolutely all) of wallet providers, no matter what size!
8- How would a user use ARC58 abstract account to make a simple pay or axfer transaction? 
9- What are the hard guarantees that ensures and verifies plugin rekeys back
10- Finally I humbly suggest considering full separation of abstract account and plugins and make "Abstract Account Plugins" a separated ARC because plugins need a much deeper dive and harder effort on security therefore in need of their own separated concern and ARC.
Again I am so grateful to @joe-p and @kylebeee for all amazing work we all benefit from as a community. These were just humble concerns of a builder with all love and appreciation.

joe-p:
> Dropped the named plugins because they seemed somewhat redundant to me

The rationale for the named plugins was to allow for discoverability for ARC-based plugins. For example, let's say there's `ARCXYZ` that defines the interface for an opt-in plugin. That ARC defines an app that everyone uses for opt-ins, but then another ARC comes along that implements the `ARCXYZ` interface but only allows opt-ins from whitelisted creators. In both cases a dapp/wallet might want to try to use the `ARCXYZ` plugin, but they won't know what app ID to use. Instead, they just specify `ARCXYZ` as the plugin and whatever app the user set as their `ARCXYZ` plugin would be used.

> Added two plugins (account recovery & spending-limits) that are interesting from a wallet's perspective

Great! Love seeing more use cases for this ARC as it gives more context into how it can be used

> I changed the arc58_changeAdmin method to implement this plugin. Previously, only the admin account could change the admin, but now the admin account and the abstracted account can call arc58_changeAdmin.

I understand the value here, but I am concerned it opens up a new security risk. Currently, an admin account is always guaranteed to have control over the account regardless of what plugins do. Even if a malicious plugin drains the account, the admin still has the ability to revoke the malicious plugin and get control of the account back. With this change a malicious plugin can completely take control of the account (akin to a delegated lsig being used for a rekey).

Perhaps if we want this feature it's an additional flag that exists within ARC58 itself and wallets give an extra warning when a plugin is requesting that permission.

> I've opened a security-relevant issue in my fork, it would be great to get your feedback on that! Maybe I should create the issue in this repo here instead?

Good catch! I think the solution here is to simply change

```ts
if (txn.sender === this.controlledAddress.value && txn.rekeyTo === this.getAuthAddr())
```

to

```ts
txn.sender === this.controlledAddress.value &&
  txn.rekeyTo === this.controlledAddress.value;
```

As for the logistics of implementation issues, I think linking to an external issue is fine for now. Perhaps make the issue upstream in `joe-p/account_abstraction` just for better traceability?

> 1- Aside from opt-in remediation what specific benefits proposed abstract accounts bring to users (I have red @k13n fork as well and saw that subscription plugin but that is 100% achievable without abstract accounts and with one signature for recurrent payments as well and this proposal does not add any advantage to that).

Some of the use cases for ARC58 can be done via delegated lsigs, but ARC58 adds additional security and composability benefits. For example, ARC58 enables revocation and timeout-based authorizations. ARC58 is also inherently more transparent because for any given account I can lookup all the plugins that current have authority over the account. This is impossible with delegated lsigs because they are simply signatures. If the lsig hasn't been used yet, there is no way to tell what signatures currently exist for a given address.

Because ARC58 is app-based, it enables other apps to leverage plugins, which is something that is not possible with lsigs.

> 2- Given the fact that each rekey is risky and this ARC proposing all operations based on rekeys and as of results many risk prone events happen, one may ask: Isn't there a simpler and less risky way?

I think the only other technical solution given the tools we currently have are delegated logic signatures, which have the security and composability concerns mentioned above.

> 3- One of the main usecases in this proposal is onboarding: Using such a complex and risk-prone (each plugin acceptance may be an existential threat to the initial account) is not proportional to the issue it is going to solve, IMHO. The risk vs reward does not match.

I agree. In many cases there are simpler solutions for onboarding an EOA. For example ARC59 for asset opt-ins. This ARC, however, is much more flexible and provides a framework for more advanced onboarding scenarios. The burder of this complexity lies upon the onboarder not the one getting onboarded.

> 4- In ARC content there are no propositions on Undoing/Canceling account abstraction and no methods for it.

Undoing the account abstraction is simply a matter of rekeying to an EOA. Once the `controlledAddress` is rekeyed to an EOA, plugins will no longer be able to send transactions.

> 5- The mechanics of Plugin verification is unclear and ARC just invites users to DYOR which is contradictional with onboarding mission of this ARC. I personally think this ARC at best can be for advanced users and most certainly not for onboarding.

Agreed. Plugin verification is outside of the scope of this ARC but very important for adoption.

> 6- For Onboarding simplicity is the king and explaining about how this all works to a simple user is a hard job.

I somewhat agree but I also think it's all about abstraction and UX. When we onboard someone by having them create a keypair they don't need to understand eliptic-curve cryptography. When someone signs into Amazon with passkeys they have no knowledge of the underlying cryptography. The ultimate goal is to abstract away the details of keypair management and ARC58 provides a framework for one piece of that puzzle.

> 7- This ARCs needs to be either supported by all wallets (since it is dealing with an account type) or none to mitigate the risk of Algorand blockchain foundamental features consistency (e.g. all wallets import and export accounts using Mnemonics but those will not work with abstract accounts if a wallet does not support this ARC). I personally think that foundamental features specifications and custom implementations targeting those, should only be allowed if accepted and implemented by all (and absolutely all) of wallet providers, no matter what size!

Agreed. This is why the Wallet Council and ATAC exist.

> 8- How would a user use ARC58 abstract account to make a simple pay or axfer transaction?

The `controlledAddress` needs to be rekeyed to an EOA which signs the transaction and then rekeyed back to the abstracted account app. The best way to do this from wallets/dApps is still an important question that we need to answer.

> 9- What are the hard guarantees that ensures and verifies plugin rekeys back

`verifyRekeyToAbstractedAccount` ensures that by the end of the atomic group the `controlledAddress` is rekeyed back to the abstracted account app. This is always used when a plugin is used and it's an optional feature of rekeying (called "flash rekeying")

> 10- Finally I humbly suggest considering full separation of abstract account and plugins and make "Abstract Account Plugins" a separated ARC because plugins need a much deeper dive and harder effort on security therefore in need of their own separated concern and ARC.

Yes agreed. The actual plugins themselves are outside of the scope of ARC58 and should be their own ARCs, but having use-cases for ARC58 makes it easier to discuss what we really need from this ARC.

emg110:
Thank you for response @kylebeee , I appreciate this and the great work you guys are doing very much.

Am I correct to assume the only unique feature this ARC provides is addressing opt-in fricktion? Would you mind kindly list those unique benefits or refer me to them on document (features that are only possible using this ARC)?

I just am trying to underestand what that overhead complexity and risk of many rekeys between different plugins and the overhead of maintaining trust between all those is worthing to not having optin.

Million thanks again

joe-p:
> Am I correct to assume the only unique feature this ARC provides is addressing opt-in fricktion? Would you mind kindly list those unique benefits or refer me to them on document (features that are only possible using this ARC)?

This ARC simply provides a way to safely delegate control of one account to applications. Use cases like opt-in are provided as an example but not within the scope of this ARC. 

> Note: I am 100% in favor of abstract accounts and the initial approach of rekeying an account to smart contract address but the methodology details and excessive use of rekey is what concerns me and I think may be there are more effective ways. may be

There are two ways to delegate control of an EOA. Rekeying and signing a logic sig program. Lsigs have the problems mentioned above. This ARC utilizes rekeys but provides a way to ensure specific properties of the delegation, like who can initiate them, when they expire, and revocation.

k13n:
> The rationale for the named plugins was to allow for discoverability for ARC-based plugins. For example, let's say there's ARCXYZ that defines the interface for an opt-in plugin. That ARC defines an app that everyone uses for opt-ins, but then another ARC comes along that implements the ARCXYZ interface but only allows opt-ins from whitelisted creators. In both cases a dapp/wallet might want to try to use the ARCXYZ plugin, but they won't know what app ID to use. Instead, they just specify ARCXYZ as the plugin and whatever app the user set as their ARCXYZ plugin would be used.

Discoverability is an important topic here, but would the name really help? What if a plugin appears to implement an ARC based on its name but does not actually implement it? What if the plugin implements multiple ARCs?

I've been thinking how we'd likely implement this in Defly Wallet and this might address discoverability and some of the security issues that have been raised here.

From within Defly itself we'd only allow to enable a set of whitelisted plugins (opt-in, recovery, spending limit, etc.). Those plugins have to be open source and they have to be carefully vetted by ourselves (Defly) and the community. In addition, I believe plugins should be immutable. Once a plugin is approved, we'd compute the hashes of its compiled approval and clear-state program and hard-code them in Defly. When we look up the account state of an abstracted account, we'd look at all plugins in this account and show if one is approved according to the plugin's hash or not (if not we'd warn the user that she's got an unrecognized, unapproved plugin). For each whitelisted plugin we'd implement a custom UI in Defly to maximize its utility. For example, for a spending-limit plugin we'd have a UI where the user can set spending limits, monitor them, and delete them. We'd start with a very small set of plugins, but over time we'd hope that developers start writing their own plugins and ask them to be integrated into Defly (always following the process outlined above).

Just thinking out loud, but we could even deploy a "registry contract" that implements this whitelist of approved plugins on-chain. The `arc58_addPlugin` function could then ask that registry contract if the plugin's approval & clear-state program is whitelisted & approved. This would make the whitelist transparent. Of course this raises other questions, like who controls that registry, etc.

> I understand the value here, but I am concerned it opens up a new security risk. Currently, an admin account is always guaranteed to have control over the account regardless of what plugins do. Even if a malicious plugin drains the account, the admin still has the ability to revoke the malicious plugin and get control of the account back. With this change a malicious plugin can completely take control of the account (akin to a delegated lsig being used for a rekey).
Perhaps if we want this feature it's an additional flag that exists within ARC58 itself and wallets give an extra warning when a plugin is requesting that permission.

That's a fair point. My hope was that with the process outlined above where we'd only allow whitelisted plugins, we'd not run into a situation where a plugin maliciously tries to change the admin account. But having this extra permission flag also sounds like a nice approach!

> 7- This ARCs needs to be either supported by all wallets (since it is dealing with an account type) or none to mitigate the risk of Algorand blockchain foundamental features consistency (e.g. all wallets import and export accounts using Mnemonics but those will not work with abstract accounts if a wallet does not support this ARC). I personally think that foundamental features specifications and custom implementations targeting those, should only be allowed if accepted and implemented by all (and absolutely all) of wallet providers, no matter what size!

Hard disagree, @emg110. You need to let individual projects push their products and you cannot expect that one project holds back just because other projects don't pull along (immediately). You cannot control everything through committees & standardization bodies because that just kills innovation and slows everything down to a standstill. It's great that these committees exist and I'm a member of some of them because I think it's valuable to have them, but they cannot force individual projects to adopt or not adopt certain features.

> The actual plugins themselves are outside of the scope of ARC58 and should be their own ARCs, but having use-cases for ARC58 makes it easier to discuss what we really need from this ARC.

Agreed. I think it's good to have one or two example plugins in the ARC to showcase what can be done and how it can be done. In my fork I added two more plugins, but probably it's better if they are outside of this repo.

emg110:
Thank you for your response @k13n, I cannot find anything mentioning "Committee" or anything supervisory in any of my comments (where did that come from dude?). Those are just comments and concerns of one builder in ecosystem who happened to have witnessed more rainy days than many others.
That being said I invite you to kindly read my logical reasoning and rationale for that suggestion once more:

I simply stated: 
For fundamental features only (not everything) it would shatter the overall fundamental robustness and reliability image of the chain if something like account (especially with all that emphasis on onboarding in ARc58) goes south or faces a mismatch in features between wallets. Also, a fundamental feature like accounts (especially when used for onboarding) needs to be acting all the same across wallets for the sake of dApp interoperability as well. This is not true for most of the features which are best to be implemented by the highest innovation of their authors. My concern was just because Account is a fundamental feature and therefore IMHO better be harmonized.

Also just by moving plugins to a separate ARC and concern and also removal of the Onboarding mission for ARC58, 99% of those concerns will be auto-eliminated since no one objects to an innovative feature being available for advanced users who know what they are doing (even if it is just one wallet provider who innovates and deploy it)

Thanks again for very productive, non-personal and pure technical conversation we have here, from fruits of which whole community will benefit.

k13n:
No hard feelings MG, and I hope my writing isn't too blunt. I wasn't directing my statement at you, but wanted to state it more generally that I don't think all wallets need to do the exact same thing even if it concerns more fundamental issues. I made the reference to committees because ATAC and the wallet council was mentioned somewhere in this thread (maybe by someone else).

And I think with that I also explained why I disagree with your reasoning. People, especially newcomers, often don't realize that they can bring an account created in one wallet into another wallet. They often say they have a "Defly account" or a "Pera account". Hence, I think it is completely fine that one wallet offers an account type that other's don't. In fact, right now Defly offers multisig accounts whereas most other wallets don't. That's also a fundamental feature, but if we had to wait to deploy multisigs until everybody is on-board, we'd still be waiting right now :)

emg110:
that is indeed a general issue: The lack of harmony on fundamental features across wallets in Algorand. On that, I agree.
I agree that multisigs are a fundamental feature and I believe it should have been implemented by all wallets the same. The fact that we have a shortcoming on one subject does not dictate that we repeat the same mistake and add to that lack of harmony I honestly believe. 
IMHO, the main mission of those committees you mentioned is to bring that harmony.
I once again stand my original statement: Fundamental features (Algorand Protocol and core SDK level features) should be harmonized across all well-known wallets in an ecosystem, otherwise, it will bring a reputation and technology risk which leads to isolation of the ecosystem and decrease in user & dev base.

emg110:
Back to this ARC's main subject and just for the future record : I still firmly believe that **smart signatures** are the way to go for account abstraction. We cannot leave a foundamental feature of Algorand blockchain unused because it may have risks if not implemented correctly (it's like not using the web because it has obvious risks if not implemented securely). I firmly beleive that we should not hack our way around an available and powerful feature and start using it within wallet providers. Also plugin features as well can be achieved that way by either only using smart signatures or a combination of smart contract apps and smart signatures. it is much safer in sandwidching rekey transactions and re-use them over and over does not seem to me as a comprehensive architectural solution to abstract account requirement for Algorand.
Screenshot 2024-08-13 at 13 41 29

kylebeee:
Hey @emg110, i fundamentally disagree with you and think your point of view on this doesnt properly address the drawbacks and limitations that delegated smart signatures have. I think aside from you we have pretty unanimous agreement on this ARC and why its implemented the way it is

Missjones2829:
Merge

emg110:
> Hey @emg110, i fundamentally disagree with you and think your point of view on this doesnt properly address the drawbacks and limitations that delegated smart signatures have. I think aside from you we have pretty unanimous agreement on this ARC and why its implemented the way it is

Regarding user experience, how users differ ARC 58 and adverse Rekey activity:
https://x.com/PeraAlgoWallet/status/1930388959026393384

joe-p:
Haven't done a full review of the new ARC & impl but some high level feedback.

1. I definitely see a lot of value in the escrow system, but I am worried that it adds a lot of complexity to this ARC and locks us into ASA and ALGO. In the future users might want to use different types of SC-based assets which would fundamentally be incompatible with the current escrow system. 

2. Similarly, I am concerned about the ABI method selectors because it adds a lot of complexity to the app and does something that could be solved at the plugin level. There's also the concern of static analysis.

My general thinking is that we want this ARC to be as simple as possible to get initial adoption and then add additional features (complexity) on top of it once the main ARC is adopted. As is, I would imagine the timescale for ecosystem adoption is on the magnitude of years. Adding more complexity extends that timeline. Even if we all look at the code and agree that the features are nice, things will bubble up at various implementation points and add on to the development time. This is what happened with the ASA inbox, which took ~4 years to go from initial implementation/ARC to production use and that was for a huge pain point in the ecosystem (opt ins) that everyone wanted to solve. That ARC also only required wallet integration, whereas 58 requires both wallet and dapp integrations to be properly supported. 

I don't want to to be to pessimistic, but I also want to be realistic about what adoption of this ARC really looks like and keep that in mind as we add features to it.

kylebeee:
I understand & agree on most of what you're saying.

I do think the escrow system provides enough extra functionality & security that it should be part of this ARC.

In general i agree things should be as simple as they can be for what they are trying to achieve & i believe this ARC will eventually be the bedrock for which most users interact with the Algorand network.

We've been refining this smart wallet approach for a few years now & while it feels mature to me i recognize theres a large amount of tooling around this that needs to exist to make adoption easier & provide clarity to adopters both on the wallet & dapp side of integrations.

For now i am fine keeping this in a draft state. We can circle back to helping this ARC become a standard after the Akita wallet goes into production & we see adoption & demand for this as a standard start to grow along with more tools that make integration easier.

Roseymr (approved):

PR #270, opened 2024-01-11
ARC-0013: Web Extension Provider Interface

kieranroneill:
## Summary

This proposal intends to outline a common interface that, at it's core, will subscribe to some fundamental goals:

1. Creating a common interface between dapps and providers.
2. Allowing providers to work in unison.
3. Give full authority of provider choice to the user.

It consolidates the previous ARCs; wallet transaction signing ([ARC-0005](./arc-0005.md#specification)), wallet address discovery ([ARC-0006](./arc-0006.md#specification)), wallet transaction network posting ([ARC-0007](./arc-0007.md#specification)) and wallet transaction signing & posting ([ARC-0008](./arc-0008.md#specification)), to comprehensively outline interactions between a dapp and provider.

## Considerations

* The explicit use of the [`BroadcastChannel`](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) API is opinionated, but its use provides a more modern, secure and convenient API for communication within the context of browser extensions.


emg110:
Amazing! just since I already PRed an ARC with this name (ARC13) may I ask to change the ARC number and name please @kieranroneill ! Thank you in advance buddy!

PR #272, opened 2024-01-12
ARC-0027: Provider Message Schema

kieranroneill:
## Summary

This proposal intends to outline a common message schema to be used in dApp/provider (wallet) communication. 

The proposal attempts to build off the fundamental work of the previous ARCs; wallet transaction signing ([ARC-0005](./arc-0005.md#specification)), wallet address discovery ([ARC-0006](./arc-0006.md#specification)), wallet transaction network posting ([ARC-0007](./arc-0007.md#specification)) and wallet transaction signing & posting ([ARC-0008](./arc-0008.md#specification)), to comprehensively outline a common message schema, and to extend this schema with an aim to allow more robustness with multiple AVM (Algorand Virtual Machine) chains.

Aside from extending the schema of the previous ARCs, this proposal also adds a couple of new methods that aim to "fill the gap". These include:

* A discovery method that is used to get information about providers such as what methods are supported and what chains.
* A "disable" method that allows a dApp to remove itself from providers.


ARCs/arc-0027.md
```diff
@@ -0,0 +1,1611 @@
...
+              },
+              "methods": {
+                "type": "array",
+                "description": "A list of methods available on the provider for the chain",
+                "items": {
+                  "type": "string"
+                }
+              }
+            },
+            "required": ["id", "methods"]
```
drichar, line 312:
```suggestion
            "required": ["genesisHash", "genesisId", "methods"]
```

ARCs/arc-0027.md
```diff
@@ -0,0 +1,1611 @@
+---
+arc: 27
+title: Web Extension Wallet Interface
```
ehanoc, line 3: What do you think of keeping this agnostic to the platform? We could just say Wallet Interface for instance

ARCs/arc-0027.md
```diff
@@ -0,0 +1,1611 @@
...
+| A wallet provides a separate SDK that is installed by a dapp to interact with the wallet.                                                                                                                                                                                                                                                                                                                                            | - Allows wallets full control over how dapps interact with wallets, ie. provide a bespoke UX.                                                                                                                                                                                                                               | - Dapps need to install each SDK separately, thereby increasing the amount of development time, and complexity, needed to implement each wallet separately.                                                                      |
```
ehanoc, line 25: Wouldn't this be a single SDK for dapp developers? They would just need to comply with the API, no? 

Maybe that's something could be provided in the ARC as a reference implementation. Thoughts?

ARCs/arc-0027.md
```diff
@@ -0,0 +1,1611 @@
...
+
+### Overview
+
+The essence of the provider interface will lie in using the simple messaging bus mechanism of the <a href="https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API">BroadcastChannel API</a>.
+
+A dapp will broadcast a message and the provider's <a href="https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts">content scripts</a> will listen for these broadcasted messages and act upon them, responding with another broadcast message that references the original request broadcast.
+
+The dapp will, itself, listen to broadcasted responses from any providers. Below is a simple diagram for the flow of data for an enable request:
+
+![Simple BroadcastChannel API using web extensions](../assets/arc-0027/simple_message_bus.png)
```
ehanoc, line 58: I like it, is there samples for us to include here and test?

ARCs/arc-0027.md
```diff
@@ -0,0 +1,1611 @@
...
+  - **MUST** be a <a href="https://www.rfc-editor.org/rfc/rfc4122">UUIDv4</a> compliant string
+  - **MUST** uniquely identify the provider
+* `result.stxns`:
+  - **MAY** have an item that is a base64 encoding of a signed transaction
+  - **MAY** have an item that is null if the transaction at the same index in the request's `txns` list was not signed by the provider
+  - **MAY** be empty
+* `reference`:
+  - **MUST** be `arc0027:sign_txns:response`
+
+#### Sign Bytes
```
ehanoc, line 982: I'm a fan of us supporting signing for multiple purposes and arbitrary data signing. But `signBytes`: 
- seems to give freedom to the app to define what `purposes` this can be used for with a simple string. Seems that it can be gamed to trick users to sign things that they shouldn't, even transactions. 
- Doesn't consider multiple encoding schemes
- Validation of the content to be signed should be matched against a known schema? 

That's one of the things i'm trying to address in [ARC-52#SignMetadata](https://github.com/algorandfoundation/ARCs/pull/239/files#diff-5f82fbf5efe9e776afbc7c317caf254ed75d9428879851529dc85f34b7338345R41) is to enforce the context of what we are signing and validate it quickly from a list of possible schema templates (i.e authentication requests, messages, identity), so that the user doesn't have to understand msgpack encoding or CBOR and the wallet can still validate what is signing. 

It's a longer conversation though.

ehanoc (commented):
I'm a big big fan of this proposal. 

Overall, to finalize it, maybe we could include a reference implementation for us to test ourselves?

ARCs/arc-0027.md
kieranroneill: I wholeheartedly agree. I have been playing with this with [Algorand Provider](https://algorand-provider.agoralabs.sh/) + Kibisis and, if I recall, Kibisis displays the data if it is:
* A valid UTF-8 string
* A parseable JWT

For example, for JWTs, if it is correctly parsed it shows this screen:
Screen of Kibisis JWT signing

It also shows some warnings if the JWT has expired or the subject/sub of the JWT does not match the account to sign. 

As you can workout, I am a fan of JWTs :sweat_smile: 

(It was actually the only reason I built Kibisis as I wanted to implement, for the then, app I was building to use user JWT authentication using an Algorand account)

But you are right, the schema of this function requires more work as it is not fit for consumption.

ARCs/arc-0027.md
```diff
@@ -0,0 +1,1611 @@
...
+type: Standards Track
+category: Interface
+created: 2023-12-24
+---
+
+# Web Extension Provider Interface
+
+## Abstract
+
+Building off of the previous ARCs relating to; wallet transaction signing ([ARC-0005](./arc-0005.md#specification)), wallet address discovery ([ARC-0006](./arc-0006.md#specification)), wallet transaction network posting ([ARC-0007](./arc-0007.md#specification)) and wallet transaction signing & posting ([ARC-0008](./arc-0008.md#specification)), this proposal aims to extend and comprehensively outline a common interface between multiple web extension wallets and dapps by utilizing the pub/sub semantics of the <a href="https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API">BroadcastChannel API</a>.
```
kewde, line 16: Using a  `BroadcastChannel` is an interesting choice and would need a bit more inspection. AFAIK web extensions do not typically inject in iframes, with this approach we may be allowing embedded iframes to trigger a wallet popup as long as that origin is also open in a tab elsewhere.

ARCs/arc-0027.md
```diff
@@ -0,0 +1,1611 @@
...
+      "type": "object",
+      "properties": {
+        "host": {
+          "type": "string",
+          "description": "A domain name of the provider"
+        },
+        "icon": {
+          "type": "string",
+          "description": "A URI pointing to an image"
+        },
```
kewde, line 285: Can we force this to not be a remote URI? Ideally we could make it match https://github.com/wallet-standard/wallet-standard/blob/master/packages/core/base/src/wallet.ts#L24

ARCs/arc-0027.md
kewde: Do I understand it correctly that if multiple tabs are open and thus content scripts are listening to the BroadcastChannel, then they need to avoid registering multiple times? If registering multiple times is allowed, would the content script need to synchronize internally to ensure it only registers once and doesn't deal with the request multiple times?

ARCs/arc-0027.md
```diff
@@ -0,0 +1,1687 @@
...
+  - **MAY** be empty
+
+<sup>[Back to top ^][top]</sup>
+
+### Response Message Schema
+
+```json
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "/schemas/request-message",
```
PhearZero, line 592: Small Nit: Looks like this one should be response-message

kieranroneill: Thank you! One can miss stuff like this in a copy + paste :sweat_smile:

PhearZero: No worries! Trying out the specification in Kotlin/Java and was doing a lot of copy + paste myself!

SudoWeezy:
@kieranroneill How far are we to move this ARC to the review stage? 
I think it would be great to move forward with this one now that we have deprecated ARC 6-11

kieranroneill:
> @kieranroneill How far are we to move this ARC to the review stage? I think it would be great to move forward with this one now that we have deprecated ARC 6-11

@SudoWeezy Sounds good! I have been battle testing this proposed spec in [Kibisis](https://kibis.is/), through [AVM Web Provider](https://avm-web-provider.agoralabs.sh/) and [UseWallet](https://github.com/TxnLab/use-wallet), so I am happy to give feedback and discuss it further.

PR #289, opened 2024-03-23
arc-0003 - Better definition of Fungible assets

scholtz:
People do not use the ARC3 for fungible assets when minting tokens because pera explorer does not process fungible assets automatically because people are minting NFTs wrong.

I suggest this change to allow minting NFTs also in different then log(10) scales. If person wants to issue 42 tickets for the event he cannot do it according to this standard at the moment.

This lead to the situation that the automated scripts considers everything minted with this standard as NFT and even though there are fungible tokens minted with this standard it does not process it right.

The suggested solution is to allow minting of specific number of ASAs, for example 42, with decimals 0, and consider this as fractional NFT. This rule will have cap to 1000000 tokens. This will fix all minted NFTs from past to be correctly classified as NFTs and will allow correctly specify ARC3 Fungible tokens mints.

### DIFF
--- a/ARCs/arc-0003.md
+++ b/ARCs/arc-0003.md
@@ -86,6 +86,14 @@ An ASA is said to be a *fractional non-fungible token* (*fractional NFT*) if and
 
 > In other words, the total supply of the ASA is exactly 1.
 
+or 
+
+* *Total Number of Units* (`t`) **MUST** be less then or equal to 1000000.
+* *Total Number of Units* (`t`) **MUST** be equal to number of NFT fractions.
+* *Number of Digits after the Decimal Point* (`dc`) **MUST** be equal to zero.
+
+An ASA is said to be a *fungible token* (*FT*) if any of the conditions in this section are not met and complies with the rest of this standard.
+
 ### JSON Metadata File Schema
 
 > The JSON Medata File schema follow the Ethereum Improvement Proposal <a href="https://eips.ethereum.org/EIPS/eip-1155"> ERC-1155 Metadata URI JSON Schema</a> with the following main differences:


cusma:
Just chiming in to point out that this has been part of the [ARC-0003 discussion](https://github.com/algorandfoundation/ARCs/issues/3#issuecomment-882760108) back in the old days.

At that time there was not consensus abut calling such token NFTs. 

The most agreed proposal (which wasn't part of the final spec) was: adding the entity **Collection** to ARC-0003 spec, with the following requirement: `decimals = 0` (without any supply cap, which I think is totally arbitrary and useless).

This being said, I think this **should not** be part of the ARC-0003, which defines **NFTs**.

Instead, I think this COULD be a child-ARC inheriting the ARC-0003 spec entirely (to facilitate clients wrt URI/metadata/etc. already standardized), adding the requirement: `decimals = 0`.

scholtz:
@SudoWeezy 

> I am not sure I understand, but 42 tickets are Fungible Tokens, right? Why should they be considered Fractional NFT?

Because they are NFTs. You may call with this logic all fractional nfts the FTs.

If i have airplane ticket we call it NFT. Right?

If i have the ticket to the event which has 42 occupancy I call it NFT. Currently i cannot represent it using this standard.

If i have the ticket to the event which has 100 occupancy I call it NFT and currently i can represent it as fractionalized nft with 2 decimals and 100 amount.

Another example is .. there is one appartment which has 42 shares, first person owns 1 share and other person owns 41 shares. Currently you cannot create fractionalized NFT according to ARC3 standard.

@cusma Did we learn something since then? We just created issue where there is huge uncertaintly if there is NFT or FT and if you issue FT according to this standard it will not show in Pera until you contact them solve the ticket with them have administration overhead and pray that only they are not compliant with this standard.

cusma:
> If i have the ticket to the event which has 42 occupancy I call it NFT. Currently i cannot represent it using this standard.

Again, according to the public discussion happened over ARC-0003 the example you mentioned does not fall under the agreed definition of NFT (which is either "Pure" or "Fractional"). It's not me saying that, this was an open and public discussion that reached an agreement over what should be defined as NFT under this standard and the example you posed is not.

As I said before, the most agreed definition of the example you proposed was the concept of "Collection" (i.e. a set of unitary replicas of an asset, like 42 tickets, 420 copies of a gaming card, etc.). Each unit of the "Collection" is not divisible per-se but has different copies.

> Also worth mentioning the ARC3 was meant to define FTs as well and it was marketize that way and people did issue FT tokens according to ARC3 standard.
>
> New ARC is bad as it will create just more uncertainty how to deal with all tokens issued on algorand according to this standard.

We can debate about ARC-0003 being the best place to add this "Collection" concept. In this discussion we should consider all the pros and cons about an update or a child-ARC, given the current adoption of the ARC-0003. Again, I think this should go under a child-ARC but I could be convinced otherwise if there are reasonable arguments.

scholtz:
Man, if we approve this pull request we fix the history, fix bots, and everybody will be happy.

This is not true after we merge this pull request. `The category you mention is not an NFT according to the ARC-0003 standard`

Why do you have guys problem with solving issues what business have. More ARCs on this topic will create mess and will not fix the history and will not solve anything.

cusma:
> This is not true after we merge this pull request. `The category you mention is not an NFT according to the ARC-0003 standard`

I can propose a PR on which I define a "Cusma NFT" category that requires `total = 4269` and `decimal = 2` and solve my own use case. But this is not how the a common agreed standard works.

> Why do you have guys problem with solving issues what business have. More ARCs on this topic will create mess and will not fix the history and will not solve anything.
> 
> Dont try to make from algorand a dinousaur that cannot addopt to things found in the implementation phase.

Maybe I expressed myself wrong, let me iterate: what I'm saying is that ARC-0003 **already solves** your need, without any modification of the standard.

You don't need 42 fungible token to be defined as NFT in order to take advantage of the ARC-0003 standard.

scholtz:
> You don't need 42 fungible token to be defined as NFT in order to take advantage of the ARC-0003 standard.

One of us do not understand other one.

1) how does appartment which is divided into 42 parts can be issued as fractional nft using the arc-0003 standard?
2) my issue is that all my fungible assets i have minted under arc-0003 standard did not show up in the pera wallet automatically because pera wallet considers all arc0003 assets as nfts and they consider nothing as FT because this standard is bad and many people issue "according to them" nfts under arc0003 not in compliant way. And thats why the assets has to be manually processed by them each time new FT issued.

cusma:
> One of us do not understand other one.

Communication is notoriously hard.

1. ARC-0003 defines conventions for **fungible** _AND_ **non-fungible** tokens;
2. ARC-0003 identifies two kind of NFT: pure and fractional;
3. ARC-0003 discussion agreed that a **collection** (i.e. `decimals = 0`) was not strong enough to be distinguished from a **fungible** token.

> 1. how does appartment which is divided into 42 parts can be issued as fractional nft using the arc-0003 standard?

You don't issue it as a fractional NFT, but this doesn't mean you can't use ARC-0003.

Your example is a **fungible** token, which can still take advantage of ARC-0003 conventions. Quoting the abstract:

> The goal of these conventions is to make it simpler for block explorers, wallets, exchanges, marketplaces, and more generally, client software to display the properties of a given ASA.

I guess I'm repeating myself but: you can use ARC-0003 (e.g. Asset URL ends with `#arc3`, etc.) to apply conventions over an ASA with 42 units and 0 decimals, the standard allows it and clients should know how to treat this case if they implement the standard. You just don't call it a NFT.

cusma:
> @cusma I understand everything you said and i agree with your points if users would use it this way. Can you please react on why the pera is considering all ARC3s as NFTs and not the FTs?

@scholtz unfortunately I can not speak on behalf of whoever adopts the standard. I think wallets' teams would be better suited to address this question.

> How other people should issue FTs? There is no other ARC to my knowledge that allows you to issue FT. And usage of this arc solves the issue that the icon is automatically shipped to all wallets or products that is capable of reading the arc3. Just people do not use it because noone knows that thay can do it (for example in whole text except of header there is no mention of word fungible (this PR also solves this)).. So bots in pera and others considers all from arc3 as non fungible..

ARC-0003 scope was primarily the definition of conventions _"to make it simpler for block explorers, wallets, exchanges, marketplaces, and more generally, client software to display the properties of a given ASA"_. I'd say NFTs definition are just an bonus "label" introduced by the standard, not its "core". 

If ARC-0003 adopters decide to implement it partially (i.e. applying it only if ASA are labeled as NFT) I think we can not blame the standard for it, nor it means that the standard is "wrong" or "shortsighted".

This being said, I do think there are no sufficient elements to revert the public conversation over the distinction between "collections" (i.e. `decimals = 0`) and "fungibles" token.

Moreover I think that the additional requirement proposed:

`Total Number of Units (t) MUST be less then or equal to 1000000`

is arbitrary (why 1000000 and not 1000000000?).

> I agree with your point that Communication is notoriously hard.. Please be more dev friendly.. PLS

If I seemed rude I apologize, when I said that communication is hard I meant that I probably wasn't expressing myself clearly.

scholtz:
> is arbitrary (why 1000000 and not 1000000000?).

Because with higher numbers it is preferable to use the 1/N method so that the sum of all parts gives the 1.

Its unlikely that someone will issue 1000001 NFTs and he will want to call it fractional NFT.

But if you want we can set it to any number where decimals equal zero, and have the fractional nfts be called everything that has decimals zero. I have not seen any FT that does not have any decimals. What do you think?

I wonder what others have to say about the proposed change..

cusma:
> Can you please react on why the pera is considering all ARC3s as NFTs and not the FTs?

@scholtz seems [this seems like a counter example](https://explorer.perawallet.app/asset/1336655079/) about what you said. It is an ARC-3 fungible ASA (so, not an NFT), which is correctly interpreted by Pera.

Should we colse this PR?

PR #295, opened 2024-05-03
Fix error in ARC54 total (and add addresses)

joe-p:
Both @SilentRhetoric and myself agree that this was an error in the ARC. Total supply should *not* remove the burned supply. This metric is captured via `Available` supply. 

Also added in addresses to the deployments.


SilentRhetoric:
Agreed.  Thank you, Joe, for spotting this and creating the PR to correct the error.

PR #302, opened 2024-07-16
ARC-62 - ASA Circulating Supply

cusma:
This ARC introduces a standard to extend the definition of circulating supply for Algorand Standard Assets (ASA) and its client-side retrieval. An Algorand Python reference implementation is suggested (available at `./assets/arc-0062`).

For an initial discussion we should focus just on `./ARCs/arc-0062.md`.

The majority of reference implementation assets are AlgoKit auto-generated files. In case you are interested in the reference implementation as well, the main files are the contract (source file [here](https://github.com/cusma/ARCs/blob/arc62/assets/arc-0062/smart_contracts/circulating_supply/contract.py)) and the tests (source files [here](https://github.com/cusma/ARCs/tree/arc62/assets/arc-0062/tests)).


ARCs/arc-0062.md
```diff
@@ -0,0 +1,333 @@
...
+  "desc": "Get ASA circulating supply"
+}
+```
+
+The `arc62_get_circulating_supply` **MUST** be a _read-only_ ([ARC-22](./arc-0022.md))
+method (getter).
+
+### Usage
+
+Getter calls **SHOULD** be _simulated_.
```
SilentRhetoric, line 100: To enable the use case in which getter calls are made on-chain by another application (and not just simulated), would it be reasonable to include the following statement?

> The `arcXXX_get_circulating_supply` method **SHOULD** not require more foreign references than can be provided on a single applicationCall transaction within the constraints of the Algorand Virtual Machine.

This also raises another question, which is: Should the ARC require that its implementations specify the minimum AVM version required to call the method?

cusma: I think this is an implementation detail. I would rather add a recommendation for the reference implementation to use an AVM version that supports resource pooling. The specification of the method interface does not really care about _how_ resources are populated (e.g. they could be added to the AppCall or provided somehow within the same Group Transaction).

dolphinkitty:
ARC-62 requires the circulating supply app ID to be placed in the ASA's URL field. This conflicts with its typical use within fungible tokens of linking to the project website.

Maybe a more flexible solution that can support both link types would be to leverage the NFT metadata approach of linking to IPFS CIDs, allowing a single ASA URL field to point to a JSON object containing multiple resource links. For example, a fungible token could link to its project website, circulating supply application ID, and token image within a single metadata file.

Therefor, maybe we should  remove the requirement that "The Circulating Supply App ID **MUST** be declared as [ARC-82](./arc-0082.md) URI in the ASA URL field"

And then beyond the scope of ARC-62, we probably need standardized rules for how the URL field should be used and a standardized schema for any related metadata. I've shared a preliminary ARC proposal with Stephane, regarding ASA URL field usage and metadata schema.

cusma:
> ARC-62 requires the circulating supply app ID to be placed in the ASA's URL field. This conflicts with its typical use within fungible tokens of linking to the project website.
> 
> Maybe a more flexible solution that can support both link types would be to leverage the NFT metadata approach of linking to IPFS CIDs, allowing a single ASA URL field to point to a JSON object containing multiple resource links. For example, a fungible token could link to its project website, circulating supply application ID, and token image within a single metadata file.
> 
> Therefor, maybe we should remove the requirement that "The Circulating Supply App ID **MUST** be declared as [ARC-82](./arc-0082.md) URI in the ASA URL field"
> 
> And then beyond the scope of ARC-62, we probably need standardized rules for how the URL field should be used and a standardized schema for any related metadata. I've shared a preliminary ARC proposal with Stephane, regarding ASA URL field usage and metadata schema.

Hi @dolphinkitty, ARC-62 already supports ARC-3 and ARC-19 to declare ARC-62 App ID if the ASA has off-chain metadata. The addition of ARC-82 URI simplifies use cases in which off-chain metadata are not needed (e.g., fungible tokens) without forcing users to add IPFS (or other off-chain storage) dependencies. This is better for client responsiveness and latency since, with an single Algod call you can discover the ARC-62. Maybe we can relax the requirement from **MUST** to **RECOMMENDED**. This will be discussed in the Developer Council meeting tomorrow, Friday, March 7⋅16:00 – 17:00 (CET) Video call link: https://meet.google.com/mvy-uwha-uvt

PR #304, opened 2024-07-19
ARC-2 How to handle ARC data inside a note

SudoWeezy:
We would like to have a more precise version of how to read notes related to ARCs.

Here is a little list of options we have in mind:

1. `af/arc:j{<N1>:{...}, <N2>:{...}, ...}` (no 0 padding for <N>), allows multiple ARCs in one note field, harder to parse (maybe) 
1. `af/arc<N>:j{...}` (no 0 padding for <N>) single ARC per note field, maybe easier to fetch and parse 
1. `arc<N>:j{...}` (with 0 padding for <N>) single ARC per note field, maybe easier to fetch and parse

### DIFF
--- a/ARCs/arc-0002.md
+++ b/ARCs/arc-0002.md
@@ -2,7 +2,7 @@
 arc: 2
 title: Algorand Transaction Note Field Conventions
 description: Conventions for encoding data in the note field at application-level
-author: Fabrice Benhamouda (@fabrice102)
+author: Fabrice Benhamouda (@fabrice102), Stéphane BARROSO (@SudoWeezy), Cosimo BASSI (@cusma)
 discussions-to: https://github.com/algorandfoundation/ARCs/issues/2
 status: Final
 type: Standards Track
@@ -41,12 +41,9 @@ where:
     * Names starting with `a/` and `af/` are reserved for the Algorand protocol and the Algorand Foundation uses.
 
 - `<arc-number>` is the number of the ARC with 0 padding:
-  - Regexp to satisfy: `0\d{3}|\d{4}`
-  - Only contain a four-digit number, either zero-padded or not, matching:
-    - 0\d{3} for zero-padded numbers (e.g., 0001, 0456).
-    - \d{4} for non-padded four-digit numbers (e.g., 1234, 5678).
-    - Be exactly four digits long.
-
+  - Regexp to satisfy: `\b(0|[1-9]\d*)\b`:
+      - Only contain a digit number, without any padding
+  
 * `<data-format>` is one of the following:
     * `m`: <a href="https://msgpack.org">MsgPack</a>
     * `j`: <a href="https://json.org">JSON</a>
--- a/ARCs/arc-0002.md
+++ b/ARCs/arc-0002.md
@@ -87,6 +87,15 @@ The restrictions on dApp names were chosen to allow most usual names while avoid
 
 > This section is non-normative.
 
+Consider ARC-20, that provides information about Smart ASA's Application.
+
+Here a potential note indicating that the Application ID is 123:
+
+* JSON without version:
+    ```
+    arc20:j{"application-id":123}
+    ```
+
 Consider a dApp named `algoCityTemp` that stores temperatures from cities on the blockchain.
 
 Here are some potential notes indicating that Singapore's temperature is 35 degree Celsius:
--- a/ARCs/arc-0002.md
+++ b/ARCs/arc-0002.md
@@ -87,7 +87,7 @@ The restrictions on dApp names were chosen to allow most usual names while avoid
 
 > This section is non-normative.
 
-Consider [ARC-20]('./arc-0020.md'), that provides information about Smart ASA's Application.
+Consider [ARC-20](./arc-0020.md), that provides information about Smart ASA's Application.
 
 Here a potential note indicating that the Application ID is 123:
 
--- a/ARCs/arc-0002.md
+++ b/ARCs/arc-0002.md
@@ -40,10 +40,11 @@ where:
          * be at most 32 characters long
     * Names starting with `a/` and `af/` are reserved for the Algorand protocol and the Algorand Foundation uses.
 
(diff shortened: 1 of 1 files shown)


cusma:
My favorite one is option `2`.

Vidhyanandcs:
I would suggest option 3. 
The reasons are as follows: 
1) Very simple 
2) Easier to fetch and parse
3) No reference to algorand foundation (option 1 & 2 has a prefix `af` which i feel is unnecessary)

HashMapsData2Value:
I assume that j in this case is signaling that what's coming after is JSON?

cusma:
> I assume that j in this case is signaling that what's coming after is JSON?

Yes, but it's just for example sake. The standard should work with any ARC-2 `<data-format>. The proposal is just about the `<dapp-name>` harmonizations in case of ARCs.

emg110:
Option 3 it is better IMHO since it is not bound to any entity but ARC itself. It is simpler and more readable

pbennett:
inclined to side w/ option 3 as well.  af is kind of redundant - arc prefix is already descriptive enough.

cusma:
Seems we are leaning towards option 3.

One consideration: ARC-2 requires the `<dapp-name>` to "be at least 5 characters long". In order to avoid wasting the limited bytes of the transaction's note field with `0` padding, we could require a single `0` padding just for single digit ARCs.

k13n:
I'm also for option 3, seems clear enough

ARCs/arc-0002.md
```diff
@@ -33,14 +39,24 @@ where:
          * be at least 5 characters long
          * be at most 32 characters long
     * Names starting with `a/` and `af/` are reserved for the Algorand protocol and the Algorand Foundation uses.
+
+- `<arc-number>` is the number of the ARC with 0 padding:
+  - Regexp to satisfy: `0\d{3}|\d{4}`
+  - Only contain a four-digit number, either zero-padded or not, matching:
+    - 0\d{3} for zero-padded numbers (e.g., 0001, 0456).
+    - \d{4} for non-padded four-digit numbers (e.g., 1234, 5678).
+    - Be exactly four digits long.
```
k13n, line 48: I don't understand why the regex `0\d{3}|\d{4}`. Doesn't `0\d{3}` match a subset of what `\d{4}` matches? If so, couldn't the whole regex be simplified to `\d{4}`?

ARCs/arc-0002.md
```diff
@@ -2,7 +2,7 @@
 arc: 2
 title: Algorand Transaction Note Field Conventions
 description: Conventions for encoding data in the note field at application-level
-author: Fabrice Benhamouda (@fabrice102)
+author: Fabrice Benhamouda (@fabrice102), Stéphane BARROSO (@SudoWeezy), Cosimo BASSI (@cusma)
```
cusma, line 5:
```suggestion
author: Fabrice Benhamouda (@fabrice102), Stéphane Barroso (@SudoWeezy), Cosimo Bassi (@cusma)
```

ARCs/arc-0002.md
```diff
@@ -33,14 +39,22 @@ where:
...
     * `j`: <a href="https://json.org">JSON</a>
     * `b`: arbitrary bytes
     * `u`: utf-8 string
 * `<data>` is the actual data in the format specified by `<data-format>`
 
-**WARNING**: Any user can create transactions with arbitrary data and may impersonate other dApps. In particular, the fact that a note field start with `<dapp-name>` does not guarantee that it indeed comes from this dApp. The value `<dapp-name>` cannot be relied upon to ensure provenance and validity of the `<data>`. 
+**WARNING**: Any user can create transactions with arbitrary data and may impersonate other dApps. In particular, the fact that a note field start with `<dapp-name>` does not guarantee that it indeed comes from this dApp. The value `<dapp-name>` cannot be relied upon to ensure provenance and validity of the `<data>`.
+
+**WARNING**: Any user can create transactions with arbitrary data, including arc numbers, which may not correspond to the intended standard. An arc number included in a note field does not ensure compliance with the corresponding standard. The value of the arc number cannot be relied upon to ensure the provenance and validity of the <data>.
```
cusma, line 57:
```suggestion
**WARNING**: Any user can create transactions with arbitrary data, including ARC numbers, which may not correspond to the intended standard. An ARC number included in a note field does not ensure compliance with the corresponding standard. The value of the ARC number cannot be relied upon to ensure the provenance and validity of the <data>.
```

cusma (commented):
Should we add an example for ARCs usage (similar to dApps)?

SudoWeezy:
> Should we add an example for ARCs usage (similar to dApps)?

I don't know if it is necessary as it is just replacing `algoCityTemp` by `arc<arc-number>`
But if you have any particular example in mind, feel free to add one.

cusma:
> I don't know if it is necessary as it is just replacing `algoCityTemp` by `arc<arc-number>` But if you have any particular example in mind, feel free to add one.

I was thinking about the following addition to the **Reference Implementation** section:

```
Consider ARC-62, that provides information about ASA's Circulating Supply Application.

Here a potential note indicating that the Application ID is 123:

* JSON without version:
    ```
    arc62:j{"application-id":123}
    ```
```

Feel free to discard if not needed.

SudoWeezy:
> ```
> Consider ARC-20, that provides information about Smart ASA's Application.
> 
> Here a potential note indicating that the Application ID is 123:
> 
> * JSON without version:
>     ```
>     arc20:j{"application-id":123}
>     ```
> ```

Seeing this example got me asking. Should we clearly define in ARC-2 common fields like 
app-id asa-id? Or do we need to let this choice to the ARC that uses the note
We discuss a lot about minimizing the size of the note,

cusma:
> Seeing this example got me asking. Should we clearly define in ARC-2 common fields like app-id asa-id? Or do we need to let this choice to the ARC that uses the note We discuss a lot about minimizing the size of the note, so it might be worth to keep the field as short as possible

I think `<data>` is too ARC-specific to be standardized here in ARC-2.

PR #310, opened 2024-09-11
Add ABI Calls support to ARC-0026

emg110:
ABI Calls are a necessary building block for decentralized SOA interoperability and user-chain interaction. The support is included in this PR.

### DIFF
--- a/ARCs/arc-0026.md
+++ b/ARCs/arc-0026.md
@@ -128,7 +128,7 @@ Request 150 units of Asset ID 45 from an address
 algorand://TMTAD6N22HCS2LKH7677L2KFLT3PAQWY6M4JFQFXQS32ECBFC23F57RYX4?amount=150&asset=45
 ```
 
-Call claim(uint64,uint64)byte[] method on contract 11111111 from an address
+Call claim(uint64,uint64)byte[] method on contract 11111111 paying a fee of 10000 micro ALGO from an address
 
 ```
 algorand://TMTAD6N22HCS2LKH7677L2KFLT3PAQWY6M4JFQFXQS32ECBFC23F57RYX4?app=11111111&method=claim(uint64,uint64)byte[]&arg=20000&arg=474567&asset=45&fee=10000
--- a/ARCs/arc-0026.md
+++ b/ARCs/arc-0026.md
@@ -27,7 +27,7 @@ Elements of the query component may contain characters outside the valid range. 
 algorandurn     = "algorand://" algorandaddress [ "?" algorandparams ]
 algorandaddress = *base32
 algorandparams  = algorandparam [ "&" algorandparams ]
-algorandparam   = [ amountparam / labelparam / noteparam / assetparam / appparam /  methodparam / argparam/ boxparam / assetarrayparam / accountarrayparam / apparrayparam / otherparam ]
+algorandparam   = [ amountparam / labelparam / noteparam / assetparam / appparam /  methodparam / argparam / boxparam / assetarrayparam / accountarrayparam / apparrayparam / otherparam ]
 amountparam     = "amount=" *digit
 labelparam      = "label=" *qchar
 assetparam      = "asset=" *digit
--- a/ARCs/arc-0026.md
+++ b/ARCs/arc-0026.md
@@ -27,19 +27,29 @@ Elements of the query component may contain characters outside the valid range. 
 algorandurn     = "algorand://" algorandaddress [ "?" algorandparams ]
 algorandaddress = *base32
 algorandparams  = algorandparam [ "&" algorandparams ]
-algorandparam   = [ amountparam / labelparam / noteparam / assetparam / otherparam ]
+algorandparam   = [ amountparam / labelparam / noteparam / assetparam / appparam /  methodparam / argparam/ assetarrayparam / accountarrayparam / apparrayparam / otherparam ]
 amountparam     = "amount=" *digit
 labelparam      = "label=" *qchar
 assetparam      = "asset=" *digit
+appparam        = "app=" *digit
 noteparam       = (xnote | note)
 xnote           = "xnote=" *qchar
-note            = "note=" *qchar
+note            = "note=" *qchar 
+methodparam     = "method=" *qchar
+argparam        = "arg=" (*qchar | *digit)
+accountarrayparam = "account=" *qchar
+assetarrayparam = "asset=" *digit
+apparrayparam   = "app=" *digit
 otherparam      = qchar *qchar [ "=" *qchar ]
 ```
 
 Here, "qchar" corresponds to valid characters of an RFC 3986 URI query component, excluding the "=" and "&" characters, which this specification takes as separators.
 
 The scheme component ("algorand:") is case-insensitive, and implementations must accept any combination of uppercase and lowercase letters. The rest of the URI is case-sensitive, including the query parameter keys.
+
+Note 1: the first `app=` in the presence of the method interprets as the destination app Id and the next ones in sequence would be considered app array members.
+
+Note 2: the `asset=` in the presence of the method interprets as an asset array member.
 
 !!! Caveat
     When it comes to generation of an address' QR,  many exchanges and wallets encodes the address w/o the scheme component (“algorand:”). This is not a URI so it is OK.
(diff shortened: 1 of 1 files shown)


ARCs/arc-0026.md
```diff
@@ -99,12 +120,20 @@ Request 150 units of Asset ID 45 from an address
 algorand://TMTAD6N22HCS2LKH7677L2KFLT3PAQWY6M4JFQFXQS32ECBFC23F57RYX4?amount=150&asset=45
 ```
 
+call claim(uint64,uint64)byte[] method on contract 11111111
```
cusma, line 123:
```suggestion
Call claim(uint64,uint64)byte[] method on contract 11111111 from an address
```

ARCs/arc-0026.md
```diff
@@ -99,12 +120,20 @@ Request 150 units of Asset ID 45 from an address
...
+
+```
+algorand://TMTAD6N22HCS2LKH7677L2KFLT3PAQWY6M4JFQFXQS32ECBFC23F57RYX4?app=11111111&method=claim(uint64,uint64)byte[]&arg=20000&arg=474567&asset=45
+```
+
 ## Rationale
 
+The addition of ARC4 support to ARC26 is a mandatory step for effective interactivity of Algorand operations with users and increases interoperability amongst users as well
+
```
cusma, line 132: I think this section should focus more on the general and broader "Rationale" of the design choices rather than just the ARC4 addition.

cusma (commented):
It looks generally a good and useful addition.

I'm wondering, given that we are updating this spec, if wouldn't be a good opportunity to add not only a support for AppCall transactions but also for other transactions types (e.g. KeyRegistration).

Also, this makes me think that this spec does not clearly define a "transaction type" under which certain arguments are forbidden (e.g. trying to specify an `method` under a Payment transaction request).

ARCs/arc-0026.md
```diff
@@ -27,20 +27,31 @@ Elements of the query component may contain characters outside the valid range.
 algorandurn     = "algorand://" algorandaddress [ "?" algorandparams ]
 algorandaddress = *base32
 algorandparams  = algorandparam [ "&" algorandparams ]
-algorandparam   = [ amountparam / labelparam / noteparam / assetparam / otherparam ]
+algorandparam   = [ amountparam / labelparam / noteparam / assetparam / appparam /  methodparam / argparam/ boxparam / assetarrayparam / accountarrayparam / apparrayparam / otherparam ]
```
cusma, line 30:
```suggestion
algorandparam   = [ amountparam / labelparam / noteparam / assetparam / appparam /  methodparam / argparam / boxparam / assetarrayparam / accountarrayparam / apparrayparam / otherparam ]
```

emg110:
Yes @cusma , avoiding other transaction types for app calls was intentional to avoid much unneccessary complexity and just have ABI calls in full supported by ARC 26 , which is long over due IMHO

ARCs/arc-0026.md
emg110: Thank you

ARCs/arc-0026.md
```diff
@@ -99,12 +128,20 @@ Request 150 units of Asset ID 45 from an address
 algorand://TMTAD6N22HCS2LKH7677L2KFLT3PAQWY6M4JFQFXQS32ECBFC23F57RYX4?amount=150&asset=45
 ```
 
+Call claim(uint64,uint64)byte[] method on contract 11111111 from an address
```
cusma, line 131:
```suggestion
Call claim(uint64,uint64)byte[] method on contract 11111111 paying a fee of 10000 micro ALGO from an address
```

ARCs/arc-0026.md
```diff
@@ -56,10 +70,25 @@ The scheme component ("algorand:") is case-insensitive, and implementations must
...
-- asset: The asset id this request refers to (if Algos, simply omit this parameter) 
+- asset: The asset id this request refers to (if Algos, simply omit this parameter). When the method is available, the asset is an Asset Array member for the ABI call.
+
+- method: The full ABI method to call. E.g. claim(uint64,uint64)byte[]
+
+- arg: The argument to be added to ABI call arguments array.
+
+- box: The box name reference.
+
+- fee: Optional static transaction fee.
```
cusma, line 81: I think it is worth mentioning that this number MUST be expressed in microALGO.

emg110: Done! Committed! Thanks for the great points again. I'm grateful @cusma !

emg110:
> It looks generally a good and useful addition.
> 
> I'm wondering, given that we are updating this spec, if wouldn't be a good opportunity to add not only a support for AppCall transactions but also for other transactions types (e.g. KeyRegistration).
> 
> Also, this makes me think that this spec does not clearly define a "transaction type" under which certain arguments are forbidden (e.g. trying to specify an `method` under a Payment transaction request).
> 
> EDIT: I have a lot of point TBD, like references to things that may change at runtime (depending on the execution path or the implementation), like: apps, boxes, accounts, assets, fees, etc.
> 
> I see these are being discussed on other channels, so I will defer my comments there.

I 100% agree but scenarios during which these ABI call URIs become useful are the ones in PR description: SOA interoperability and user-chain interaction. Example: A service or a dApp creates those in runtime based on a requirement and shows the QRCode to get user mobile consent during an interrupted flow (Like walletconnect itself) for security purposes. 
I need to bri ng your attention to templating section as well! Using templating (ZeroAddress as Sender account address) any recipient wallet or dApp can replace connected user account and reconstruct a transaction from it which makes an ABI call.
Since the use case for ABI call URIs dictates it being created by the owner of the service/dApp which knows runtime requirements, it is safe to assume all those used params are known to URI issuer.
The issue with people in the channel is that when it comes to service architecture they see only a few scenarios and not the the whole story.

This PR especially can give rise to offline/indirect communications possibilities between dApps, services and users.

emg110:
Another great example would be Authenticate with your mobile device:
- User logs into a dApp on desktop using NFD name (account address behind does the login to smart contract)
- Now user wants to login on smart TV or another desktop without re-entering credentials
- Smart TV creates an Algorand URI to an ABI call for user account with smart TV serial number as arg to dApp and pops the QRCode
- User scans QRCode with wallet which is compliant to ARC26+this PR
- The transaction gets generated on mobile wallet , signed with account in URI if availabel on wallet and sent
- dApp on Smart TV or another desktop (subscribed to ASC events) listens and receives the TXN on ledger and checks serial number and user account and lets user in
-

scholtz:
Please add network attribute with genesis name
Please add more examples mainly how to use the note field

scholtz:
Lets please create arc260 for QRs with the protocol definition avm+web://arc260?txs=base64url([tx1,tx2,..]) where tx1 is the uint8 of the algosdk signed or unsigned tx. 

This will solve the issues 
- any app call can be signed using qr code - pay,axfer,app,keyreg, unreg...
- multiple transactions can be requested to be signed at once (pay, and call appcall)
- note field issues
- restriction to specific round
- restriction to specific network
- sender not defined

emg110:
> Please add network attribute with genesis name Please add more examples mainly how to use the note field

adding examples for the note field is out of this PRs scope but sure I will add those during weekend.
The rest of the comments should be under a separate topic as those are not the subject of this specific PR. I add those to the general ARC26 discussion.

tasosbit:
There is an arc26-like implementation in the wild that implements keyregs, which would be good to standardize in light of incentives rolling out soon.
Now that arc26 will cover more than just `pay` and `axfer` types, it would be good to have an explicit `type=` or `txtype=` parameter to disambiguate (as opposed to figuring it out from fields present/missing.)

emg110:
Thank you so much @tasosbit. Great point and it will be added to this PR as soon as possible.

cusma:
> There is an arc26-like implementation in the wild that implements keyregs, which would be good to standardize in light of incentives rolling out soon.
> Now that arc26 will cover more than just `pay` and `axfer` types, it would be good to have an explicit `type=` or `txtype=` parameter to disambiguate (as opposed to figuring it out from fields present/missing.)

I very much agree! @emg110 I had the same request in my previous comment.

emg110:
Closed in favor of [ARC -79 ](https://github.com/algorandfoundation/ARCs/pull/319)

PR #315, opened 2024-10-09
ARC-64: AVM Run Time Errors based on program bytecode

cusma:
This ARC formalizes the conventions to rise AVM run time errors based on program bytecode.

### DIFF
--- a/ARCs/arc-0065.md
+++ b/ARCs/arc-0065.md
@@ -1,0 +1,118 @@
+---
+arc: 65
+title: AVM Run Time Errors In Program
+description: Informative AVM run time errors based on program bytecode
+author: Cosimo Bassi (@cusma), Tasos Bitsios (@tasosbit), Steve Ferrigno (@nullun)
+discussions-to: https://github.com/algorandfoundation/ARCs/issues/XXX
+status: Draft
+type: Standards Track
+category: ARC
+created: 2024-10-09
+
+---
+
+## Abstract
+
+This document introduces conventions for rising informative run time error messages
+on the Algorand Virtual Machine (AVM) directly from the program bytecode.
+
+## Motivation
+
+The AVM does not offer native opcodes to catch and raise run time errors.
+
+The lack of native error handling semantics could lead to fragmentation of tooling
+and frictions for AVM clients, who are unable to retrieve informative and useful
+hints about the occurred run time failures.
+
+This ARC formalizes a convention to rise AVM run time errors based just on the program
+bytecode.
+
+## Specification
+
+The keywords "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**",
+"**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" 
+in this document are to be interpreted as described in <a href="https://datatracker.ietf.org/doc/html/rfc2119">RFC 2119</a>.
+
+> Notes like this are non-normative.
+
+### In Program Errors
+
+When a program wants to emit a run time error code, contained in the bytecode, 
+it **MUST**:
+
+1. Push to the stack the bytes string containing the error code;
+1. Push to the stack the `log` opcode;
+1. Push to the stack the `err` opcode.
+
+Upon a program run time failure, the Algod API returns both the failed *program 
+counter* (`pc`) and the `log` content.
+
+> The AVM programs bytecode have limited sized. In this convention, the errors are part of the bytecode. It is **RECOMMENDED** to use fixed length error codes or shorr error messages.
+
+### Example
+
+The program example raises the error code `ERR:001 - Invalid Method` for any application 
+call to methods different from `m1()void`.
+
+```teal
+#pragma version 10
+
+txn ApplicationID
+bz end
+
+method "m1()void"
+txn ApplicationArgs 0
+match method1
+byte "ERR:001 - Invalid Method"
+log
+err
+
+method1:
+b end
+
+end:
+int 1
+```
+
+Full Algod API response of a failed execution:
+
+```json
+{
+    "data": {
+        "app-index":1004,
+        "eval-states": [
+            { 
+                "logs": ["RVJSOiBJbnZhbGlkIE1ldGhvZA=="]
+            }
+        ],
+        "group-index":0,
+        "pc":41
+    },
+    "message":"TransactionPool.Remember: transaction ESI4GHAZY46MCUCLPBSB5HBRZPGO6V7DDUM5XKMNVPIRJK6DDAGQ: logic eval error: err opcode executed. Details: app=1004, pc=41, opcodes=pushbytes 0x4552523a20496e76616c6964204d6574686f64 // \"ERR:001 - Invalid Method\"; log; err; label2:"
+}
+```
+
+The `message` field contains the error code `ERR:001 - Invalid Method`.
+
+## Rationale
+
+This convention for AVM run time error codes presents the following PROS and CONS.
+
+### PROS
+
+- No additional artifacts required to bind the failed *program counter* (`pc`) to 
+the informative *error code*;
+- Error codes are returned directly in the Algod API response, no additional client
+operation needed.
+
+### CONS
+
+- Error codes consume program bytecode size.
+
+## Security Considerations
+
+> Not applicable.
+
+## Copyright
+
+Copyright and related rights waived via <a href="https://creativecommons.org/publicdomain/zero/1.0/">CCO</a>.


ARCs/arc-0065.md
```diff
@@ -0,0 +1,117 @@
...
+> Notes like this are non-normative.
+
+### In Program Errors
+
+When a program wants to emit a run time error code, contained in the bytecode, 
+it **MUST**:
+
+1. Push to the stack the bytes string containing the error code;
+1. Push to the stack the `log` opcode;
+1. Push to the stack the `err` opcode.
```
nullun, line 44: nit: these are opcodes that interact with the stack rather than being put on the stack.
```suggestion
1. Execute the `log` opcode to use the bytes from the top of the stack;
1. Execute the `err` opcode to immediately terminal the program.
```

joe-p:
I think it would make sense to include a special prefix to the log to signify the fact that it is an error message. This would remove the requirement of parsing the error message for the last three TEAL lines to determine if the last log was indeed an error message, which seems a bit fragile to me. 

On a related note, I think we should include a reference implementation for the showing the error message client-side.

Otherwise I don't have a problem with the ARC but it's not really clear to me where this would be used. In order to interact with an ARC4 method you need the signature, and if you can share the signature you might as well share ARC56 which has the source mapping with errors. I think it would be good to expand upon the motivation section with scenarios in which this is preferable.

joe-p:
Scratch that last part. I could imagine this being useful where you have an app ID for an ARC implementation where the implementation varies so you might not necessarily have the source map. Ie. ARC200 tokens. 

Still, I think it would be useful to add some more context to the ARC describing these sort of scenarios.

robdmoore:
From memory you only get logs back from a failed transaction if you do a simulate call (been a while since I checked so I could be wrong)? AlgoKit utils does have some logic to perform simulate on error (if debug config enabled). Also, the nice thing about this is the current pattern which is to use a TEAL comment and then if source maps are present so the 3 lines surrounding the error line would work nicely still with this model (and in fact would also work if you did a disassembly too?)

robdmoore:
Long story short: I like this technique

cusma:
> I think it would make sense to include a special prefix to the log to signify the fact that it is an error message. This would remove the requirement of parsing the error message for the last three TEAL lines to determine if the last log was indeed an error message, which seems a bit fragile to me.

In the example we suggest the usage of fixed-length error codes (`ERR:<code>`) or a their "full version" which include the error message itself (`ERR:<code> - <message>`).

I think we could formalize this better in the ARC, defining a precise regex for the error code and messages. @nullun WDYT?

@joe-p @nullun do you have suggestion for the error code/message regex?

> On a related note, I think we should include a reference implementation for the showing the error message client-side.

It's hard to come up with a general reference implementation, since the program could be anything and contain any error. It was easier to demonstrate a specific example. What do you suggest as reference implementation on top of the existing example?
 
> Otherwise I don't have a problem with the ARC but it's not really clear to me where this would be used. In order to interact with an ARC4 method you need the signature, and if you can share the signature you might as well share ARC56 which has the source mapping with errors. I think it would be good to expand upon the motivation section with scenarios in which this is preferable.

This approach to AVM run time errors simplifies _a lot_ the overhead of generating, synching, shipping and parsing (client side) the Source Map jointly with the Algod API response. Also, this would work also for non-ARC4 Smart Contracts. It's a general approach, closer to the pure AVM (due to the lack of a native error raising opcode).

nullun:
> From memory you only get logs back from a failed transaction if you do a simulate call

This was actually changed a few months back in https://github.com/algorand/go-algorand/pull/5875 and released in v3.22.0, allowing failed transactions to return PC, appID, stack, scratch, and logs. Although `goal` and I'm pretty sure the SDKs haven't been updated to read or expose this extra data to the user when it does fail. Submitting a transaction to the `/v2/transactions` API endpoint will return the data shown in this ARC though.

nullun:
> > I think it would make sense to include a special prefix to the log to signify the fact that it is an error message. This would remove the requirement of parsing the error message for the last three TEAL lines to determine if the last log was indeed an error message, which seems a bit fragile to me.
> 
> In the example we suggest the usage of fixed-length error codes (`ERR:<code>`) or a their "full version" which includes the error message itself (`ERR:<code> - <message>`).
> 
> I think we could formalize this better in the ARC, defining a precise regex for the error code and messages. @nullun WDYT?

Definitely agree with Joe, this should likely be defined as part of the standard so there's a sane way to parse them. I do like what is already there. `ERR:<code> - <message>` is straight forward enough and I feel like anyone (with or without a technical background) would understand this is an error message hopefully giving them a reasonable amount of information. If we wanted something more application specific (imagining people might be searching these errors online) maybe the appID could be included somewhere?

> @joe-p @nullun do you have suggestion for the error code/message regex?

Something like `s/^ERR:(\w*)\s-\s(.*)$/` ?

Screenshot 2024-10-10 at 09 31 15

joe-p:
`ERR:` prefix makes sense to me. 

One other thing to think about that isn't currently in the ARC is encoding. It's implied that error messages are UTF-8 strings but it should be explicit about that. Or we could think of a way to do UTF-8 strings now but also support event signatures, so this could be combined with something like ARC56 for dynamic error messages that don't take up too many bytes on-chain.

cusma:
@nullun @joe-p @robdmoore I've updated the ARC based on discussion feedback! If it's ok I'll try to push it for a final call.

ARCs/arc-0065.md
```diff
@@ -36,23 +36,38 @@ in this document are to be interpreted as described in <a href="https://datatrac
...
+### Error format
+
 > The AVM programs bytecode have limited sized. In this convention, the errors are
-> part of the bytecode. It is **RECOMMENDED** to use fixed length error codes or 
-> short error messages.
+> part of the bytecode, therefore it is good to mind errors' formatting and sizing.
+
+> Errors consist of a _code_ and an optional _short message_.
+
+Errors **MUST** statisfy the regexp: `s/^ERR:(\w*)\s-\s(.*)$/`.
```
nullun, line 56: I realise this regex wouldn't satisfy the below example of `ERR:042` where there isn't a short message.

Maybe the following would be better? Welcome others input here.
`s/^ERR:(\w*)(\s?-\s?(.*))?$/`

nullun: Just realised my example above created a third capturing group. This may be better: `ERR:(\w*)(?:\s?-\s?(.*))?` Which then gives you 2 capturing groups, the first for the error code, the second for the error message.

cusma: Good point! Also: both numeric (`ERR:0042`) and alphanumeric error codes should be supported (either "labels" or things like "hex" codes, like `ERR:BAD-REQUEST` or `ERR:0xAA`). I'll update the spec!

ARCs/arc-0065.md
```diff
@@ -0,0 +1,136 @@
...
+            }
+        ],
+        "group-index":0,
+        "pc":41
+    },
+    "message":"TransactionPool.Remember: transaction ESI4GHAZY46MCUCLPBSB5HBRZPGO6V7DDUM5XKMNVPIRJK6DDAGQ: logic eval error: err opcode executed. Details: app=1004, pc=41, opcodes=pushbytes 0x4552523a20496e76616c6964204d6574686f64 // \"ERR:001 - Invalid Method\"; log; err; label2:"
+}
+```
+
+The `message` field contains the error `ERR:001 - Invalid Method`.
```
nullun, line 113: As discussed, I think changing it to rely on the `logs` array would be better. Just decode each value and compare against the regex.
Here's an example I did via the command line with an application that attempts to use ABI methods first, then falls back to non-ABI methods. It resulted in two logged errors before failing.
```
curl -s -X POST http://127.0.0.1:4001/v2/transactions --data-binary @appl_call.stxn | jq -r '."data"."eval-states"[0]."logs"[] | @base64d' | grep -E "ERR:(\w*)(?:\s?-\s?(.*))?"
ERR:01 - Invalid Method
ERR:02 - No arguments supplied
```

tasosbit: IIRC the SDKs (e.g. JS) currently only throw the error message, e.g. when you do `sendRawTransaction` in JS/TS.

To me, including the error code in the surfaced message makes it significantly more useful

ARCs/arc-0065.md
```diff
@@ -0,0 +1,169 @@
...
+
+AVM compilers (and related tools) **SHOULD** provide two error compiling options:
+
+1. The one specified in this ARC as **default**;
+1. The one specified in [ARC-64](./arc-0064.md) as fallback, if compiled bytecode
+size exceeds the AVM limits.
+
+> Compilers **MAY** optimize for program bytecode size by storing the error prefixes
+in the `bytecblock` and concatenating the error message at the cost of some extra
+opcodes.
```
tasosbit, line 149: Relying on `.logs` is great when you have them, but I believe a lot of SDKs would hide `.logs` when e.g. throwing from a sendRawTransaction. The error message is mostly guaranteed to be propagated through the SDK, backend->frontend in case of a backend executor, etc. Hiding the error code from the algod error message would significantly detract value imo.

Brindle79 (commented):
Merg all node stacks  in one holding

cusma:
> Merg all node stacks in one holding

Please stop spamming.

cusma:
@SudoWeezy since ARC-56 is now finalized, are we good to push this to the Last Call?

SudoWeezy:
Sure, will do it right now

PR #327, opened 2024-12-12
[ARC-81] Partkey integrity hash

tasosbit:
This ARC proposes a new identifier for checking the integrity of consensus participation keys, including in the context of key registration transactions.

The integrity hash aims to be a short, human readable identifier that hashes all relevant participation key material, which node runners can reference at a glance in order to verify the integrity of the participation keys between their node, wallet or later on an explorer.


tasosbit:
Given that participation keys are not portable between networks for the majority of users, should be include the network (e.g. via genesis hash) in the integrity hash? This would prevent accidents where a keyreg was meant for testnet for ended up on mainnet.

urtho:
The ARC-78 URL strives to be as minimal as possible and does not include GH or network-id .
Adding network id to ARC-81 might make it incompatible with ARC-78.

tasosbit:
> Adding network id to ARC-81 might make it incompatible with ARC-78.

Not sure how it makes them incompatible. If ARC78 QR doesn't encode network/GH, that is all the more reason to include it in the integrity hash. The information (GH) would be independently known everywhere where you would want to generate an integrity hash - node, wallets, explorers. 

> The ARC-78 URL strives to be as minimal as possible and does not include GH or network-id . 

Arguably ARC78 should include network as well - in the TUI we currently have to warn users to make sure they are on the correct network when going through the QR. If we include the network in the encoded URL, the wallet can detect a mismatch and warn or refuse appropriately. But even if we don't, not sure what would be stopping the wallet from including its current network GH in the integrity hash calculation

tasosbit:
Second thread: Should this cover keyreg offline txns as well as @urtho suggested offline?

If so, what do you think about same payload structure, including the keys & rounds as all-zero?

tasosbit:
Updated the proposal:

- includes genesis hash, so that integrity hashes are specific to networks
- specify keyreg offline integrity hashes
- updated implementation reference & added tests

cusma:
@tasosbit: is there any intention or plan to push this ARC forward?

Roseymr (approved):

PR #332, opened 2025-02-11
[ARC-71] Updates after Dev Council

SudoWeezy:
The ARC should use the term non-transferable ASA (NTA) rather than Soul Bound Token or Soul Bound ASA. Soul Bound token should only be mentioned in the Abstract/Rationale to explain where the design idea is coming from. Keeping the closeout possibility

Removing the "Soul bound" term makes it clear that an NTA is an ASA before anything else. Soul Bound

To have a Token completely unremovable from an account with a burn/revoke authority separate from the user, another ARC should be created.

### DIFF
--- a/ARCs/arc-0071.md
+++ b/ARCs/arc-0071.md
@@ -1,7 +1,7 @@
 ---
 arc: 71
-title: Consensual Soul Bound ASA
-description: Interface Soul Bound Algorand Standard Asset
+title: Non-Transferable ASA
+description: Parameters Conventions Non-Transferable Algorand Standard Asset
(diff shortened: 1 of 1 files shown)


LoafPickleWW:
How is this ARC different than Default Frozen?

SudoWeezy:
> How is this ARC different than Default Frozen?

It is really close, but "Default Frozen" is not a standard.
It's just an explanation of how to do it, a way for explorers/wallets to give a specific name to this kind of Asset. Also, it brings a way for people outside of the ecosystem a clear way of how to do it. (They might not be familiar with ASAs).

PR #341, opened 2025-05-27
ARC-84: ASA-Compatible Smart Contract Tokens

joe-p:
This ARC proposes a new smart contract token standard that aims to have ecosystem comparability and trusted accounting logic

For those unfamiliar with GitHub, you can view an easily readable version of the proposal here: https://github.com/joe-p/ARCs/blob/arc84/ARCs/arc-0084.md

TODO:

- [ ] Ensure all desirable use cases are reflected in the ARC
- [ ] Add URL to params
- [ ] Add holders count
- [ ] Add `data`/`note` field to `Transfer` data structure for attaching arbitrary data
- [ ] Define circulating supply mechanism(s)
- [ ] Use separate app specifically for metadata
- [ ] Way more testing
- [ ] More demos/examples

### DIFF
--- a/ARCs/arc-0084.md
+++ b/ARCs/arc-0084.md
@@ -1,0 +1,15 @@
+---
+arc: 84
+title: ASA-Compatible Smart Contract Tokens
+description: A standard for smart contract tokens that are programable but also ASA-compatible
+author: Joe Polny (@joe-p) 
+discussions-to: <URL>
+status: Draft
+type: Standards Track
+category: ARC
+subcategory: Application
+created: 2025=05-27
+requires: 4, 22
+---
+
+**TODO:** This is currently a placeholder to reserve the ARC number
--- a/ARCs/arc-0084.md
+++ b/ARCs/arc-0084.md
@@ -79,9 +79,487 @@ sequenceDiagram
 ```
 
 ## Specification
-The key words "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**", "**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**", and "**OPTIONAL**" in this document are to be interpreted as described in <a href="https://www.ietf.org/rfc/rfc2119.txt">RFC-2119</a>.
 
-The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Algorand platforms (go-algorand, ...).
+### Transfer Interface
+
+#### ARC56 Transfer Interface
+
+<details>
+<summary>Click to reveal JSON</summary>
+
+```json
+{
+    "methods": [
+        {
+            "name": "arc84_transfer",
+            "args": [
+            {
+                "name": "dataApp",
+                "type": "uint64"
+            },
+            {
+                "name": "transfers",
+                "type": "(uint64,address,address,uint64)[]"
+            }
+            ],
+            "returns": {
+            "type": "void"
+            },
+            "actions": {
+            "create": [],
+            "call": [
+                "NoOp"
+            ]
+            }
+        },
+    ]
+}
+```
+
+</details>
+
+#### TEALScript Transfer Interface Signatures
+
+```typescript
+export type TokenId = uint64;
+
+export type Transfer = {
+  tokenId: TokenId;
+  from: Address;
+  to: Address;
+  amount: uint64;
+};
+
+arc84_transfer(dataApp: AppID, transfers: Transfer[]): void
+```
+
+### Data Interface
+
+#### ARC56 Data Interface
+
+<details>
+  <summary>Click to reveal JSON</summary>
+  
+```json
+{
+    "methods": [
+    {
+      "name": "createApplication",
+      "args": [
+        {
+          "name": "transferApp",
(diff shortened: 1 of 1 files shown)


ARCs/arc-0084.md
```diff
@@ -0,0 +1,750 @@
...
+  symbol: bytes<8>;
+  /** The total supply */
+  total: uint64;
+  /** The number of digits to use after the decimal point when displaying the asset */
+  decimals: uint64;
+  /** The address that can modify the token metadata */
+  manager: Address;
+  /** The address of the app that implements the transfer hook. If this is not set, transfers are always approved */
+  transferHookApp: AppID;
+};
```
pbennett, line 511: You're missing the url field from the asset params.

ARCs/arc-0084.md
```diff
@@ -0,0 +1,750 @@
...
+
+/** ***************
+* Getter Methods
+**************** */
+
+arc84_collection_minted(id: CollectionId): uint64 
+
+arc84_metadata(key: MetadataKey): Metadata 
+
+arc84_balanceOf(id: TokenId, account: Address): uint64
```
pbennett, line 564: I'd also recommend an arc84_numHoldersOf(id: TokenId).
Determining the number of holders for a token is a common need, and with a global data registry, fetching literally all boxes (or even prefix-match fetch of all) will be painful, yet updating a counter on transitions from/to 0 balance will be easy to handle.

ARCs/arc-0084.md
joe-p: I originally intentionally omitted it because I thought it could just be a metadata field, but since it's such an essential param for clients probably worth it to add to the params

pbennett: You can't say it's an ASA upgrade/replacement and remove the literal one field many need for token and nft images (and that want it to be mutable).  😀

joe-p: Yeah that's fair. I was originally thinking we could just define a standard for how the URL is specified in the metadata, but adding it to params just makes way more sense for ASA compatibility.

ARCs/arc-0084.md
```diff
@@ -0,0 +1,750 @@
...
+
+### Allowances With Expiration & Cooldown: Safely Enable Delegation Features
+
+Approvals are a key feature of other smart contract tokens, such as ERC20 and ARC-200, but they have serious security implications. For example, a large amount of EVM exploits occur due to users giving protocols token approvals and that same protocol being exploited years later after they no longer us the protocol. This is the driving rationale for having an expiration time for allowances in ARC-84. By setting an expiration time, users can be sure a protocol will not have access to their tokens when they are done using the app without having to take any explicit action. Similarly, the cooldown mechanism enables a highly-requested feature, subscriptions, in a secure manner. By being supported directly in ARC-84, there is no implementation ambiguity and all apps in the ecosystem can clearly display allowance information to users.
```
fergalwalsh, line 627: AFAIU allowances and approvals exist in ERC20 mostly to work around the fact that there are no transaction groups in the EVM. It is not possible for example to do a swap by doing a transfer of USDC + swap call in one transaction. Instead the swap contract is given approval to spend USDC, followed separately by a contract call to swap usdc to something. 

As I mentioned above, subscriptions and other allowance based features seem more appropriate for a wallet app standard than an asset standard.

ARCs/arc-0084.md
```diff
@@ -0,0 +1,750 @@
...
+
+Smart contract token standards on Algorand have to choose between `uint64` and `uint256` for supply and balance arithmetic. Past ARCs, such as ARC-200, have chosen `uint256` to align with EVM standards. This allows for a better bridging experience with non-AVM chains that use `uint256`, but hurts the compatibility with the current ASA-based ecosystem. This ARC prioritizes compatibility with the AVM ecosystem, thus uses `uint64` over `uint256`. Compatibility with non-AVM chains should be tackled as a separate problem.
+
+### Collections: On-Chain Guarantees for Valuable Collection Data
+
+The collection an NFT is in contributes to the value of the NFT, so it seems to make sense to include some guarantees about that collection on-chain. ARC-84 allows collections to have guaranteed immutability for metadata, a mint cap for tokens in a collection, and a manager address to control who can mint tokens in the collection and modify metadata. These primitive functionalities can be extended to offer as much if not more features than what is currently available through other ARCs such as ARC-53.
+
+## Backwards Compatibility
+
+### Contract Logic: Bridging ASAs
```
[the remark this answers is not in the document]
fergalwalsh: Thanks. So if the taxes are only enforced when bridging but the ASA is then freely tradable doesn't that encourage users to use the ASA version instead of the ARC84 version? I'm struggling to understand how you can have both an ASA & ARC84 of the same asset and still enforce any of the features of the ARC84.

cusma:
@joe-p: is there any intention or plan to push this ARC forward?

joe-p:
Right now there isn't an immediate use case. That being said, I think this is the best approach to SC tokens that we've had and there was general supportive consensus in the dev council. As such, I think it makes sense to merge as a draft so we can refer to it more easily going forward if we ever want to focus on SC tokens again.

Roseymr (approved):

Roseymr (approved):

Roseymr (approved):

PR #344, opened 2025-06-17
ARC-85: Revocable Decentralized Recovery (ReDeRec)

tasosbit:
This ARC outlines a new approach to Decentralized Recovery which enables revocability.

Instead of sharing private key shards with custodians, an Algorand account can be rekeyed to a specially crafted multisig so that the owner account can self-sign, and the recovery accounts ("custodians") can combine in a multisig threshold fashion (2/3, 3/4, etc) to help recover the account.

Self-signing is enabled by repeating the owner account's keys in the subsigners field, in order to satisfy the threshold requirement as a standalone signer. Revoking custodians is as simple as rekeying to another account.

For those unfamiliar with GitHub, you can view an easily readable version of the proposal here: https://github.com/tasosbit/ARCs/blob/arc85-active-derec/ARCs/arc-0085.md

TODO:

[ ] Test cases
[ ] Reference Implementation?

### DIFF
--- a/ARCs/arc-0085.md
+++ b/ARCs/arc-0085.md
@@ -1,9 +1,9 @@
 ---
 arc: 85
-title: Active DeRec
+title: Revocable Decentralized Recovery (ReDeRec)
 description: Revocable Decentralized Recovery protocol utilizing rekeying and multisigs instead of sharing mnemonic shards.
 author: Tasos Bitsios (@tasosbit)
-discussions-to: <URL>
+discussions-to: https://github.com/algorandfoundation/ARCs/pull/344
 status: Draft
 type: Standards Track
 category: ARC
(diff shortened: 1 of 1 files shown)


ARCs/arc-0085.md
```diff
@@ -0,0 +1,74 @@
...
+        "ALICE234..",
+        "BOB345..",
+        "EVE456..",
+        "STEVE567..",
+    ]
+}
+```
+
+## Backwards Compatibility
+This method requires a wallet that supports rekeying and multi-signature accounts. Beyond that on-chain compatibility should be 100%, as transaction signature schemes are transparent as far as Applications are concerned. Off-chain systems validating account ownership for "login" functionality must be able to support multi-sig accounts.
```
cusma, line 60:
```suggestion

- On-Chain: Guaranteed, as both _multisig_ and _rekey_ are native protocol features, and transaction signature schemes are transparent to Applications.

- Off-Chain: Wallet **MUST** support _rekeying_ and _multisig_ accounts. Off-chain systems validating account ownership for "login" functionality **MUST** support _multisig_ accounts.
```

ARCs/arc-0085.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Test Cases
+TODO
+
+## Reference Implementation
+TODO
+
+## Security Considerations
+Note that when accounts are closed out (i.e. taken to a zero ALGO balance) their rekeying status reverts to the default, so users of active DeRec are RECOMMENDED to maintain a minimum balance requirement above the base account MBR (0.1 ALGO) as a safety precaution against accidental close-outs.
```
cusma, line 69:
```suggestion
Note that when accounts are closed out (i.e., taken to a zero ALGO balance), their rekeying status reverts to the default, so users of active DeRec are **RECOMMENDED** to maintain a minimum balance requirement above the base account MBR (0.1 ALGO) as a safety precaution against accidental close-outs.
```

acfunk:
When discussing wallets in the Backwards Compatibility section, it may be worth mentioning that while Multi-Sig support is required, an additional feature of bypassing the signature collection step when the signer has enough weight is highly encouraged to make the UX bearable.

cusma:
@tasosbit: is there any intention or plan to push this ARC forward?

Roseymr (approved):

PR #347, opened 2025-07-16
ARC-87: Key Name Standard

PhearZero:
This ARC attempts to codify a standard key name specification.


PhearZero:
>One thing I think we should think about: supporting some sort of reference semantics so the key for a field can be stored in a box

This is a great idea and optimization, it should be able to fit into the semantics of this specification as a standalone ARC. Heavy +1 on the `$` as the operator and a dedicated prefix. The mapped keys can be treated like an object to make it easier to apply the template variables! 

The only real downside to adding a template to the key names is the increase in downstream cycles, it would require additional requests/information to index the resolved data. Splitting it into multiple ARCs may be the best for separation of concerns

```typescript
// Pre-processed output from ARC87
const arc87 = {
  "${k}": {
     "myKey": "data for super long key"
  },
  "${another.path}": "data for template path"
}

// Additional Processing by ARC? (based on ARC87)
const keys_obj = {
   "k": "this is a really long key that for some reason is extra long even though it probably doesn't need to be this long but idk maybe someone has a key this long",
   "another": { "path": "hello from a nested long string, reusing the ARC87 primitive and addressable everything"}
}

const obj = withARCXKeyMapping(arc87, keys_obj)
// key names are combined
```

I am hopeful that it engages a conversation around key name length, even at 64 it should fit quite a lot of use cases. 128 would most likely be ideal.

Future ARC possibility:

We can also have sets of keys that have additional meaning like the Blob/File examples. If a contract has a `$type` or some known key collection at a path, we can have tooling around the shapes and highly optimize/pack the key names and values. These can be formal proofs, ways of optimizing space or any other approach (Leaving the base assumption unchanged with the separator/indexes)

SudoWeezy (dismissed):
Just fixing the linting to pass the CI

PR #356, opened 2025-09-22
ARC-89: ASA Metadata Registry

cusma:


ARCs/arc-0089.md
```diff
@@ -0,0 +1,327 @@
...
+##### Create Metadata
+
+To create the Asset Metadata:
+
+- The ASA **MUST** _exist_, and
+
+- The Asset URL (`au`) **MUST** conform to this specification (see the [Asset URL
+section](#asset-url)), and
+
+- The authorization **MUST** be restricted to the Asset Manager.
```
p2arthur, line 121: I think it’d be clearer if we more explicitly distinguish the on-chain ASA from the registry’s Asset record. I’m seeing new records with fields like asset_manager, which can blur whether we mean the ASA’s manager or a property of the asset record in the ARC89 registry.

pbennett:
Yup, I was pointing out in offline convo that callers really can't just be expected to 'read metadata' and get an opaque blob of bytes that's on them to figure out what to do with it. There needs to be an expectation of what data is stored there and in what format, so I'd suggested utf-8 encoded json.  As someone who would presumably be reading metadata from this, I don't want to have to code reading arbitrary encodings or formats (json, jsonb, msgpack, gz'd text, tar, gzip, xhtml, .. ?).  arc3 is already json - let's just say it should be json like arc3.  
Other arcs can codify other properties that could be in the metadata json (like team accounts, app ids (arc62), etc).

cusma:
Since the Asset Metadata are _mutable_ (unless set as _immutable_), would it be useful to record the block/timestamp of the last metadata update?

PR #368, opened 2026-01-21
ARC-62 - Integrate with ARC-54, ARC-89, and update reference implementation

cusma:
This PR does not introduce any breaking change to the standard; the interface is preserved.

We introduce support for ARC-54 (ASA burning address) and ARC-89 (which extends this ARC).

We provide an updated reference implementation for a _singleton_ App that will reduce friction on the adoption.

UPDATE: Based on feedback from the Dev Council meeting, the `custom` labels have been increased to 4.


cusma:
> The reference implementation of the _singleton_ App provides the following "non-circulating" labeled-addresses for an ASA:
> 
> * Reserve (Treasury)
> * Burned (ARC-54)
> * Locked
> * Custom (user-defined meaning)
> 
> Should we add some more usual non-circulating labels as:
> 
> * Vested
> * Escrowed
> * Staked
> * Collateralized
> * Frozen / Siezed
> 
> Or maybe add just other "custom" addresses (with user-defined meaning)?

Based on feedback from the Dev Council meeting, the `custom` labels have been increased to 4.

Roseymr (approved):

